PR #205, opened 2014-12-10
Builds two gems from one repo.

jdantonio:
This update will now build two gems from within one repo.

### Building

Commands to build the gems:

* `rake build:core` under MRI or Rbx will build the pure-Ruby `concurrent-ruby` gem
* `rake build:core` under JRuby will be the Java-specific `concurrent-ruby` gem
* `rake build:ext` under MRI will build the `concurrent-ruby-ext` gem with uncompiled C extensions
* `rake build:native` under MRI will build the `concurrent-ruby-ext` gem with pre-compiled C extensions for the current platform
* `rake build` under MRI runs both `rake build:core` and `rake build:ext`
* `rake build` is an alias for `rake build:core` under JRuby or Rbx
* The [rake-compiler-dev-box](https://github.com/ruby-concurrency/rake-compiler-dev-box) scripts have not yet been updated

### Installing

The `concurrent-ruby-ext` gem lists `concurrent-ruby` as a runtime dependency. Installing `concurrent-ruby-ext` will cause `concurrent-ruby` to be installed.

Installing `concurrent-ruby` alone will install the JRuby-specific gem under JRuby. It will install the pure-Ruby gem under any other platform.

### Using

All the user ever needs to do is `require 'concurrent'` (or a load a subset of the gem, as in `require 'concurrent/atomic'` or `require 'concurrent/atomics'`) and the C extensions will load *if an only if* running under MRI *and* the `concurrent-ruby-ext` gem has been installed.

### Testing

The scripts in the `build-tests` directory will test the various gem packages built with the aforementioned rake tasks. The `runner.rb` script consolidates all the individual tests against the relevant builds. The individual `*.gem` files need to be in the `pkg` directory and the `runner.rb` script needs run from a directory *other* than the root directory of this repo. The approach I normally use is:

1. Clone the [ruby-concurrency/rake-compiler-dev-box](https://github.com/ruby-concurrency/rake-compiler-dev-box) repository
1. Clone the [ruby-concurrency/concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby) repository *under* the `ruby-concurrency/rake-compiler-dev-box` repository
1. Use the build scripts to build the gem packages needing tested
1. Run the `test-concurrent-ruby.sh` script from within the `ruby-concurrency/rake-compiler-dev-box` repository

This will not test the Windows builds. The `ruby-concurrency/rake-compiler-dev-box/tests/windows-test.cmd` file will do that, but it requires a Windows machine and manual setup of various Ruby versions.


pitr-ch:
hi @jdantonio, what was the reason behind dropping pre-compiled gems for unix-like systems?

jdantonio:
@pitr-ch My builds were broken. I decided it was better to release 0.8 without them and figure it out later rather than hold off the release.

My professional experience with C and C++ is several years old and was strictly with Visual C++ and the full MS stack. I have virtually no experience building on Linux. One of our users reported that the pre-compiled extensions weren't working on Heroku. The way the error messages worked on 0.7.0 I couldn't determine the root of the problem. It could have been that Heroku was downloading the pure-Ruby gem or it could have been a problem with the extension gem itself. Under 0.8 I changed the way the error messages worked and created a project specifically for testing on Heroku (https://github.com/jdantonio/heroku-tester) and determined it was the latter: Heroku was correctly installing the pre-compiled extension but Ruby was raising an exception when it attempted to load the .so module. It was raising a `LoadError` exception and the error message stated that the .so file did not exists. I was able to run some shell commands against Heroku and determine that the file was in the correct place and had the right permissions. My best guess is that the C code was crashing on load and Ruby was giving a bad error message.

The most frustrating thing about this is that the pre-compiled extension gem worked fine on the Ubuntu VM I used to build the gems!

Since I have little experience building C on Linux I wasn't able to figure out what the problem was. I assume it has something to do with the version of Ruby or the version of Ubuntu I used for building but I honestly don't know. I've left all the build commands for Linux in the build scripts but commented them out. Once I can figure out what's going on I would be happy to start releasing the pre-compiled Linux binaries, but I need to make sure they work consistently first.

pitr-ch:
@jdantonio Thanks for the detailed explanation!

schmurfy:
It may have nothing to do with this but on linux/unix the library will link against system libraries if they are needed as dependencies, if you compile it on a system with a specific version of said dependency and then try to run it on another one with different version it may not work.
The others obvious problems are different architecture and 32/64 bits but I imagine you already ruled them out.

If you just run ldd on a binary / shared library the system will tell you what are the dependencies, something like this:

```
$ ldd /usr/local/lib/libnet.so
	linux-vdso.so.1 =>  (0x00007fffde5ff000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0f164ff000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0f16ae1000)
```

Issue #117, opened 2014-06-23
Intermittently Failing Tests


jdantonio:
I'm still trying to address all the intermittently failing tests. It's a [Sisyphean](http://en.wikipedia.org/wiki/Sysiphus) task. Please use this thread to link to any failing Travis test runs that you believe are due to buggy tests.

jdantonio:
Moved to #263

PR #217, opened 2015-01-01
Timer set fixes

### LINKED ISSUE #216: TimerSet doesn't pop any timers until the first timer pops
```
$ cat timer_set_bug_test.rb 
require 'concurrent'

puts Time.now

Concurrent::timer(12) { puts Time.now }

sleep 1

Concurrent::timer(6) { puts Time.now }

sleep 15
```

I'd expect this to print three timestamps, each 6 seconds apart. Instead, the six-second timer doesn't fire until after the twelve-second timer:

```
$ ruby timer_set_bug_test.rb 
2015-01-01 11:15:39 +0000
2015-01-01 11:15:51 +0000
2015-01-01 11:15:51 +0000
```

I'm pretty confident that I can see the issue (we never signal the condvar), so I'm preparing a patch now.

rkday:
This fixes #216 and adds a UT for it. I also think I spotted a possible race condition (where we peek from a queue, do something with that task, and then pop from the queue, without guarding against the possibility that the head of the queue has changed since we peeked).

I checked that:
* `rake spec SPEC=spec/concurrent/executor/timer_set_spec.rb` passes
* `rake spec SPEC=spec/concurrent/executor/timer_set_spec.rb` fails without my commit 894ac23

### DIFF
--- a/lib/concurrent/executor/timer_set.rb
+++ b/lib/concurrent/executor/timer_set.rb
@@ -55,11 +55,10 @@ module Concurrent
           @queue.push(Task.new(time, args, task))
           @timer_executor.post(&method(:process_tasks))
         end
-
-        true
       end
-      @condition.signal
 
+      @condition.signal
+      true
     end
 
     # For a timer, #kill is like an orderly shutdown, except we need to manually
--- a/lib/concurrent/executor/timer_set.rb
+++ b/lib/concurrent/executor/timer_set.rb
@@ -130,8 +130,20 @@ module Concurrent
         interval = task.time - Time.now.to_f
 
         if interval <= 0
+          # We need to remove the task from the queue before passing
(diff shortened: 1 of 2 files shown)


jdantonio:
Thank you!

With respect to the failed Travis build, I've never seen that particular test fail on JRuby before. I don't believe it has anything to do with this patch so I've added it to the list of intermittently failing tests listed in Issue #117. We can look into it separate from this pull request.

jdantonio:
@rkday The failing JRuby tests seems to be a TravisCI issue. Our `.travis.yml` file specified `jruby-19mode` as the main JRuby build target. It appears that the fuzzy install that Travis initiates now installs `jruby 9.0.0.0-SNAPSHOT` for that directive. I've created PR #218 to fix.

PR #239, opened 2015-02-11
Prune thread pool only after handling request.

SebastianEdwards:
Fixes an issue where all workers can be pruned leaving no running workers to handle the incoming request.


pitr-ch:
LGTM, Thanks!
Relevant discussion can be found here https://gitter.im/ruby-concurrency/concurrent-ruby?at=54dade43ea6e362732ffa844

jdantonio:
The original intent *was* to allow the pool to drain to zero (when no minimum pool size was set). When that code was written we didn't have our current actors. If the original behavior is causing problems then it definitely needs changed. 

@SebastianEdwards Thank you very much for finding and fixing this!

PR #240, opened 2015-02-11
JavaAtomicFixnum and JavaAtomicBoolean - ThreadContext should not be an object field.

lucasallan:
@pitr-ch Could you take a look? As you suggested, ThreadContext should not be an object field.


jdantonio:
:+1:

pitr-ch:
:+1: looks good, thanks

PR #249, opened 2015-02-23
Fix ScheduledTask sometimes not accepting an interval of 0

obrok:
ScheduledTask measured `Time.now` in 2 places with some computation in
between and it sometimes arrived at an inconsistent result.

Sorry for the ugly spec - this is basically the way I reproduced it and I didn't want to bind too much to the implementation by stubbing out `TimerSet.calculate_schedule_time` with a sleep.

Also - it only happened on jruby for me - maybe the clock is more accurate?

### DIFF
--- a/lib/concurrent/scheduled_task.rb
+++ b/lib/concurrent/scheduled_task.rb
@@ -25,8 +25,9 @@ module Concurrent
     # @since 0.5.0
     def execute
       if compare_and_set_state(:pending, :unscheduled)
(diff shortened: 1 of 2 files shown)


jdantonio:
Many of our tests are ugly, but sometimes that the only way to *truly* test this library. I'd be afraid of any concurrency library that used too many test doubles. If your spec revealed the problem and verified the solution it's good for me!

PR #251, opened 2015-02-24
default `count` in CountDownLatch to 1

tenderlove:
it's very common to use a CountDownLatch with a count of 1.  Default
`count` to 1 so that users can just call `new` and get a valid object
with a count of 1.

I'd like to use this to [replace the Latch implementation in Rails](https://github.com/rails/rails/blob/a0580e974b4a058a983de6c593e4573bd94b76f0/activesupport/lib/active_support/concurrency/latch.rb).


jdantonio:
Thank you for the PR! Please let us know if there are other features you'd like to see.

The failed tests are unrelated to this PR and and known to be buggy tests (Issue #117).

tenderlove:
@jdantonio thank you!

PR #252, opened 2015-02-24
move `initialize` to be in one place, and always define JavaCountDownLatch

tenderlove:
we can keep the parameter checks in one place if there is one definition
of initialize.  This patch also always defines JavaCountDownLatch, so we
can reduce the conditional a bit

I don't know if this patch is worthwhile, just an RFC.  If we move `initialize` to the concrete implementation, we can slightly reduce some of the duplicate code.  `init_internal` should probably be private, and I don't understand how to do the documentation tags.  The code duplication is so tiny that I'm not sure this is worthwhile, so if you don't want the patch, I understand!

### DIFF
--- a/lib/concurrent/atomic/count_down_latch.rb
+++ b/lib/concurrent/atomic/count_down_latch.rb
@@ -12,12 +12,6 @@ module Concurrent
   #   When the latch counter reaches zero the waiting thread is unblocked and continues
   #   with its work. A `CountDownLatch` can be used only once. Its value cannot be reset.
   class MutexCountDownLatch
-    def init_internals(count) # :nodoc:
-      @mutex = Mutex.new
-      @condition = Condition.new
-      @count = count
-    end
-
     # @!macro [attach] count_down_latch_method_wait
     #
     #   Block on the latch until the counter reaches zero or until `timeout` is reached.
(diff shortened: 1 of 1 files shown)


headius:
Looks ok to me.

lucasallan:
:+1:

jdantonio:
In principle I'm OK with this approach but we're getting 11 test failures on every version of MRI and JRuby and 20 failures on Rbx. I'll pull the branch to my local machine and see if I can figure out what the issue is.

jdantonio:
The reason for the failing tests is that this changes the class hierarchy.

When I started working with JRuby-specific classes I was very unfamiliar with JRuby. I created the class hierarchy such that the pure-Ruby version of every class was always available on every runtime as a fallback. What this means is that on JRuby we expect to see both `MutexCountDownLatch` and `JavaCountDownLatch` and--more importantly--that they both work. With this patch neither `MutexCountDownLatch` nor `JavaCountDownLatch` works correctly by itself because neither has the appropriate constructor. Only `CountDownLatch` works. In practice this should make no difference since I expect most users to simply use `CountDownLatch` and never use `MutexCountDownLatch` directly. But it's a breaking, not-backward-compatible change and it strays from the convention used consistently throughout the gem.

Now that this gem has matured and stabilized I'm comfortable with us changing the convention of always having the pure-Ruby implementation available. If we were to go that route we'd want to be explicit about the new convention and we'd probably want to update everything within the gem to match. And we'd need to consider updating the major version number since the changes would break the current API.

For now I'm going to close the PR w/o merging. If anyone feels strongly about changing the aforementioned internal convention then we can open a new issue and talk about it.

@tenderlove Thank you very much for the patch! It highlights an area where we may be able to refactor and make improvements.

tenderlove:
Sounds good! :-D

pitr-ch:
Thanks @tenderlove, as @jdantonio said this could be improved. We could create `AbstractCountDownLatch` as a common parent of both implementations which wold hold the common code and implement only `init_internal` methods. Then we can keep the convention we have but also clear this up.

tenderlove:
Ya, probably pushing the constructor up to an abstract class would work out better than what I did.  The repeated code was so small though, I was hesitant to even send this PR. ¯\_(ツ)_/¯

PR #253, opened 2015-02-24
use a monotonic clock for timeouts

tenderlove:
We should use a monotonic clock for timeouts so that if the time changes
on the system it will not impact the timeout code.

### DIFF
--- a/lib/concurrent/atomic/condition.rb
+++ b/lib/concurrent/atomic/condition.rb
@@ -42,13 +42,13 @@ module Concurrent
     # @param [Object] timeout nil means no timeout
     # @return [Result]
     def wait(mutex, timeout = nil)
-      start_time = Time.now.to_f
+      start_time = clock_time
       @condition.wait(mutex, timeout)
 
       if timeout.nil?
         Result.new(nil)
       else
-        Result.new(start_time + timeout - Time.now.to_f)
+        Result.new(start_time + timeout - clock_time)
       end
     end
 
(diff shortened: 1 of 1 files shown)


lib/concurrent/atomic/condition.rb
```diff
@@ -66,5 +66,17 @@ def broadcast
       true
     end
 
+    private
+
+    if defined?(Process::CLOCK_MONOTONIC)
+      def clock_time
+        Process.clock_time Process::CLOCK_MONOTONIC
```
jdantonio: @tenderlove Should this method be `clock_gettime`?

This is a Ruby feature I wasn't aware of until you created this PR so thank you for pointing it out. This solves a problem we've discussed in another issue. A ton of tests are failing, however. Should the method call be this? http://ruby-doc.org//core-2.2.0/Process.html#method-c-clock_gettime

tenderlove: Yes it should be.  That's what I get for not running the tests. >_<

I'll update the patch

jdantonio:
No sweat. Our tests take a long time to run. Even I skip them occasionally!

tenderlove:
Also, this code will result in allocations of floating point objects.  If we only care about resolution down to the second, we could do:

```
irb(main):001:0> Process.clock_gettime Process::CLOCK_MONOTONIC, :second
=> 347802
irb(main):002:0>
```

Which will only give integers (so they should be allocation free)

chrisseaton:
> this code will result in allocations of floating point objects

Aren't small-enough `Floats` tagged these days?

tenderlove:
> Aren't small-enough Floats tagged these days?

Actually you're totally right.  If they are small enough, they won't be allocated.  So nevermind! :sparkles:

Issue #248, opened 2015-02-16
Concurrent::RejectedExecutionError when running with minitest/autorun via JRuby


iNecas:
I've been hitting Concurrent::RejectedExecutionError when running my test-stuite. It was ok with MRI, but switching to JRuby was the game-changer. It turns out there seems to be an at_exit trap that shuts down the Java thread pool. I've first tried to fix that using the same approach as in this test suite, reinitializing the Concurrent.configuration. However, there still was an issue with `@root` variable in actor keeping the reference to an old thread pool via Delayed (https://github.com/ruby-concurrency/concurrent-ruby/blob/124ea7e5de7477747ebc149bb4b9c34e22457995/lib/concurrent/actor.rb#L42-L46)

So I needed to reinitialize this as well the hackish way:

https://github.com/iNecas/dynflow/commit/b9a7bf2345404e35dda26bf8ecb86f0cf92fcc8c

What do you think about having set of helpers that helps with setting the testing environment and this kind of issues (I guess there might be more: we're not using all the concurrent-ruby features.)

jdantonio:
@iNecas We definitely need to figure out a better way to handle shutting down the Java thread pools.

The reason for the `at_exit` hook is that Java's thread pools will prevent the JVM from shutting down if the thread pool is not explicitly shut down. Under JRuby our thread pools are simple wrappers around the thread pools from `java.util.concurrent` and it's the underlying thread pool that causes the problem. When I first encountered this it took me a while to figure out what was happening so I grew concerned that users would have the same problem. The `at_exit` hook solved the problem by forcing the pool to be explicitly shut down. Our current `Actor` implementation came much later and that's when the `at_exit` hook became a problem.

I wonder if we're trying to solve this problem the wrong way. Rather than add extra layers *around* the thread pools, could we solve the problem allowing the creator of a given pool to *not* create the `at_exit` hook? Something like:

```ruby
pool = Concurrent::FixedThreadPoool.new(5, set_shutdown_hook: false)
```

[...]

Issue #227, opened 2015-01-16
Document or improve global-pool's auto termination


pitr-ch:
If `concurrent-ruby` is used in tests of other project, the thread pools are terminated before the tests are executed. Reason is the order of `at_exit` callbacks, termination runs before the test run is executed.

rkday:
This was part of my motivation for https://github.com/ruby-concurrency/concurrent-ruby/pull/199 - I hit exactly this with minitest/autorun, and it's much easier to spot if you get an exception by default.

It's easy to work around by setting `Concurrent.configuration.auto_terminate = false` in a minitest_helper.rb file.

pitr-ch:
the solution is partly also discussed in #248

jdantonio:
Addressed in PR #255.

Issue #149, opened 2014-08-10
Remove configuration of global pools


pitr-ch:
-   keep them lazy initialized
-   rename: task -> short-non-blocking, operation -> long-blocking
-   add more global executors like an immediate executor
-   remove configuration object, move global pools to `Concurrent` space
-   possibly improve parsing of :executor option 
    -   to also handle agent two pools
    -   to understand :immediate too
    -   etc.
-   what to do with logging?

@chrisseaton @jdantonio, I've tried to capture the result of the discussion.

jdantonio:
Addressed in PR #255.

PR #176, opened 2014-10-23
First draft IVar-Future-Promise-Probe unification

pitr-ch:
Not ready for merge. Opening discussion. Compatibility layer is totally missing for now.

PR #255, opened 2015-02-27
Thread Pool Updates

jdantonio:
Addresses #248, #227, and #149.

### Changelog

* Added `at_exit` handler to Ruby thread pools (already in Java thread pools)
  - Ruby handler stores the object id and retrieves from `ObjectSpace`
  - JRuby disables `ObjectSpace` by default so that handler stores the object reference
*	Added a `:stop_on_exit` option to thread pools to enable/disable `at_exit` handler
* Updated thread pool docs to better explain shutting down thread pools
* Simpler `:executor` option syntax for all abstractions which support this option
* Added `Executor#auto_terminate?` predicate method (for thread pools)
* Added `at_exit` handler to `TimerSet`
* Simplified auto-termination of the global executors
  - Can now disable auto-termination of global executors
  - Added shutdown/kill/wait_for_termination variants for global executors
* Can now disable auto-termination for *all* executors (the nuclear option)
* Simplified auto-termination of the global executors
* Deprecated terms "task pool" and "operation pool"
  - New terms are "io executor" and "fast executor"
  - New functions added with new names
  - Deprecation warnings added to functions referencing old names
* Moved all thread pool related functions from `Concurrent::Configuration` to `Concurrent`
  - Old functions still exist with deprecation warnings
  - New functions have updated names as appropriate
* All high-level abstractions default to the "io executor"
* Fixed bug in `Actor` causing it to prematurely warm global thread pools on gem load
  - This also fixed a `RejectedExecutionError` bug when running with minitest/autorun via JRuby
* Added `LazyReference`, a simpler and faster varition of `Delay`
  - Updated most internal uses of `Delay` with `LazyReference`
* Moved global logger up to the `Concurrent` namespace and refactored the code

### To-Do

From #248

* [x] Add a new option to the constructor of the Java-backed thread pools that controls the creation of the `at_exit` handler
  - [x] The default will be to *enable* the `at_exit` handler
  - [x] This will support the "simple and easy" base case
  - [x] And will ensure backward compatibility with the current release
* [x] Create an `auto_shutdown?` (or similarly named) predicate method to our Java-backed thread pools that indicates whether the `at_exit` handler has been set
* [x] Update the pure-Ruby thread pools with the *exact* same `at_exit` behavior as the Java-backed thread pools
  - [x] It was an oversight that our pure-Ruby thread pools are not 100% compatible with our Java-backed thread pools
  - [x] However, we need to make sure that this behavior is backward compatible with the current release
* [x] Update the global config to ensure only one `at_exit` handler ever exists for each global thread pool
  - [x] Our default global thread pools should be created without their own `at_exit` handlers
  - ~~[ ] Let the `at_exit` handlers created within the global config handle the global thread pools~~
* [x] Update the behavior of the global configuration `auto_terminate` feature
  - [x] Remove unnecessary `respond_to?` checks
  - ~~[ ] Ensure proper handing of `at_exit` handler creation for all thread pools automatically created within the global config~~
  - ~~[ ] Ensure proper handling of `at_exit` when the user explicitly sets a global thread pool~~
* [x] Create a `Configuration.shutdown_global_thread_pools` (or similarly named) method
  - [x] Give advanced users who disable the `at_exit` handlers the ability to explicitly shutdown the global thread pools
  - [x] Potentially create shutdown/kill/wait methods that mirror the methods on the thread pools themselves
* [x] Create a global "autotest friendly" configuration option
  - ~~[ ] It should include a `require 'concurrent/autotest'` feature for simple setup~~
  - [x] It should prevent any `at_exit` handlers from being created for *all* thread pools, regardless of how they are created
  - [x] Yet it should also ensure that the JVM should not hang because the thread pools are not shutdown
  - [x] And it should be as unobtrusive as possible (I hope to avoid coupling thread pools to some global registry if possible)
* [x] Document, document, document

From #149

* [x] keep them lazy initialized
* [x] rename: task -> short-non-blocking, operation -> long-blocking
* ~~[ ] add more global executors like an immediate executor~~
* [x] remove configuration object, move global pools to `Concurrent` space
* [x] possibly improve parsing of :executor option 
  - ~~[ ] to also handle agent two pools~~
  - [x] to understand :immediate too
  - [x] etc.
* [x] what to do with logging?


jdantonio:
**ALL:** With the exception of documentation, I believe this PR to be complete. I plan to merge it Tuesday night after work. All feedback is welcome.

pitr-ch:
@jdantonio I was on PTO for a week so I am still catching up on the conversations. Could you give me little more time to go over it?

jdantonio:
@pitr-ch Absolutely!

.travis.yml
```diff
@@ -26,4 +26,4 @@ matrix:
     - rvm: jruby-head
     - rvm: 1.9.3
 
-script: "rake compile && bundle exec rspec --color --backtrace --tag ~unfinished --seed 1 --format documentation ./spec"
+script: "rake compile && bundle exec rspec --color --backtrace --tag ~unfinished --seed 1 ./spec"
```
pitr-ch: I think the `--format documentation` was there to see the last test where it was hanging. Is there another way?

lib/concurrent/lazy_reference.rb
```diff
@@ -0,0 +1,62 @@
...
+  #
+  # Because of its simplicity `LazyReference` is much faster than `Delay`:
+  #
+  #            user     system      total        real
+  #     Benchmarking Delay...
+  #        0.730000   0.000000   0.730000 (  0.738434)
+  #     Benchmarking LazyReference...
+  #        0.040000   0.000000   0.040000 (  0.042322)
+  #
+  # @see Concurrent::Delay
```
pitr-ch: Do you have the benchmark source code somewhere? I am surprised that it's that much.

lib/concurrent/lazy_reference.rb
```diff
@@ -0,0 +1,62 @@
...
+    def value
+      # double-checked locking is safe because we only update once
+      return @value if @fulfilled
+
+      @mutex.synchronize do
+        unless @fulfilled
+          begin
+            @value = @task.call
+          rescue
+            @value = @default
```
pitr-ch: The error should be at least logged otherwise it's too hidden and hard to find.

lib/concurrent/lazy_reference.rb
pitr-ch: sorry, never mind, found it.

lib/concurrent/agent.rb
```diff
@@ -14,39 +14,41 @@ module Concurrent
...
     def initialize(initial, opts = {})
       @value                = initial
       @rescuers             = []
       @validator            = Proc.new { |result| true }
       self.observers        = CopyOnWriteObserverSet.new
       @serialized_execution = SerializedExecution.new
-      @task_executor        = OptionsParser.get_task_executor_from(opts)
-      @operation_executor   = OptionsParser.get_operation_executor_from(opts)
+      @io_executor          = get_executor_from(opts) || Concurrent.global_io_executor
+      @fast_executor        = get_executor_from(opts) || Concurrent.global_fast_executor
```
pitr-ch: This removed the option to configure just custom `io` or `fast` executor.

lib/concurrent/agent.rb
```diff
@@ -102,24 +104,24 @@ def validate(&block)
...
     #
-    # @yield the operation to be performed with the current value in order to calculate
+    # @yield the fast to be performed with the current value in order to calculate
     #   the new value
     # @yieldparam [Object] value the current value
     # @yieldreturn [Object] the new value
     # @return [true, nil] nil when no block is given
     def post(&block)
-      post_on(@task_executor, &block)
+      post_on(@io_executor, &block)
```
pitr-ch: This is the other way around in clojure. Default is fast, and post_off is for io. I think we should have sam behavior.

lib/concurrent/lazy_reference.rb
jdantonio: Exception handling is what `Delay` is for. If we start adding features to `LazyReference` it begins to converge with Delay and we loose both the simplicity and the performance. The current features of Delay have grown way beyond the initial use case it was designed to solve, and that's OK. But we still need something that can handle the core use case in a highly performant way. I was able to replace almost every use of Delay within this library with LazyReference as it exists right now, which makes a strong case for the current implementation being spot-on.

lib/concurrent/lazy_reference.rb
jdantonio: :+1:

lib/concurrent/agent.rb
jdantonio: Yep, I missed that. I'll work on it.

lib/concurrent/agent.rb
jdantonio: The intent was to preserve the previous behavior, but the name change proved to be more complicated than I thought it would be. I kept getting confused about which of the previous names mapped to the new names! :-) I'll revisit each of the changes.

lib/concurrent/configuration.rb
```diff
@@ -1,124 +1,296 @@
...
-  # A gem-level configuration object.
-  class Configuration
+  # Suppresses all output when used for logging.
+  NULL_LOGGER = lambda { |level, progname, message = nil, &block| }
 
-    # a proc defining how to log messages, its interface has to be:
-    #   lambda { |level, progname, message = nil, &block| _ }
-    attr_accessor :logger
+  # initialize the global executors
+  class << self
```
pitr-ch: What's the reason behind `class << self` here? What effect does it have on `@@class_var`s ?

lib/concurrent/configuration.rb
```diff
@@ -1,124 +1,296 @@
...
+    end
+
+    # @!visibility private
+    @@global_timer_set = LazyReference.new do
+      TimerSet.new(stop_on_exit: @@auto_terminate_global_executors.value)
+    end
+  end
+
+  def self.global_logger
+    @@global_logger.value
```
pitr-ch: I think class variables are considered bas style, AFAIK also they are slower then constants and instance vars in MRI. If the intention is to hide, `@ivar` might be better otherwise constants are fine.

lib/concurrent/configuration.rb
jdantonio: To get the appropriate context for the class variables which need to be access from within the class methods. This code runs when the file is first loaded by Ruby, this ensuring that all those objects get created when the library loads. If we didn't wrap this code in the `self << class` block the class variables would not have been the same as what is being referenced in the class methods.

lib/concurrent/executor/executor.rb
```diff
@@ -59,6 +59,38 @@ def handle_fallback(*args)
...
+        if Concurrent.auto_terminate_all_executors?
+          this = ObjectSpace._id2ref(id)
+          this.kill if this
+        end
+      end
+    end
+
+    def create_java_at_exit_handler!(this)
+      at_exit do
+        this.kill if Concurrent.auto_terminate_all_executors?
```
pitr-ch: Won't this leak the pools?

lib/concurrent/configuration.rb
```diff
@@ -1,124 +1,296 @@
...
+    # @!visibility private
+    @@auto_terminate_global_executors = AtomicBoolean.new(true)
+
+    # @!visibility private
+    @@auto_terminate_all_executors = AtomicBoolean.new(true)
+
+    # @!visibility private
+    @@global_fast_executor = LazyReference.new do
+      Concurrent.new_fast_executor(
+        stop_on_exit: @@auto_terminate_global_executors.value)
```
pitr-ch: This may behave unexpectedly when the pool is initialized before the auto-termination is disabled.

lib/concurrent/configuration.rb
jdantonio: If there is a significant performance difference we can look at constants, but this approach hides direct access to the wrapper object itself. With this approach a user must go through the accessor methods we define, which hides the implementation details, which in turn gives us a greater ability to refactor in the future. The user should not need to know that a global thread pool is wrapped inside of a `LazyReference` object and also know the API of that object. The user should simply be able to post to the global thread pool.

lib/concurrent/configuration.rb
```diff
@@ -1,124 +1,296 @@
...
+  #   ensure that they are shutdown when the application ends. When
+  #   changed to false, the `at_exit` handlers will be circumvented
+  #   for *all* Concurrent Ruby thread pools running within the
+  #   application. Even those created within other gems used by the
+  #   application. This method should *never* be called from within a
+  #   gem. It should *only* be used from within the main application.
+  #   And even then it should be used only when necessary.
+  def self.disable_auto_termination_of_all_executors!
+    @@auto_terminate_all_executors.make_false
+  end
```
pitr-ch: there is possibly the counterpart missing.

lib/concurrent/configuration.rb
```diff
@@ -1,124 +1,296 @@
...
+      idletime:        60,          # 1 minute
+      max_queue:       0,           # unlimited
+      fallback_policy: :caller_runs # shouldn't matter -- 0 max queue
+    )
+  end
+
+  def self.new_io_executor(opts = {})
+    ThreadPoolExecutor.new(
+      min_threads: [2, Concurrent.processor_count].max,
+      max_threads: Concurrent.processor_count * 100,
```
pitr-ch: we should consider same default as on java pool http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/concurrent/Executors.java#Executors.newCachedThreadPool%28%29, unlimited threads. To prevent deadlock when all available tasks are blocked by work waiting in the queue, etc. I think if there is too many work being posted the the pool it needs to be contained differently.

lib/concurrent/configuration.rb
jdantonio: Missing on purpose. It's on by default and *most* users will probably not have the need/knowledge to change it. Once an advanced user turns it off, I didn't see any reason they would need to turn it back on. So I considered the counterpart method to be YAGNI.

lib/concurrent/configuration.rb
jdantonio: Absolutely! I didn't know what Java's defaults were when I first introduced the global thread pools so I guessed. The Java team are a pretty smart group so I'm always happy to lean on their expertise.

lib/concurrent/lazy_reference.rb
pitr-ch: yeah I understand, sounds good! Still when there is bug/exception in the block it is not even logged which makes it very hard to find. Personally I would raise it on every attempt the read the value, to be able to hunt the bug faster.

jdantonio: That's reasonable. We can definitely extend the API. I would just like to be very careful about keeping it as simple and performant as possible.

lib/concurrent/lazy_reference.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    # The calculated value of the object or the default value if one
+    # was given at construction. This first time this method is called
+    # it will block indefinitely while the block is processed.
+    # Subsequent calls will not block.
+    #
+    # @return [Object] the calculated value
+    def value
+      # double-checked locking is safe because we only update once
+      return @value if @fulfilled
```
pitr-ch: I don't think this is fine. @fullfilled may be seen as true before the @value is also made readable. As it is AFAIK it'll work on MRI because of GIL and on JRuby because `@ivars` have volatile access (but both is undocumented behavior). It'll probably break on Rubinius.

lib/concurrent/configuration.rb
pitr-ch: I've tried to remove the `self << class` and it works too. Removing the class variables may be better to avoid the problems with scope definition. They always felt weird to me.

lib/concurrent/configuration.rb
pitr-ch: In that case `@ivar` are better choice or private constants (not sure when they were introduced though).

lib/concurrent/lazy_reference.rb
pitr-ch: It should be possible by second check which would keep the fast path untouched:
```ruby
return @value if @fulfilled
raise @value if @failed
```
or something like that

lib/concurrent/delay.rb
```diff
@@ -81,7 +91,8 @@ def reconfigure(&block)
 
     private
 
-    def execute_task_once
+    # @!visibility private
```
pitr-ch: is it needed when it's already a private method?

lib/concurrent/executor/executor.rb
```diff
@@ -59,6 +59,38 @@ def handle_fallback(*args)
...
+      end
+    end
+
+    def create_ruby_at_exit_handler!(id)
+      at_exit do
+        if Concurrent.auto_terminate_all_executors?
+          this = ObjectSpace._id2ref(id)
+          this.kill if this
+        end
+      end
```
pitr-ch: It might be better to use a single register of executors scheduled for termination (as mentioned). This is leaking at_exit blocks.

lib/concurrent/executor/executor_options.rb
```diff
@@ -0,0 +1,52 @@
...
+    # @!visibility private
+    def get_executor_from(opts = {}) # :nodoc:
+      if (executor = opts[:executor]).is_a? Symbol
+        case opts[:executor]
+        when :fast
+          Concurrent.global_fast_executor
+        when :io
+          Concurrent.global_io_executor
+        when :immediate
+          Concurrent::ImmediateExecutor.new
```
pitr-ch: Better to also have a single global `ImmediateExecutor` to avoid unnecessary object creation.

lib/concurrent/executor/executor.rb
```diff
@@ -59,6 +59,38 @@ def handle_fallback(*args)
...
     end
+
+    def auto_terminate?
+      !! @auto_terminate
+    end
+
+    protected
+
+    def enable_at_exit_handler!(opts = {})
+      if opts.fetch(:stop_on_exit, true)
```
pitr-ch: I would argue that custom thread creation is more in a land of advanced users so it might may sense to make the default `false`. Hm, but that may be subjective.

lib/concurrent/executor/executor_options.rb
pitr-ch: It could also be good to have SingleThreadExecutor and similar executors without configuration.

pitr-ch:
Thanks for all the work on executors! And sorry that it took me so long to get to it :/

lib/concurrent/lazy_reference.rb
pitr-ch: With locking LazyReference is about 3.5x faster than Delay.

jdantonio: I had a chance to spike some of your suggestions last week while pair programming at a meetup but haven't had an opportunity to work on it since. We ran the benchmarks without the DCL and got the same result you did. That's still a very significant improvement so I plan to remove the check on line 46.

pitr-ch: BTW you may find this interesting http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

pitr-ch: and +1 for keeping for now. But I would like to explore options how to make the `Delay` faster.

jdantonio: I'm all for making Delay faster, but that's not the primary reason I created LazyReference. I created LazyReference because the basic use case for Delay is no longer easily or intuitively supported. The initial use case for Delay was to lazy load a resource in such a way that the first call to `value` would block the current thread *indefinitely* and execute the block *on that thread*:

```ruby
# original implementation
d = Concurrent::Delay.new{ sleep 10; 'foo' } # no constructor parameters
d.value # blocks the current thread and executes the block on the current thread
```

The current behavior of Delay is different:

```ruby
# current implementation
d = Concurrent::Delay.new{ sleep 10; 'foo' } # no constructor parameters
d.value # blocks the current thread and executes the block on a *global thread pool*
```

Notice that the current version of Delay blocks *2* threads: The current thread and the thread on the global thread pool that executes the block. This is very different behavior and also the source of the bug with the root actor (which we originally thought was a problem with autotest).

To get the same behavior as the *original* Delay implementation we must pass constructor arguments:

```ruby
# current implementation
d = Concurrent::Delay.new(executor: Concurrent::ImmediateExecutor.new){ sleep 10; 'foo' } # passing an executor into the constructor
d.value # blocks the current thread and executes the block on the current thread
```

This syntax is not every obvious nor very intuitive and is easy to get wrong. If we, the authors of this gem, got it wrong with the root actor, it's reasonable to assume that our users will get it wrong in critical situations, too.

What's most interesting about this base case is that every use of Delay within this gem follows the ImmediateExecutor pattern. Every one. Moreover, I was able to replace every use of Delay within this gem with LazyReference without losing functionality (with the exception of the root actor, which remains a Delay). That strongly indicates that this base use case is the most important use case and, therefore, the use case that should be supported by the simplest use of the abstraction. LazyReference does this, Delay does not.

pitr-ch: I think `Delay` should be by default executed on `immediate` executor. I would say that it's a bug that it does not (and it is probably mine, sorry). 

I've just tried tweaking the new delay implementation in the unifying PR I got it down to 0.71sec vs 0.61sec (LazyReference). Which probably means that the current delay could be also optimized, hence I would consider removing LazyReference in favor of optimized (current) Delay with default immediate executor (,or keep using current Delay because we can be reasonable confident that we can get the times down later). (Meant in scope for next release)

jdantonio: @pitr-ch Most of the changes to Delay came from you (as of 0.8.0 my only commits were the initial implementation). I assumed that the behavior change was intentional and something you felt was necessary, which is why I created LazyReference rather than change Delay. My preference is to have only once class but I didn't want to undo that facet of your work. I'm comfortable calling the current behavior a bug, setting the default behavior (when no arguments are passed to the constructor) to immediate execution, and removing LazyReference.

I'm working on several updates based on some of your feedback and suggestions, but haven't had much time this week. I plan to work on it more this weekend and also respond to several of your other comments. Can you please push your optimized Delay code up to GitHub so I can take a look? I don't see any recent commits on #176.

pitr-ch: @jdantonio I'll look at the delays what ca be done there.

I've just manually inlined methods on the critical path (`value` method) to remove nested locking. I'll require bigger changes for it to be usable on all methods without manual inlining (which would be hell to maintain). 

```ruby
# Future#value
def value(timeout = nil)
  wait timeout
  synchronize { @value }
end
# changed to this
def value(timeout = nil)
  synchronize do
    unless @touched
      @touched = true
      @promise.touch
    end
    @condition.wait @mutex, timeout if @state == :pending
    @value
  end
end
```

I plan to decouple the synchronization and behavior into different methods which should allow inheritance for the classes to work and to avoid unnecessary nested locking like it's now in concurrent-ruby and also in #176.

.travis.yml
jdantonio: I'm not sure if there's another way, but `--format documentation` has been in .travis.yml since [June of last year](https://github.com/ruby-concurrency/concurrent-ruby/commit/8672425108802baf816cd004f94b58c8f4c11191#diff-354f30a63fb0907d4ad57269548329e3). I only removed it a couple of weeks ago--then immediately added it back when we had a hanging test. I think we should keep it for now. Once we stabilize our tests we can remove it, but the full output is pretty valuable right now.

lib/concurrent/agent.rb
jdantonio: I looked back at the earlier code and the current implementation (`post` uses `io_executor`, `post_off` uses `fast_executor`) matches the implementation in 0.8.0. I dont' have a strong preference either way so I'll change our implementation to match Clojure.

lib/concurrent/configuration.rb
jdantonio: I was not aware that constants could be made private. I did some Googling and some experimentation. It looks like `private_constant` was introduced in 1.9.3 but [isn't documented](http://ruby-doc.org/core-1.9.3/Module.html). Both Rbx and JRuby seem to support it. I definitely like that option better.

lib/concurrent/executor/executor.rb
jdantonio: Probably, but I don't know is that's avoidable. But I also can't image an application creating more than a handful of thread pools. I would expect that most would simply use the global thread pools and those are going to live until the application exits. The original implementation used `self` within the `at_exit` block but I don't think that was safe.

lib/concurrent/configuration.rb
jdantonio: The `at_exit` handler checks the auto-termination setting again. The best practice should be to set auto-termination at application start, before the pools warm up. Doing so will prevent the `at_exit` handlers from being created. Should the user disable auto-termination *after* one or more pools warm up, the `at_exit` handler will behave as expected because it checks again. So it's safe either way, but slightly more efficient in the former case.

lib/concurrent/delay.rb
jdantonio: It's not needed, but I like being explicit about intent.

lib/concurrent/executor/executor.rb
pitr-ch: There may be long running applications which may be creating pool for periodic cpu-heavy processing or something. Then it would slowly consume memory. I think we cannot release it like this.

lib/concurrent/configuration.rb
pitr-ch: I think the problem lies only in the global executors, they check again the `auto_terminate_all_executors` flag but not `auto_terminate_global_executors`.

Issue #256, opened 2015-02-27
Replace `Time.now` with new monotonic clock


jdantonio:
Monotonic clock support was added in PRs #253 and #254. There are numerous placers where we use `Time.now` in our code. We should replace as many of those as possible with monotonic clocks. The current `clock_time` implementation within `Condition` should be moved to the `Concurrent` namespace and made public. Each of the following classes should then be updated, if possible:

- ~~Agent (used for observation)~~
- [x] Condition
- [x] RubyThreadPoolExecutor
- [x] RubyThreadPoolWorker
- [x] TimerSet
- ~~IVar (used for observation)~~
- [x] ScheduledTask
- ~~[x] TimerTask (uses `timer` for scheduling and Time.now for observation~~

pitr-ch:
I would treat this as a bug, it may cause ugly/hard-to-find bugs.

jdantonio:
Unless there is some way to make guarantees on systems without a monotonic clock, I don't see that we *can* consider it a bug. It's a limitation in Ruby that we can't do anything about. I plan to write extensive documentation explaining possible clock problems and attach that documentation to all abstractions within this library that use the clock, but I think that's the best we can do.

I'm open to suggestions for alternate solutions.

pitr-ch:
We may need to do more C coding :/ borrowing ideas from https://github.com/copiousfreetime/hitimes AFAIK the gem uses monotonic clock.

jdantonio:
Writing C code to do this will only help a limited audience. It would provide better support for MRI 1.9.3 and 2.0, but only when the C extension gem is installed. It won't help Rubinius users. Support for Ruby 1.9.3 [has ended](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/) so we're mostly talking about MRI 2.0 and Rubinius. As much as I would like to offer a perfect solution, this is a problem that eventually goes away as people upgrade to the newer versions of Ruby.

It should be fairly easy to copy the [C code from MRI](https://github.com/ruby/ruby/blob/trunk/process.c) and create a pure-C fallback so I'll look into that. But we'll still  have to document that *some* users may have problems in certain edge cases.

pitr-ch:
You are right I forgot that we need also a pure-Ruby solution. In that case we could do a slow but correct implementation in pure-Ruby. Pseudo implementation follows
```ruby
synchronize do
  @correction ||= 0 # compensating any back time shifts
[...]

PR #258, opened 2015-03-01
Clock time

jdantonio:
Addresses #256 

* Use monotonic clock for all timers via `Concurrent.monotonic_time`
  - Use `Process.clock_gettime(Process::CLOCK_MONOTONIC)` when available
  - Fallback to `java.lang.System.nanoTime()` on unsupported JRuby versions
  - Pure Ruby implementation for everything else
  - Effects `Concurrent.timer`, `Concurrent.timeout`, `TimerSet`, `TimerTask`, and `ScheduledTask`
* Deprecated all clock-time based timer scheduling
  - Only support scheduling by delay
  - Effects `Concurrent.timer`, `TimerSet`, and `ScheduledTask`


lib/concurrent/utility/clock_time.rb
```diff
@@ -0,0 +1,18 @@
...
+    def clock_time
+      java.lang.System.nanoTime() / 1_000_000_000.0
+    end
+  else
+    def clock_time
+      Time.now.to_f
+    end
+  end
+
+  module_function :clock_time
```
pitr-ch: `monotonic_time` may be describing the method better.

lib/concurrent/utility/monotonic_time.rb
```diff
@@ -0,0 +1,147 @@
...
+          if @last_time < corrected_now
+            @last_time = corrected_now 
+          else
+            @correction += @last_time - corrected_now + 0.000_001
+            @last_time = @correction + now
+          end
+        }
+      end
+    end
+  }.new
```
pitr-ch: -   minor: there should be `do end` instead of `{}` for multiline blocks
-   Why is the class hidden and anonymous? That seems very unusual. 
-   Why usual pattern for different implementations is not followed? (`JavaMonotonicClock` etc.)

lib/concurrent/utility/monotonic_time.rb
```diff
@@ -0,0 +1,147 @@
...
+    COUNT.times{ CLOCK.get_interval_native(native_now) }
+  end
+
+  puts "Ruby interval..."
+  bm.report do
+    COUNT.times{ CLOCK.get_interval_ruby(ruby_now) }
+  end
+end
+
+puts "Native: #{CLOCK.get_time_native}, Ruby: #{CLOCK.get_time_ruby}"
```
pitr-ch: Is the big comment here with benchmark left intentionally?

jdantonio: I left it there so that others could see it an reproduce it, but I planned to remove it eventually.

I wanted to merge the PR because it effected several classes for which we had intermittently failing tests. I felt it was better to work on those tests with this code in place. But I knew that you and others might not have time to review the PR before I merged it so I left the benchmark data in the code so that others could see it.

lib/concurrent/utility/monotonic_time.rb
jdantonio: I experimented with a couple different implementations of this and didn't like any of them. The only reason this code exists is because not all runtimes support `Process.clock_gettime(Process::CLOCK_MONOTONIC)`. At ope point I had a set of classes that follow our normal convention but I couldn't think of a single use case for them. I can't think of a single reason a user would need to create an object of this type. All anyone should ever need is `Concurrent.monotonic_time`. The class/object underneath that is just an implementation detail. In an early version I only created `GLOBAL_MONOTONIC_CLOCK` on systems which needed it. I didn't like having a global variable that wasn't available on all platforms so I implemented it universally. But I also wanted to discourage users from creating new instances of the class. A single clock is all that's necessary. So I made it anonymous and hidden from the yardoc.

I agree that it's an odd implementation, but it's also a very odd feature--one that should go away over time as people move to newer versions Ruby.

If the internal inconsistency (with respect to class name and hierarchy) bothers people I don't have a problem with us changing it. In that case I will still want to hide the classes from yardoc. And also add a notice that the class shouldn't be used directly, instead recommending only the use of `Concurrent.monotonic_time`.

jdantonio: The reason I used `{}` for that block is that I really dislike calling methods on an `end` statement. The natural replacement for that line would be `end.new` and that syntax always seems difficult to read. Since the convention of using `begin/end` for multiline blocks is just that--a convention--and not universally accepted by all Rubyists, I went with the syntax that made the most sense to be in this case.

If the team wants to develop a style guide for this gem (or adopt one created by another project) I'm happy to support that. I've never rejected a PR based on style, nor have I ever rejected a PR because someone changed the style of existing code. It's just something that's never been a high priority for me.

pitr-ch: I see your point. Nevertheless It already happen to me more than once that I though there is no use-case for something, but there was. For that reason I think it's better to assume that the use-case is just not obvious. In this case testing might be the reason to provide the class, to be able to create new instance injecting mocked Time, then messing with the clock testing that the monotonic condition holds, without interfering with the global instance in the gem.

lib/concurrent/utility/monotonic_time.rb
pitr-ch: I think it's valuable to keep it, could you move it to a separate file please?

Issue #146, opened 2014-08-09
Provide a read-write lock


alexdowad:
Just discovered this project; it looks like you guys have put together a very nice suite of utilities.

It looks like you don't have a read-write lock yet (one which allows multiple concurrent readers, but only one concurrent writing). A couple years ago, I needed one for a Ruby project and couldn't find anything good, so I developed my own implementation. Here it is: https://github.com/alexdowad/showcase/blob/master/ruby-threads/read_write_lock.rb

I put a fair amount of time into testing and verifying this implementation, and also had a couple of other Ruby developers do code review and test it on their machines as well. It has worked in production for me, without any problems detected. So while more review would be good, it is already somewhat mature.

A RW lock which "spins" when the lock is taken would be lighter (less memory footprint per lock), but is more "dangerous" -- the user needs to know what they are doing when they select such a lock.

If you would like to use this RW lock, please take it and distribute it under your existing license conditions. Of course, I may also use it in other places and distribute it under other licenses.

PR #259, opened 2015-03-04
ReadWriteLock donated by @alexdowad

jdantonio:
Merges `ReadWriteLock` donated by @alexdowad in #146.

* Copied the source file to the `lib/concurrent/atomic` directory
* Moved the class under the Concurrent namespace
* Moved the stress tests to `examples/benchmark_read_write_lock.rb`
* Updated the stress tests to use `OptionsParser`
* Created a spec file

To-do

- [x] Yardoc
- [x] Tests
- [x] Refactor
- [x] Changelog
- [x] Readme


alexdowad:
@jdantonio, thank you very much for cleaning up my code!

jdantonio:
@alexdowad No clean up was necessary. Your code is awesome! All I had to do was integrate it into the project. Thank you very much for donating it!

lib/concurrent/atomic/read_write_lock.rb
```diff
@@ -0,0 +1,168 @@
...
+      @writer_q     = ConditionVariable.new # queue for waiting writers
+      @writer_mutex = Mutex.new             # to protect writer queue
+    end
+
+    def with_read_lock
+      acquire_read_lock
+      result = yield
+      release_read_lock
+      result
+    end
```
pitr-ch: better to wrap `release_read_lock` in `ensure` block

lib/concurrent/atomic/read_write_lock.rb
```diff
@@ -0,0 +1,168 @@
...
+
+    def with_write_lock
+      acquire_write_lock
+      result = yield
+      release_write_lock
+      result
+    end
```
pitr-ch: ditto

lib/concurrent/atomic/read_write_lock.rb
```diff
@@ -0,0 +1,168 @@
...
+      while(true)
+        c = @counter.value
+        raise "Too many reader threads!" if (c & MAX_READERS) == MAX_READERS
+
+        # If a writer is waiting when we first queue up, we need to wait
+        if c >= WAITING_WRITER
+          # But it is possible that the writer could finish and decrement @counter right here...
+          @reader_mutex.synchronize do 
+            # So check again inside the synchronized section
+            @reader_q.wait(@reader_mutex) if @counter.value >= WAITING_WRITER
```
pitr-ch: This looks weird I need to revisit it later. Shouldn't be `@counter` writes be also protected by this mutex? Otherwise the `@counter` can be updates also inside this `synchronize` block.

pitr-ch:
@alexdowad Thanks for the code!

lib/concurrent/atomic/read_write_lock.rb
alexdowad: It doesn't matter if `@counter` is updated elsewhere when a reader is in this `synchronized` block. `@counter` is always updated with an atomic CAS, so updates to `@counter` can never be lost.

The point of this line is so that a reader doesn't sleep forever if it tries to acquire the lock at the moment a writer is releasing it. (Remember, the only thing that can make a reader sleep is if a writer already holds the lock. Readers don't block readers.)

Now, when a writer releases the lock, it acquires `@reader_mutex` before waking up the waiting readers. If the lock-acquiring reader gets into this `synchronized` block before the lock-releasing writer, then the reader will go to sleep and atomically release `@reader_mutex`, then the leaving writer will get into its `synchronized` block and wake the reader up. On the other hand, if the leaving writer gets into its `synchronized` block first, `@counter` will already be updated and visible to all threads by that time the reader gets into its `synchronized` block, the condition will be false and it will not sleep. In either case, the reader will not sleep forever.

pitr-ch: Ah I see. Not obvious at first glance. Is the implementation based on some other implementation or paper? I would like to have a link for the documentation.

pitr-ch:
The complex conditions could be extracted to methods to improve readability. E.g.
```ruby 
def readers(c = @count.value)
  c & MAX_READERS
end

def waiting_writer?(c = @counter.value)
  c >= WAITING_WRITER
end

# etc
```

lib/concurrent/atomic/read_write_lock.rb
alexdowad: The original `read_write_lock.rb` credited Doug Lea (author of Java's `ReentrantReadWriteLock`) for inspiration -- those credits have been removed.

This implementation doesn't follow `ReentrantReadWriteLock` 100%. It was based simply on the idea of encoding state information in an integer, atomically-swapping the integer to "claim" the lock (or the right to be woken up when the lock is available), and sleeping on a condition queue when the lock is not available. Additionally, a goal from the very beginning was that the "fast path" should be lock-free and only use a CAS on entry and one on exit.

The details of the code flowed naturally out of those ideas.

alexdowad: I might add, the only details which do *not* flow directly from the above ideas are: when both readers and writers are queued, and the lock becomes available, who should be given the first chance to acquire it? If a reader wants to acquire the lock, but discovers writers are waiting, should it try to "grab" the lock before any of the writers do, or let them go ahead?

I didn't get those decisions right the first time; experimentation was required to achieve both good throughput and fairness (avoiding starvation).

pitr-ch: Great thanks for the info. That means that we may be able to use `ReentrantReadWriteLock` on JRuby.

alexdowad: Yes -- have a look at the benchmark code in this PR. (On JRuby, I believe I benchmark against `ReentrantReadWriteLock`.)

jdantonio: I'll add the reference to `java.util.concurrent.ReentrantReadWriteLock` in my next commit.

lib/concurrent/atomic/read_write_lock.rb
jdantonio: My next commit will have an `ensure` block that releases the lock but is also sensitive to errors raised *before* the lock is acquired (such as when the maximum number of readers/writers is reached).

pitr-ch: right, it has to be:
```ruby
def with_read_lock
  acquire_read_lock
  begin
    yield
  ensure
    release_read_lock
  end
end
```
not just ensure at the end, good catch

jdantonio:
I believe this PR is ready to merge. It has nearly 100% code coverage, full yardoc, better exception handling, a few readability updates (methods for some of the bitwise operations used in conditionals), and the stress tests still pass with comparable performance.

PR #262, opened 2015-03-06
Fix Intermittently Failing Tests

### LINKED ISSUE #117: Intermittently Failing Tests

jdantonio:
Addresses #117 and #263

### DIFF
--- a/lib/concurrent/utility/monotonic_time.rb
+++ b/lib/concurrent/utility/monotonic_time.rb
@@ -55,6 +55,22 @@ module Concurrent
     GLOBAL_MONOTONIC_CLOCK.get_time
   end
   module_function :monotonic_time
+
+  # Runs the given block and returns the number of seconds that elapsed.
+  #
+  # @yield the block to run and time
+  # @return [Float] the number of seconds the block took to run
+  #
+  # @raise [ArgumentError] when no block given
+  #
+  # @!macro monotonic_clock_warning
+  def monotonic_interval
+    raise ArgumentError.new('no block given') unless block_given?
+    start_time = GLOBAL_MONOTONIC_CLOCK.get_time
+    yield
+    GLOBAL_MONOTONIC_CLOCK.get_time - start_time
+  end
+  module_function :monotonic_interval
 end
 
 __END__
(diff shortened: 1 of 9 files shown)


lib/concurrent/utility/monotonic_time.rb
```diff
@@ -55,6 +55,22 @@ def monotonic_time
...
+  #
+  # @raise [ArgumentError] when no block given
+  #
+  # @!macro monotonic_clock_warning
+  def monotonic_interval
+    raise ArgumentError.new('no block given') unless block_given?
+    start_time = GLOBAL_MONOTONIC_CLOCK.get_time
+    yield
+    GLOBAL_MONOTONIC_CLOCK.get_time - start_time
+  end
```
pitr-ch: This method may not return correct results on MRI if there was time shift, it may be better kept just in test helper.

spec/concurrent/atomic/condition_spec.rb
```diff
@@ -32,37 +27,80 @@ module Concurrent
...
+            t = Thread.new do
+              mutex.synchronize do
+                latch_1.count_down
+                subject.wait(mutex)
+                latch_2.count_down
+              end
+            end
+
+            latch_1.wait(1)
+            latch_2.wait(0.1)
```
pitr-ch: It might be better practice to use Events in similar cases giving the variable some name for improved readability. 
```ruby
  before_waiting = Concurrent::Event.new
  after_waiting = Concurrent::Event.new
  mutex = Mutex.new

  thread = Thread.new do
    mutex.synchronize do
      before_waiting.set
      subject.wait(mutex)
      after_waiting.set
    end
  end

  before_waiting.wait(1)
  after_waiting.wait(0.1)
  expect(after_waiting.set?).to be_false
  thread.kill
```
But both latch and event cannot be used here because both depend on the `Condition` which is being testes here. `Queue` may be good replacement here.

Issue #263, opened 2015-03-07
Intermittently Failing Tests


jdantonio:
A running list of tests that occasionally fail and need to be fixed:

- [x] [Concurrent::Exchanger exchange without timeout can be reused](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/53489448)
- [x] [Concurrent::ScheduledTask instance #execute does nothing unless the state is :unscheduled](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/53586321)
- [x] [dataflow passes the values of dependencies into the block](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/53587463)
- ~~Exchanger tests deadlock and [never finish](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/55047572)~~ *Known bug. Will be rewritten from scratch after the 1.0 release.*
- ~~[Concurrent::ReadWriteLock#to_s includes the waiting writer count](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/55047574) *(NOTE: ReadWriteLock has not been released and a custom `to_s` method is not critical. Deleting the method is a viable option.)*~~
- [x] #timeout kills the thread on timeout
- [x] #timeout kills the thread on success
- [x] Concurrent::MVar spurious wake ups #modify returns TIMEOUT on timeout on an empty MVar
- [x] [Concurrent::TimerTask dereferenceable it should behave like dereferenceable defaults :dup_on_deref to false](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/55237542)
- [x] [Concurrent::CyclicBarrier#wait without timeout should block the thread](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/55237542)
- [x] [Concurrent::TimerSet executes a given task when given a Time](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/55874348)
- [x] "Concurrent::Condition with many waiting threads signaled wake up broadcast wakes up all threads" not able to reproduce on os x rbx-2.5.2
- [x] [Concurrent::Edge futures .schedule scheduled execution in graph](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/60946747)
- [x] [Concurrent::TimerTask dereferenceable it should behave like dereferenceable defaults :dup_on_deref to false](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/61219215)
[...]

PR #264, opened 2015-03-07
Fix Intermittently Failing Tests

jdantonio:
Addresses #263


examples/benchmark_lazy_and_delay.rb
```diff
@@ -15,9 +15,8 @@
...
 
-Benchmark.bm do |x|
-  puts 'Benchmarking Delay...'
-  x.report { n.times{ delay.value } }
-  puts 'Benchmarking Lazy...'
-  x.report { n.times{ lazy.value } }
+Benchmark.bmbm do |bm|
+  puts 'Benchmarking Delay first, then LazyReference...'
+  bm.report { n.times{ delay.value } }
+  bm.report { n.times{ lazy.value } }
```
pitr-ch:
```ruby
Benchmark.bmbm 10 do |m|
  bm.report('delay') { n.times { delay.value } }
  bm.report('lazy') { n.times { lazy.value } }
end
```
should help

PR #268, opened 2015-03-21
Edge Gem

jdantonio:
## In Progress -- Do Not Merge


pitr-ch:
Thanks, I'll submit few updates later along with the new Promises work.

jdantonio:
@pitr-ch Cool. Thank you!

I may move `Exchanger` into the edge gem, too. I'm not sure if you've seen the conversation between @chrisseaton and me in #271, but we have a race condition in `Exchanger` that we may not be able to fix easily.

pitr-ch:
@jdantonio yeah, I think it makes sense to move there all things that we are not sure about. We can always easily move them back if needed.

pitr-ch:
Continues in #274

jdantonio:
Merged into #274.

PR #269, opened 2015-03-24
Allow root promise body to be set later

pitr-ch:
Comments, suggestions? 

Please do not merge, testing with Dynflow first.

Issue #257, opened 2015-03-01
Final 1.0 Roadmap


jdantonio:
It's been suggested (#246) that we're ready for a 1.0 release. People are using this gem in production and the Rails team is considering making `concurrent-ruby` a dependency of Rails (#251). I believe we've reached a level of maturity, stability, and performance that warrants a 1.0 release. Our community deserves it. So with that in mind I would like to set a final 1.0 roadmap. Because of the major refactoring of the global configuration object (#255) being undertaken--and the number of method calls this will deprecate--I want to have one more major release before 1.0. The goals for these releases will be:

* **0.9** Stabilize the API and fix known bugs
* **1.0** Performance, stability, test cleanup, and documentation
* **Post 1.0** Major functionality and API updates

## v0.9 (codename [Ultron](http://marvel.com/avengers))

*Target Date: 18 May 2015*

This goal of this release is to stabilize the API. The core update in this release is the refactor of the global configuration object and the renaming of the global thread pools (#255). Other features are merged/in-progress updates (monotonic clock) and bug fixes.

- [x] Monotonic clock updates (#256 -- @jdantonio)
- [x] Gem configuration and global thread pool refactor (#255, #248, #227, #149 -- @jdantonio)
- ~~Depend on `ref` gem [version 2.0](https://github.com/ruby-concurrency/ref/issues/16) (@lucasallan)~~
- [x] Read-write lock (#146 -- @jdantonio)
- [x] Bug fixes
  - [x] `unlock`s outside of ensure block (#244 -- @jdantonio)
  - [x] Atomic get_and_set faulty in C (#233 -- @jdantonio)
- [x] Actor updates (@pitr-ch)
- [x] Update all dev/test gem dependencies to latest versions
- [x] `Promise` becomes subclass of `IVar` (@jdantonio, @pitr-ch)

## v1.0.pre1 (codename [Fury Road](http://www.madmaxmovie.com/))

*Target Date: 29 May 2015*

The goal of this pre-release is to clean up everything that is being removed and create the first build of the [edge](https://github.com/ruby-concurrency/concurrent-ruby/pull/268) gem. Pre-release and release candidate builds of the edge gem will occur simultaneously with pre-release and release candidate builds of the core gem. Only the version number will be different.

- [x] Remove *all* deprecated code
- [x] Move `Actor` and `Channel` to the edge gem
- [x] Merge thread_safe gem (#230 -- @jdantonio)
[...]

Issue #244, opened 2015-02-13
`unlock`s outside of ensure block


pitr-ch:
there is several `unlock` calls outside of ensure blocks in this file, see https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/atomic/copy_on_notify_observer_set.rb#L46

jdantonio:
Fixed in PR #270

Issue #139, opened 2014-07-30
Consolidate IVar childen into one object


pitr-ch:
IVar, Future, Promise, Probe have a lot in common. I think we could explore an way how to merge some or more into one class.

jdantonio:
I suggest we close this discussion and the associated PR #176 and possibly revisit in the future when we begin planning a 2.0 release.

These discussions have been a very interesting thought exercise, but they are all speculative. These changes would break many existing APIs, many of which have been in place since the earliest days of this gem (long before I invited additional maintainers). There is a great deal of API consistency because of `Dereferenceable`, `Obligation`, and `Observable`. The existing APIs are in use in production and seem to make our users happy. No one who is using this gem in production has *requested* these changes, but they *have* requested a stable 1.0 release.

I believe the best path forward for our users is to stabilize and optimize the gem with the current APIs and release 1.0.

jdantonio:
I've been traveling and otherwise very busy lately so I hadn't seen [this comment](https://github.com/ruby-concurrency/concurrent-ruby/issues/257#issuecomment-78858991) from @pitr-ch when I made my [previous comment](https://github.com/ruby-concurrency/concurrent-ruby/issues/139#issuecomment-84319084). So I wasn't aware that some of these ideas came from work being done in another project. We should definitely take that into account.

iNecas:
Yes, as @pitr-ch already mentioned, the lack of reasonable chaining (the dataflows would seem to workaround the issue, but the price of a thread needed to be allocated is too high) is quite painful for us. In the previous implementation, we had this https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/future.rb#L85-L92, and it was quite nice to work with that API.

In the https://github.com/Dynflow/dynflow/pull/139, which is taking us from our custom implementations of futures and actors to concurrent-ruby, we hit several places where the original approach was needed, but IVar (which we choose as replacement for our custom-made future) were lacking the feature. For this, we needed either to rework some of our implementations to workaround this, or came with suboptimal alternatives with using dataflows https://github.com/iNecas/dynflow/commit/6327a834252c5c96e9ee33f07d508289f2929497. This commit is there just to make our tests green, but the whole PRs is not going to be merged until we have proper solution, that doesn't require a thread for every such a operation.

[...]

PR #270, opened 2015-03-25
Promise extends IVar

jdantonio:
## Promise extends IVar

`Promise` extends `IVar` the way `Future` does. This allows a Promise to be manually (and safely) be set from any thread. It also begins the process of consolidating Promise, Future, and IVar.

See #257, #139, and #269

## IVar is a Synchronization::Object

`IVar` is now a `Synchronization::Object`.

* `Dereferenceable` supports injection of a mutex
* `Obligation` supports injection of a mutex
* `IVar` extends `Synchronization::Object`
* `IVar` passes `self` to `Obligation` as injected mutex
* `IVar`, `Promise`, and `Future` no longer call `mutex#lock`

See #273 and #275

## Moved mutex unlocking into ensure clauses.

See #244

### DIFF
--- a/lib/concurrent/delay.rb
+++ b/lib/concurrent/delay.rb
@@ -52,9 +52,20 @@ module Concurrent
 
     # Create a new `Delay` in the `:pending` state.
     #
-    # @yield the delayed operation to perform
+    # @!macro [attach] executor_and_deref_options
+    #  
+    #   @param [Hash] opts the options used to define the behavior at update and deref
+    #     and to specify the executor on which to perform actions
+    #   @option opts [Executor] :executor when set use the given `Executor` instance.
+    #     Three special values are also supported: `:task` returns the global task pool,
+    #     `:operation` returns the global operation pool, and `:immediate` returns a new
+    #     `ImmediateExecutor` object.
+    #   @option opts [Boolean] :dup_on_deref (false) call `#dup` before returning the data
+    #   @option opts [Boolean] :freeze_on_deref (false) call `#freeze` before returning the data
+    #   @option opts [Proc] :copy_on_deref (nil) call the given `Proc` passing
+    #     the internal value and returning the value returned from the proc
     #
-    # @!macro executor_and_deref_options
+    # @yield the delayed operation to perform
     #
     # @raise [ArgumentError] if no block is given
     def initialize(opts = {}, &block)
--- a/lib/concurrent/delay.rb
+++ b/lib/concurrent/delay.rb
@@ -3,6 +3,7 @@ require 'concurrent/configuration'
 require 'concurrent/obligation'
 require 'concurrent/executor/executor_options'
 require 'concurrent/executor/immediate_executor'
+require 'concurrent/synchronization'
 
 module Concurrent
 
@@ -37,7 +38,7 @@ module Concurrent
   #     execute on the given executor, allowing the call to timeout.
   #
   # @see Concurrent::Dereferenceable
-  class Delay
+  class Delay < Synchronization::Object
     include Obligation
     include ExecutorOptions
 
@@ -59,7 +60,8 @@ module Concurrent
     def initialize(opts = {}, &block)
       raise ArgumentError.new('no block given') unless block_given?
 
-      init_obligation
+      super(&nil)
+      init_obligation(self)
       set_deref_options(opts)
       @task_executor = get_executor_from(opts)
 
@@ -145,16 +147,15 @@ module Concurrent
     # @yield the delayed operation to perform
     # @return [true, false] if success
     def reconfigure(&block)
-      mutex.lock
-      raise ArgumentError.new('no block given') unless block_given?
-      unless @computing
-        @task = block
-        true
-      else
-        false
+      mutex.synchronize do
+        raise ArgumentError.new('no block given') unless block_given?
+        unless @computing
+          @task = block
+          true
+        else
+          false
+        end
       end
-    ensure
-      mutex.unlock
     end
 
     private
@@ -163,10 +164,11 @@ module Concurrent
     def execute_task_once # :nodoc:
       # this function has been optimized for performance and
       # should not be modified without running new benchmarks
-      mutex.lock
-      execute = @computing = true unless @computing
-      task    = @task
-      mutex.unlock
+      execute = task = nil
+      mutex.synchronize do
+        execute = @computing = true unless @computing
+        task    = @task
+      end
 
       if execute
         @task_executor.post do
@@ -176,10 +178,10 @@ module Concurrent
           rescue => ex
             reason = ex
           end
-          mutex.lock
-          set_state(success, result, reason)
-          event.set
-          mutex.unlock
+          mutex.synchronize do
+            set_state(success, result, reason)
+            event.set
+          end
         end
       end
     end
--- a/lib/concurrent/delay.rb
+++ b/lib/concurrent/delay.rb
@@ -60,7 +60,7 @@ module Concurrent
     def initialize(opts = {}, &block)
       raise ArgumentError.new('no block given') unless block_given?
 
-      super(&nil)
+      super()
       init_obligation(self)
       set_deref_options(opts)
       @task_executor = get_executor_from(opts)
--- a/lib/concurrent/ivar.rb
+++ b/lib/concurrent/ivar.rb
@@ -100,9 +100,13 @@ module Concurrent
 
     # Set the `IVar` to a value and wake or notify all threads waiting on it.
     #
-    # @param [Object] value the value to store in the `IVar`
-    # @raise [Concurrent::MultipleAssignmentError] if the `IVar` has already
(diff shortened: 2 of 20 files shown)


pitr-ch:
Thanks @jdantonio, looks much better. Testing.

lib/concurrent/promise.rb
```diff
@@ -388,6 +385,18 @@ def self.any?(*promises)
       aggregate(:any?, *promises)
     end
 
+    def set(value = IVar::NO_VALUE)
+      raise PromiseExecutionError.new('supported only on root promise') unless root?
+      super
+    end
```
pitr-ch: It might make sense to also call `execute` here.

obrok:
Sorry if this doesn't make sense, because I might not have the full context:

I run into a problem with manually setting promises where if you attach child promises `execute` will sometimes get called on your promise before you manage to set its value, triggered by one of the children. My workaround for that was the following:

```ruby
class FulfillablePromise < Concurrent::Promise
  def initialize(*)
    super
    @state = :pending
  end

  def fulfill(value)
    on_fulfill(value)
  end

  def reject(error)
    on_reject(error)
  end
end

```

jdantonio:
@obrok That is very good information. Thank you. I think the context of this PR may be a little different than your example. If I correctly understand your example, it is possible to manually set the value of a child Promise. In this use case we only want to manually set the root promise and never manually set the children. Basically:

* Create the root Promise but do not execute
* Add children to form a tree/chain
* Manually set or fail the root Promise
* Allow the chain/tree to execute normally based on the new state of the root Promise

There may be bugs in this implementation because it is just a spike, but the *intent* is that we would *never* manually set any of the children. This is why `set` and `fail` raise exceptions when not the root Promise.

Does that make sense, or do I misunderstand your example?

obrok:
It makes sense but I still think there might be a problem. Consider this function (which I imagine is the use case):

```ruby
def do_something_later
  promise = Promise.new
  Thread.new { promise.set(:result) }
  promise
end

do_something_later.then { |result| p result }.execute # I think this will output nil most of the time or possibly it's a race
```

The client of `do_something_later` is well within his rights to add children to it and possibly call execute on the children. If the root promise is in the `:unscheduled` state it will cause it to execute the default empty block and set it's value to `nil`.

lib/concurrent/promise.rb
```diff
@@ -243,6 +240,26 @@ def execute
       self
     end
 
+    def set(value = IVar::NO_VALUE, &block)
```
jdantonio: @pitr-ch This version of `set` follows your suggestion and calls `execute`. The function itself could use a little cleanup but the tests still pass. I think the behavior here is probably more accurate. I haven't changed `fail` yet but I anticipate it will be very similar except for something like this:

```ruby
@promise_body = Proc.new { |result| raise reason }
```

Thoughts?

jdantonio:
@obrok That makes sense. Once we get the behavior right for manually setting the root promise we can dig more deeply into your example.

pitr-ch:
Based on the tests it seems there is a race hidden. What is the purpose of `NO_VALUE`? I am still testing with Dynflow, looks good but I am hitting another issue with pools - investigating. We could also consider running Dynflow test suite within `concurrent-ruby` travis tests (and other projects we can find) to test some real usage scenario.

jdantonio:
`NO_VALUE` was part of @chrisseaton's original implementation of `IVar` and we've always carried it forward.

chrisseaton:
The point of `NO_VALUE` was to represent that the `IVar` currently had no value. I didn't want to use `nil` as I wanted `nil` to be a valid value for an assigned `IVar`. However a better option would be to use an array to hold the value, so an empty array for no value, or to have a flag to say if there is a value or not.

jdantonio:
This update makes `Delay` a little slower. We may want to rollback the most recent commit and reset `Delay`. Performance from that class is probably more important than making it a `Synchronization::Object`.

```
BEFORE

$ ./examples/benchmark_lazy_and_delay.rb
Rehearsal -------------------------------------------------------
Delay#value           0.360000   0.000000   0.360000 (  0.361039)
Delay#value!          0.430000   0.000000   0.430000 (  0.438568)
LazyReference#value   0.290000   0.000000   0.290000 (  0.285422)
---------------------------------------------- total: 1.080000sec

Delay#value           0.370000   0.000000   0.370000 (  0.376526)
Delay#value!          0.470000   0.000000   0.470000 (  0.470254)
LazyReference#value   0.270000   0.000000   0.270000 (  0.268231)

AFTER

$ ./examples/benchmark_lazy_and_delay.rb
Rehearsal -------------------------------------------------------
Delay#value           0.440000   0.000000   0.440000 (  0.448287)
Delay#value!          0.560000   0.000000   0.560000 (  0.563817)
LazyReference#value   0.280000   0.000000   0.280000 (  0.282739)
---------------------------------------------- total: 1.280000sec

Delay#value           0.450000   0.000000   0.450000 (  0.461807)
Delay#value!          0.540000   0.000000   0.540000 (  0.552702)
LazyReference#value   0.270000   0.000000   0.270000 (  0.273209)
```

pitr-ch:
@jdantonio I would suggest merging with `Synchronization::Object`. The correctness (`ivar` read is not volatile on Rubinius) is more important than slight performance gain. I also think we will get it back when volatile reads and final-field-constructor-visibility is abstracted in `Synchronization` to be able to remove synchronization around lot of the `ivar` reads which I think is source of the performance decrease.

jdantonio:
@pitr-ch Sounds good. I'll keep it as-is. I'll rebase this on master once you merge #274.

jdantonio:
The test that failed in the last run is one that is known to fail intermittently. See #271

jdantonio:
@pitr-ch Is it OK with you if I merge this? Or would you rather I wait until #274 is ready?

lib/concurrent/delay.rb
```diff
@@ -145,16 +147,15 @@ def wait(timeout = nil)
...
-      else
-        false
+      mutex.synchronize do
+        raise ArgumentError.new('no block given') unless block_given?
+        unless @computing
+          @task = block
+          true
+        else
+          false
+        end
```
pitr-ch: Since delay is migrated to `Delay < Synchronization::Object` it probably should be using its method `synchronize`, is there a reason for using `mutex` which I am missing?

lib/concurrent/ivar.rb
```diff
@@ -57,15 +57,13 @@ class IVar
     # @option opts [String] :copy_on_deref (nil) call the given `Proc` passing
     #   the internal value and returning the value returned from the proc
     def initialize(value = NO_VALUE, opts = {})
-      init_obligation
+      super(&nil)
```
pitr-ch: no need for `&nil` here, it is required only when initialize is called with block which should not be passed to super

lib/concurrent/ivar.rb
```diff
@@ -114,14 +121,24 @@ def set(value)
     # @param [Object] reason for the failure
     # @raise [Concurrent::MultipleAssignmentError] if the `IVar` has already
     #   been set or otherwise completed
+    # @return [IVar] self
     def fail(reason = StandardError.new)
-      complete(false, nil, reason)
+      set { raise reason }
```
pitr-ch: the original `complete(false, nil, reason)` was faster then re-raising in block `set { raise reason }`

pitr-ch:
@jdantonio I've looked through the code and add few comments. I think this is not dependent on #274. This brings some of the features of the new futures to the original implementation. (I really have to sit down soon and write a documentation for that.)

lib/concurrent/delay.rb
jdantonio: It is using the `#synchronize` method from `Synchronization::Object`, it's just doing so through a level of indirection. See [line 64](https://github.com/ruby-concurrency/concurrent-ruby/blob/ivar-promise/lib/concurrent/delay.rb#L64). `Delay` includes `Obligation` which also includes `Dereferenceable`, the latter of which creates its own mutex within the `#init_mutex` function. I've added an argument to both `#init_obligation` and `#init_mutex` that allows the mutex to be injected. `Delay` passes `self` into those initialization functions, thus allowing the delay object to be its own lock. The `#mutex` method returns `self`. The `mutex.synchronize` call on line 150 is actually `self.synchronize`.

lib/concurrent/ivar.rb
jdantonio: Calling `#complete` here is incompatible with `Promise`. A promise can be set only once so calling `#fail` more than once needs to raise an exception. That doesn't happen when we call `#complete` here. It's an easy fix. I'll override `#fail` in `Promise` and return `IVar` to the original implementation.

lib/concurrent/delay.rb
pitr-ch: Ah right, I've missed that. It seems unnecessarily complicated, but I am not sure if it makes sense to invest time in refactoring it - I still hope that the new futures will replace this class soon.

PR #272, opened 2015-04-02
Refactor and simplify pool implementation

pitr-ch:
Motivation: I've hit weird errors trying to run Dynflow on the original pool with `max_length` set to max (`32768`) it was not caching threads as expected and the process was running out of resources, failing at `2048` threads. I was able to reproduce with simple stress test (included in this PR) where the thread count was rising uncontrollably again. I've tried to fix the original implementation but it turned out to be difficult. In the end I decided to reimplement the core keeping same external API. It should be easier to understand and I've also tried to comment the code. This implementation is kind a naive not including any optimization tricks but it should be correct and we can tweak performance later using this as a base for comparison. Before merging this needs to be compared to the old implementation ensuring that it's not significantly slower.

`ns_` stands for 'no synchronization' marking methods which must be called inside synchronized block. It's there to separate synchronization from behavior and to avoid forgetting synchronization.

TODOs
- [x] Benchmarks
- [x] FixedThreadPool is not raising at the beginning to the full capacity but slowly as needed.
- [x] What to do with deprecated `#status` method? How old is it? Could it be removed?


pitr-ch:
cc @iNecas

jdantonio:
:+1: It's Easter weekend so I probably won't be able to look at this until next week, but I'm all for internally refactoring these classes. Everything else depends on us having rock-solid thread pools.

pitr-ch:
Benchmark results
```txt
MRI 2.2.0
$ bundle exec ruby examples/benchmark_pool.rb
                 old    63.000  i/100ms
                 new    82.000  i/100ms
                 old    650.824  (±14.8%) i/s -      6.363k
                 new    882.164  (± 8.4%) i/s -      8.774k

Comparison:
                 new:      882.2 i/s
                 old:      650.8 i/s - 1.36x slower

JRuby 1.7.19
bundle exec ruby --server examples/benchmark_pool.rb
                 old    34.000  i/100ms
                 new    28.000  i/100ms
                java    23.000  i/100ms
                 old    368.152  (± 8.7%) i/s -      3.672k
                 new    371.866  (± 9.9%) i/s -      3.696k
                java    315.979  (±19.3%) i/s -      2.875k

Comparison:
                 new:      371.9 i/s
                 old:      368.2 i/s - 1.01x slower
                java:      316.0 i/s - 1.18x slower
```

Weird think is that Java native pool is slower than pure ruby pools. We should probably add the Java wrapper and try again.

jdantonio:
@pitr-ch Thanks for running these performance comparisons. I agree that we should build a pure-Java wrapper and run another comparison. I'm starting to get the impression that there is significant overhead in crossing the boundary between Ruby and Java code. I assumed the interop would be faster but there is probably a lot happening there.

iNecas:
I'm glad there doesn't seem to be a dramatic impact on the performance with the polished implementation of the pools… Thanks @pitr-ch for the work here. I agree that the working pool is the crucial part in concurrent-ruby.

lib/concurrent/executor/executor.rb
```diff
@@ -5,12 +5,75 @@
...
+        end
+        self
+      end
+
+      def executors
+        @mutex.synchronize { @executors.to_a }
+      end
+    end
+
+    AT_EXIT_AUTO_TERMINATION = AtExitAutoTermination.new.install
```
jdantonio: @pitr-ch Perhaps I'm missing something here, but this looks an awful lot like a global registry of thread pools. Something I specifically said I wanted to avoid. Is that correct?

lib/concurrent/configuration.rb
```diff
@@ -160,44 +96,25 @@ def self.global_timer_set
...
-    [ global_fast_executor, global_io_executor, global_timer_set ].each do |executor|
-      Thread.new{ executor.wait_for_termination(timeout); latch.count_down }
-    end
-    latch.wait(timeout)
-  end
-
   def self.new_fast_executor(opts = {})
     FixedThreadPool.new(
-      [2, Concurrent.processor_count].max,
-      stop_on_exit:    opts.fetch(:stop_on_exit, true),
```
jdantonio: I think `auto_terminate` is a better name than `stop_on_exit`. :+1:

lib/concurrent/executor/executor_options.rb
```diff
@@ -19,31 +19,41 @@ module ExecutorOptions # :nodoc:
...
         when :immediate
-          Concurrent::ImmediateExecutor.new
+          Concurrent.global_immediate_executor
         when :operation
           Kernel.warn '[DEPRECATED] use `executor: :fast` instead'
           Concurrent.global_fast_executor
         when :task
           Kernel.warn '[DEPRECATED] use `executor: :io` instead'
           Concurrent.global_io_executor
+        when Executor
```
jdantonio: This is a nifty trick. I didn't realize you could type check in a case statement this way.

lib/concurrent/executor/java_thread_pool_executor.rb
```diff
@@ -149,15 +149,6 @@ def remaining_capacity
         @max_queue == 0 ? -1 : @executor.getQueue.remainingCapacity
       end
 
-      # This method is deprecated and will be removed soon.
-      # This method is supost to return the threads status, but Java API doesn't
-      # provide a way to get the thread status. So we return an empty Array instead.
-      def status
```
jdantonio: I had completely forgotten about this. Nice catch. It's definitely time to remove it.

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -123,142 +122,231 @@ def queue_length
...
+      end
     end
 
-    # Returns an array with the status of each thread in the pool
-    #
-    # This method is deprecated and will be removed soon.
-    def status
-      warn '[DEPRECATED] `status` is deprecated and will be removed soon.'
-      mutex.synchronize { @pool.collect { |worker| worker.status } }
+    alias_method :current_length, :length # TODO remove?
```
jdantonio: I don't remember why I created this alias in the first place, but I've come to the realization that I used to make far too many aliases. I'm :ok: to remove it.

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -123,142 +122,231 @@ def queue_length
...
+    alias_method :shutdown_execution, :ns_shutdown_execution
+
+    # @api private
+    def ns_kill_execution
+      ns_shutdown_execution
+      unless stopped_event.wait(1)
+        @pool.each &:kill
+        @pool.clear
+        @ready.clear
+        # TODO log out unprocessed tasks in queue
```
jdantonio: If we were to log from here we would have to do the same from the Java implementation. I don't feel there is a strong need to add logging, but I'm not opposed to it either. It's probably something that can wait until a future release if we decide to do it.

jdantonio:
:+1:

lib/concurrent/executor/executor.rb
pitr-ch: Ah sorry about that @jdantonio I forgot about that. Problem is that we cannot use `at_exit` hooks directly because they cannot be removed which results in leeking references. So perhaps we could create `Concurrent.at_exit` interface which will return id which can be used to deregister the hook. It would still has to have a global registry of the hooks but it would be useful API and it would be decoupled from Pools so it's kind a middle ground.

jdantonio: The current implementation is fine and I'm comfortable with us merging to master.

I don't like creeping complexity. The original `at_exit` handlers were a part of the Java thread pools since the initial implementation and no one cared. No one even noticed until we discovered a bug with autotest. In the end that bug had nothing to do with the `at_exit` handlers--it was a bug in Actor caused by an incorrect use of Delay that was, in turn, caused by a change in the default behavior of Delay. We've solved those problems. Had that bug not existed we may never have changed the `at_exit` handlers at all. Instead, we added an increasing level of complexity, added more global objects, and tightly coupled thread pools to a global registry--all so we can have the same behavior we had *before* we changed anything. I appreciate that the old implementation created a small memory leak, but I can't imagine any circumstance under which that wouldn't be trivial. No application should create more than a few thread pools. I'm not convinced that the added complexity is really warranted. That being said, I realize that my opinion is only one of many. If others are more comfortable with a global registry for shutting down thread pools then I'm fine with it, too (at least one or two others favored this approach). My overarching concern has always been the unfortunate behavior on JRuby where applications will fail to exit if the thread pools aren't properly shut down. Since we create global thread pools in the background I feel strongly that we have a responsibility to manage those thread pools for our users. This solution solves that problem so I'm good.

lib/concurrent/executor/ruby_thread_pool_executor.rb
pitr-ch: Done.

lib/concurrent/executor/ruby_thread_pool_executor.rb
pitr-ch: Only a debug level logging would make sense probably, when user has something not executing it may be because of premature termination so I would like to inform him.

pitr-ch:
TODOs:
- [x] rebase
- [ ] migrate to `Synchronization::Object`
- [ ] global registry
- merge

lib/concurrent/executor/executor.rb
pitr-ch: I also do not like the added complexity. However there is a valid case where the pools would be created periodically (maybe little obscure but I think still correct usage): long running background processor of CPU-heavy tasks of different nature, the processor can decide to create differently configured pools based on the task it is given to improve performance. The leak could cause problems if it runs long enough. For that reason I think we can justify the complexity exchanging it for fixed leak. Please look at the last commit, I've tried to reduce the complexity there https://github.com/ruby-concurrency/concurrent-ruby/commit/9db56e4a519d2f4a9b74af09c65076f305d533c6 by extracting the register to be a general `AtExit` manager.

lib/concurrent/at_exit.rb
```diff
@@ -0,0 +1,92 @@
+require 'concurrent/logging'
+require 'concurrent/synchronization'
+
+module Concurrent
+
+  # Provides ability to add and remove hooks to be run at `Kernel#at_exit`, order is undefined.
+  # Each hook is executed at most once.
+  class AtExitImplementation < Synchronization::Object
```
jdantonio: @pitr-ch I wasn't expecting you to revisit the thread pool registry, but I'm glad you did. I *really* like this. Given the nature of this library, providing synchronization protection for our at_exit handlers makes a lot of sense. I like that this is now a general utility that we can use anywhere in the library should the need arise. This is much more versatile. Thank you!

One small request: Can we please change "hook" to "handler"? I know it's a little pedantic but I think that's a better term.

pitr-ch: It's not pedantic at all. I gladly add any improvement. It'll be here in a minute.

Issue #241, opened 2015-02-13
Synchronization layer


pitr-ch:
Hello,

I would like to build a common synchronization layer in `concurrent-ruby` having following API:
```ruby
def Concurrent.lock(object); end #reentrant
def Concurrent.try_lock(object); end
def Concurrent.unlock(object); end

def Concurrent.synchronize(object, &block); end

def Concurrent.wait(object, timeout = nil); end # as Java Object#wait
def Concurrent.signal(object); end # as Java Object#notify
def Concurrent.broadcast(object); end # as Java Object#notifyAll
```
Which would have different implementation based on Ruby interpreter.

I kindly ask @brixen @headius @hone @ko1 @YorickPeterse @zzak to help with finding the best options for the different implementations. This PR is intended to serve to discus the options and issues, I'll open PR with implementation later to discuss details based on what is figured out here.

### MRI

Injecting Monitor to an object, something like:
```ruby
    require 'monitor'
    INJECTION_BARRIER = Mutex.new

    def synchronize(object)
      get_mutex(object).synchronize { yield }
    end

    def lock(object)
      get_mutex(object).enter
    end

    def unlock(object)
      get_mutex(object).exit
    end

    def get_mutex(object)
      object.instance_variable_get(:@__monitor__) or INJECTION_BARRIER.synchronize do
        object.instance_variable_get(:@__monitor__) or
            object.instance_variable_set(:@__monitor__, Monitor.new)
      end
    end
```

Similarly a `ConditionVariable` would be injected to support `wait`, `signal`, `broadcast`.

### JRuby

Delegate to methods on `JRuby` module, see this PR https://github.com/jruby/jruby/pull/2594. 

It uses `wait`, `notify`, `notifyAll` methods on Java object to implement `wait`, `signal`, `broadcast`. `synchronize` runs a `Block` wrapped in  `synchronized` Java keyword. `lock`, `unlock`, `try_lock` are using methods `monitorEnter`, `monitorExit`, `tryMonitorEnter` on `sun.misc.Unsafe`.

### Rubinius

Implement `lock`, `unlock`, `try_lock` and `synchronize` by delegation to same methods on `Rubinius` module. `wait`, `signal`, `broadcast` would be supported by injecting `ConditionalVariable` into object's instance variable (when locked).

pitr-ch:
[...]

PR #273, opened 2015-04-07
Synchronization layer

pitr-ch:
link #241

If you agree with the approach I'll merge this PR because of fixes in https://github.com/ruby-concurrency/concurrent-ruby/commit/d66f6161d78c6509712d75209d48b0854a9b79fb and I'll start migrating all the other classes to this layer in another PR. Event is migrated as an example.


jdantonio:
I'm sorry for not having an opportunity to provide detailed feedback yet. I'm participating in [SpaceApps](https://spaceappschallenge.org/announcement/) this weekend and I've been busy preparing. I did look over the PR and I really like the direction you have taken. :+1: Consolidating the helpers into 'engine' is a nice redactor. The fact that it touched so many files is an indication of how necessary that was.

pitr-ch:
Ready for merge.

lib/concurrent/atomic.rb
```diff
@@ -3,16 +3,17 @@
 # user that they should use the new implementation instead.
 
 if defined?(Atomic)
-  warn <<-RUBY
+  warn <<-TXT
```
jdantonio: We should consider removing this entirely for 0.9. since this message was introduced in 0.8. We definitely do not want it in the 1.0 release.

lib/concurrent/atomic.rb
```diff
@@ -21,7 +22,7 @@
   if /[^0fF]/ =~ ENV['FORCE_ATOMIC_FALLBACK']
     ruby_engine = 'mutex_atomic'
   else
-    ruby_engine = defined?(RUBY_ENGINE)? RUBY_ENGINE : 'ruby'
+    ruby_engine = Concurrent.ruby_engine
```
jdantonio: :+1: 

I really like this `ruby_gem` abstraction. We're starting to check this so often (and will continue to do so as we add more optimizations) that we needed a cleaner way to check.

lib/concurrent/synchronized_object.rb
```diff
@@ -0,0 +1,243 @@
...
+  #     def initialize
+  #       super
+  #       synchronize { @value = 'asd' }
+  #     end
+  #
+  #     def value
+  #       synchronize { @value }
+  #     end
+  #   end
+  class AbstractSynchronizedObject
```
jdantonio: I really like this abstraction a lot. :+1: I love that this is the result of the gem's evolution--a pattern emerged over time and we refactored into this. Thank you very much for creating this!

jdantonio:
:+1: I'll rebase #270 and #271 from master once this is merged and get back to work on those updates ASAP.

Thank you!

pitr-ch:
Found one more problem which needs fixing.

pitr-ch:
The issue I've mentioned is fixed in https://github.com/ruby-concurrency/concurrent-ruby/commit/e78b4a5bbbd20354239fbcea039f34d879981c72 but I am experiencing failing JRuby tests (e.g. cyclic_barrier), it happens only when whole suite runs, if executed in isolation it's fine :/ I need to find the reason for that before we can merge.

jdantonio:
@pitr-ch I'm free this evening. I'll work on it tonight, too. Between the two of us we should be able to figure it out. We're pretty smart dudes. :-)

jdantonio:
@pitr-ch OK, so apparently I'm not as smart as I think I am. I am able to consistently reproduce the failed tests on JRuby but I have been unable to isolate the upstream tests that are causing the problem. Have you had any luck?

pitr-ch:
Thanks @jdantonio, did you find out anything? I've also continued digging, so far I have:

- if JRuby compilation is disabled (`-J-Djruby.compile.mode=OFF`) it works as expected
- if debugging prints are added
  - `ns_wait_until` reports its return value correctly
  - any print in `CyclicBarrier#wait` is skipped, it looks like the body of this method is skipped too, it returns in the spec where the barrier's state looks like the body of `wait` was skipped
- reproducer: only `cyclic_barrier_spec` where spec `breaks the barrier and release all other threads` is run 100x until it compiles and starts to fail, on jruby-1.7.19

I am going to sleep on it before deciding what to do next :)

pitr-ch:
It looks like if `while true` is replaced with `loop do` at https://github.com/ruby-concurrency/concurrent-ruby/blob/synchronization/lib/concurrent/synchronized_object.rb#L78 it starts to work :fireworks:

jdantonio:
Interesting. I'll pull the latest branch and see if it works for me, too. Thanks for keeping at it!

jdantonio:
@pitr-ch As far as I can tell your latest update fixed it. On my local machine I rebased this branch onto master (where I merged PR #275) and the tests pass on both JRuby and MRI. I haven't pushed that update yet because I wanted to coordinate with you, first.

I'm comfortable rebasing this PR from master then merging it. That will allow us to work on the other PRs. I also plan to revisit the at_exit handlers for the global thread pools and fix the discrepancy you pointed out a couple of weeks ago.

Are you comfortable with us committing this PR?

pitr-ch:
Thanks. I'll need to fix few TODOs in the code and verify that all the failing tests are known intermittently failing tests. After that, I'll merge it.

@jdantonio please review #272 first, I've already touched the at_exit handlers there.

jdantonio:
@pitr-ch Can we delete this branch?

pitr-ch:
@jdantonio done

jdantonio:
Thank you!

PR #275, opened 2015-04-10
Fix deadlock of Future when the handler raises Exception

amutake:
I found deadlock situation in using `Future` when the handler raises `Exception` (not `StandardError`).

```ruby
require 'concurrent'

future = Concurrent::Future.new do
  raise Exception
end
future.execute
future.value #=> deadlock
```

```
/Users/shohei-yasutake/work/concurrent-ruby/lib/concurrent/atomic/condition.rb:50:in `sleep': No live threads left. Deadlock? (fatal)
	from /Users/shohei-yasutake/work/concurrent-ruby/lib/concurrent/atomic/condition.rb:50:in `wait'
	from /Users/shohei-yasutake/work/concurrent-ruby/lib/concurrent/atomic/condition.rb:50:in `wait'
	from /Users/shohei-yasutake/work/concurrent-ruby/lib/concurrent/atomic/event.rb:89:in `wait'
	from /Users/shohei-yasutake/work/concurrent-ruby/lib/concurrent/obligation.rb:80:in `wait'
	from /Users/shohei-yasutake/work/concurrent-ruby/lib/concurrent/obligation.rb:71:in `value'
	from example.rb:7:in `<main>'
```

It is because the exception raised in `SafeTaskExecutor.new(@task).execute(*@args)` is not caught and `complete(success, val, reason)` is not called.

https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/future.rb#L85-L86

So I fixed to catch `Exception` in `SafeTaskExecutor`.

### DIFF
--- a/spec/concurrent/future_spec.rb
+++ b/spec/concurrent/future_spec.rb
@@ -209,6 +209,11 @@ module Concurrent
         expect(future.value).to be_nil
       end
 
+      it 'sets the reason to the Exception instance when the handler raises Exception' do
+        future = Future.new(executor: executor){ raise Exception }.execute
+        expect(future.reason).to be_a(Exception)
+      end
+
       it 'sets the state to :rejected when the handler raises an exception' do
         future = Future.new(executor: executor){ raise StandardError }.execute
         expect(future).to be_rejected
--- a/spec/concurrent/future_spec.rb
+++ b/spec/concurrent/future_spec.rb
@@ -204,6 +204,11 @@ module Concurrent
         expect(future.value).to be_nil
       end
 
+      it 'sets the value to nil when the handler raises Exception' do
+        future = Future.new(executor: executor){ raise Exception }.execute
+        expect(future.value).to be_nil
+      end
+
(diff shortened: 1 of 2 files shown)


pitr-ch:
Thanks for reporting and for the fix!

spec/concurrent/future_spec.rb
```diff
@@ -204,6 +204,11 @@ def trigger_observable(observable)
         expect(future.value).to be_nil
       end
 
+      it 'sets the value to nil when the handler raises Exception' do
+        future = Future.new(executor: executor){ raise Exception }.execute
+        expect(future.value).to be_nil
```
pitr-ch: Could you also test that `future.reason` is set to the `Exception` instance?

amutake: Thank you for quick response!

I added a test case. Could you check it?

jdantonio:
@amutake Thank you very much for catching this and taking the time to fix it! I'll take a look at this in more detail this evening. Hopefully we can get it merged today.

PR #276, opened 2015-04-21
Synchronization: Remove mistakenly added lines

pitr-ch:


jdantonio:
I don't understand. Was this just debugging code? If you say this code isn't necessary I take your word for it, I'm just curious.

pitr-ch:
These methods are just experiments of next version of synchronized module (included by mistake). It's not even working yet. It will abstract volatile access to ivars and probable also CAS operations on them if required, there will be different implementations depending on Ruby interpreter. Motivation is performance, to be able to replace synchronizations with volatile reads which will be noop (meaning just the `@ivar` read) (AFAIK) on CRuby and JRuby.

jdantonio:
Makes sense. Thank you!

Issue #261, opened 2015-03-04
Namespaces cleanups


pitr-ch:
go through whole gem and make sure name-spacing is fine everywhere. It will be hard to correct after 1.0 release. One thing comes in mind the name-spacing around atomic reference (result of the gem merge) is little inconsistent, there may be more. Check `Concurrent.methods` for pollution. Hide constants which are not part of public API.

pitr-ch:
Use similar documentation friendly pattern for the different implementations:
```ruby
module Synchronization
    Implementation = case
                     when Concurrent.on_jruby?
                       JavaObject
                     when Concurrent.on_cruby? && (RUBY_VERSION.split('.').map(&:to_i) <=> [1, 9, 3]) <= 0
                       MonitorObject
                     when Concurrent.on_cruby?
                       MutexObject
                     when Concurrent.on_rbx?
                       RbxObject
                     else
                       MutexObject
                     end
    private_constant :Implementation

    class Object < Implementation
    end
  end
```

jdantonio:
The naming of `Atomic` was a [conversation in gitter](https://gitter.im/ruby-concurrency/concurrent-ruby?at=5535d43ffd52de687867e60d) earlier this week. My suggestion from that conversation:

```
We've considered changing the name of Concurrent::Atomic to Concurrent::AtomicReference before the 1.0 release and that's still an open question. We could add an Concurrent::AtomicReference alias in 0.9 and put a deprecation warning to Concurrent::Atomic then remove Concurrent::Atomic in 1.0.
```

pitr-ch:
:+1: for `AtomicReference`

jdantonio:
`AtomicReference` change was merged to master in PR #277.

ianks:
Nice. This was important because it will allow us to differentiate between different types of AtomicReferences (i.e. AtomicMarkableReference, etc.).

pitr-ch:
Current `Concurrent` namespace 
```txt
ls Concurrent
constants:
  AbstractExecutorService            Configuration                  GLOBAL_IO_EXECUTOR         NULL_LOGGER               SemaphoreImplementation
  AbstractStruct                     CONFIGURATION                  GLOBAL_LOGGER              Obligation                SerialExecutorService
  AbstractThreadLocalVar             ConfigurationError             GLOBAL_MONOTONIC_CLOCK     Observable                SerializedExecution
[...]

PR #277, opened 2015-04-23
Deprecated Concurrent::Atomic in lieu of Concurrent::AtomicReference.

jdantonio:
See #261


lib/concurrent/atomic_reference.rb
```diff
@@ -0,0 +1,104 @@
...
+
+# @!macro atomic_reference
+# @deprecated Use Concurrent::MutexAtomicReference instead.
+class Concurrent::MutexAtomic < Concurrent::MutexAtomicReference
+end
+
+# @!macro atomic_reference
+# @deprecated Use Concurrent::AtomicReference instead.
+class Concurrent::Atomic < Concurrent::AtomicReference
+end
```
pitr-ch: Do we need aliases for the different implementations? In my mind they are not included in the public API. Just assignment `Atomic = AtomicReference` might be better since it clearly says what is the main name and the alias will also correctly report full name `AtomicReference`.

lib/concurrent/atomics.rb
```diff
@@ -1,4 +1,4 @@
-require 'concurrent/atomic'
+require 'concurrent/atomic_reference'
 require 'concurrent/atomic/atomic_boolean'
 require 'concurrent/atomic/atomic_fixnum'
```
pitr-ch: Could you move 'atomic_reference' -> 'atomic/atomic_reference'?

jdantonio: Done.

lib/concurrent/atomic_reference.rb
jdantonio: Done.

jdantonio:
The test that failed in the last run is one that is known to fail intermittently. See #271

lib/concurrent/atomic/atomic_reference.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  # @!macro atomic_reference
+  class Concurrent::AtomicReference < Concurrent::MutexAtomicReference
+  end
+end
+
+# @!macro atomic_reference
+# @deprecated Use Concurrent::AtomicReference instead.
+class Concurrent::Atomic < Concurrent::AtomicReference
+end
```
pitr-ch: I meant something like:
```ruby
module Concurrent
  Atomic = AtomicReference
end
```
So it is actually the same class, because having two classes with same behavior but not-identical may be causing problems in comparisons when both names are used.

pitr-ch:
Are you able to restart the jobs on travis? It does not work for me :(

lib/concurrent/atomic/atomic_reference.rb
jdantonio: When I've tried to do that in the past yardoc wouldn't generate documentation for the 'alias' class. That's the main reason I did the subclass. What you say makes sense, though. I'll make the change and then figure out some way to get reasonable documentation in yardoc.

jdantonio:
I was able to restart it, but I had to log in to Travis in order to make the "restart" button visible.

pitr-ch:
Thanks, I saw it but it was doing nothing :/

Issue #280, opened 2015-04-25
Performance improvements for Executor#post


headius:
The `#post` method in Executor has a very large amount of overhead:

* Several closures, including one created *solely* to pass args through to the given block.
* Use of java_method on *every* call.
* Creation of `[Runnable.java_class]` array on every call.

The closures are especially heavy because they force us to run this method with a heap-based local variable scope, even though it's *only* used for passing args to task.

I have a few small patches that improve this situation. They're not suitable for a PR at the moment (quick one-off work) but they're there.

https://gist.github.com/headius/3807e0180e83ed7c98a2

This patch does the following:

* Uses a Runnable-implementing Job class to hold args + block, rather than creating another closure.
* Decorates the executors as they're created to have a submit_runnable method that always uses submit(Runnable).

I tested this with both SingleThreadExecutor and ThreadPoolExecutor using the following code (just timing total run time and no warmup):

```ruby
exe = Concurrent::ThreadPoolExecutor.new
t = Time.now
i = 0
while i < 1_000_000
  exe.post { }
  i+=1
end
puts Time.now - t
```

With my changes the run time goes from 29s to around 5s.

I can help you massage these into place, but the patch as written is just a one-off.

jdantonio:
@headius Thank you very much for noticing this. I'm currently working on some updates to our executors. I'll take a look at your patch and if I have any questions I'll ask ASAP.

headius:
It would worth doing a pass over the rest of the codebase looking for
closures in hot methods. They're a perf hit on all impls compared to
methods, and you can often proc'ify a method and get the same iterator
pattern for free.

- Charlie (mobile)

jdantonio:
@headius Maybe I'm doing something wrong, but I get this error:

```
cannot convert instance of class org.jruby.RubyModule to class java.lang.Class
```

at

```
java_alias :submit_runnable, :submit, [java.lang.Runnable]
```

Thoughts?

headius:
Ohh sorry...I found that was an issue in 1.7 after I submitted. Add
.java_class to the end of Runnable.

headius:
I fixed the gist, but I do want to point out that that particular part I only did to two of the executors, not to all of them. If there's a better way to decorate them all, that would be great...but until then this will need to be done for each place where Executors factory is used to create a new Executor.

jdantonio:
Fixed in #283

Issue #278, opened 2015-04-23
Migrate all remaining classes to Synchronization::Object


pitr-ch:
Executors should be done asap.

jdantonio:
I've spiked two different implementation of this, neither of which I am entirely happy with. Both feel like they would be completely refactored before a v2.0 release, but both could possibly work for the v1.0 release. I have performed no performance benchmarks.

* 62fe60cfc15e7da351c0a68805090e4576eee455 has the executors extend `Synchronization::Object` directly then inject themselves into the `Executor` module as the lock object.
* 490680cd3565ecbece54eed19ee989b0a17f2986 included a new `ExecutorService` class which extends `Synchronization::Object` and has Ruby/JRuby-optimized versions. The executors extend `ExecutorService`.

Thoughts? Have you been working on or thinking about a different implementation?

pitr-ch:
I was thinking about following class structure:
`AbstractExecutor` < `Synchronization::Object`
`JavaExecutor` < `AbstractExecutor`
`RubyExecutor` < `AbstractExecutor`
`RubyThreadPoolExecutor` < `RubyExecutor`
`JavaThreadPoolExecutor` < `JavaExecutor`
relying only on the `Synchronization::Object` methods not creating any Mutex instances.

I was planning to take a look at it but you were faster :)

jdantonio:
I should have some more time to work on this tomorrow night. I'll try it the way you suggested and see how it comes out.

pitr-ch:
@jdantonio few comments on the commit.

jdantonio:
See PR #283

jdantonio:
Addressed in #287.

Issue #242, opened 2015-02-13
Migrate form using mutex to synchronization layer


pitr-ch:
see #241

jdantonio:
Addressed in #287.

PR #283, opened 2015-04-28
Synchronization ExecutorService

jdantonio:
Creates a new `ExecutorService` and related series of modules and classes through which all executors use the new synchronization. Additionally:

* Partially addresses #278, #242, and #241
* Renames `PerThreadExecutor` to `SimpleExecutorService`
* Improves the performance of Java-optimized thread pools (#280)
* Updates many classes to use `ns_initialize`
* Updates `IVar` to better support subclasses using `ns_initialize`


jdantonio:
@pitr-ch This PR incorporates your suggestions. I quickly made the updates before work this morning so I expect I probably missed much.

pitr-ch:
@jdantonio Should I review or will you be working on it more first?

jdantonio:
@pitr-ch I consider this a spike and not ready to merge, but please provide your thoughts. I will be working on it more this evening. I prefer this approach over the other two spikes but we should explicitly choose one of the three approaches. I also want to incorporate the updates @headius suggested in #280  and incorporate #284 once it's been merged to master. I was unable to fully review #284 last night (I gave a presentation to a small group of students at the University of Akron) but I plan to review it more fully today. My initial impression is that #284 should go into master first.

lib/concurrent/executor/executor_service.rb
```diff
@@ -0,0 +1,310 @@
...
+  end
+
+  class RubyExecutorService < AbstractExecutorService
+
+    def initialize
+      super()
+      synchronize do
+        @stop_event    = Event.new
+        @stopped_event = Event.new
+      end
```
pitr-ch: if you wait for #284 you can use `ns_initialize` here.

pitr-ch: Correction you can consider those two fields final so:
```ruby
def initialize
  super()
  @StopEvent       = Event.new
  @StoppedEvent = Event.new
  ensure_ivar_visibility!
end
```

lib/concurrent/executor/executor_service.rb
```diff
@@ -0,0 +1,310 @@
...
+
+      def initialize
+        super()
+      end
+
+      def post(*args, &task)
+        raise ArgumentError.new('no block given') unless block_given?
+        return handle_fallback(*args, &task) unless running?
+        executor_submit = @executor.java_method(:submit, [Runnable.java_class])
+        executor_submit.call { yield(*args) }
```
pitr-ch: *This is probably already on your todo to use the Job wrapper here.*

lib/concurrent/executor/java_thread_pool_executor.rb
```diff
@@ -74,7 +75,7 @@ def initialize(opts = {})
         @executor = java.util.concurrent.ThreadPoolExecutor.new(
```
pitr-ch: Not sure how to deal with the assignment `@executor =`  here. This is ok on JRuby (depending on undocumented volatile ivar access) but not ok in general. In general it should be synchronized or fixed by `ensure_ivar_visibility!`

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -71,19 +70,20 @@ class RubyThreadPoolExecutor
     #
     # @see http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html
     def initialize(opts = {})
+      super()
+
       @min_length      = opts.fetch(:min_threads, DEFAULT_MIN_POOL_SIZE).to_i
       @max_length      = opts.fetch(:max_threads, DEFAULT_MAX_POOL_SIZE).to_i
       @idletime        = opts.fetch(:idletime, DEFAULT_THREAD_IDLETIMEOUT).to_i
       @max_queue       = opts.fetch(:max_queue, DEFAULT_MAX_QUEUE_SIZE).to_i
       @fallback_policy = opts.fetch(:fallback_policy, opts.fetch(:overflow_policy, :abort))
```
pitr-ch: same issue with the assignment.

lib/concurrent/executor/executor_service.rb
jdantonio: PRs #284 and #285 will be first, then I'll rebase and update this branch.

lib/concurrent/executor/executor_service.rb
jdantonio: Yep. I tried to implement that change on this branch but it didn't work, which resulted in the conversation in #280. Once the other PRs are merged and I get this branch working with the current implementation I'll add the JRuby improvement.

jdantonio:
The last update implements the performance improvements from #280. The performance difference is significant:

Before:

```
$ ./examples/benchmark_thread_pool_implementations.rb
Rehearsal ----------------------------------------------------------------------
Concurrent::JavaThreadPoolExecutor  12.200000   0.820000  13.020000 (  4.115000)
------------------------------------------------------------ total: 13.020000sec

Concurrent::JavaThreadPoolExecutor   8.040000   0.320000   8.360000 (  2.734000)
```

After:
```
$ ./examples/benchmark_thread_pool_implementations.rb
Rehearsal ----------------------------------------------------------------------
Concurrent::JavaThreadPoolExecutor   2.750000   0.320000   3.070000 (  0.951000)
------------------------------------------------------------- total: 3.070000sec

Concurrent::JavaThreadPoolExecutor   0.610000   0.060000   0.670000 (  0.339000)
```

lib/concurrent/executor/executor.rb
```diff
@@ -65,22 +65,32 @@ def self.executor(executor_identifier)
...
+    #
+    #   @param [Array] args zero or more arguments to be passed to the task
+    #
+    #   @yield the asynchronous task to perform
+    #
+    #   @return [Boolean] `true` if the task is queued, `false` if the executor
+    #     is not running
+    #
+    #   @raise [ArgumentError] if no task is given
+    def post(*args, &task)
```
ianks: Out of curiosity, what kind of args would you imagine being sent to a task?

lib/concurrent/executor/executor_service.rb
```diff
@@ -0,0 +1,330 @@
...
+      @stopped_event = Event.new
+      ensure_ivar_visibility!
+    end
+
+    def post(*args, &task)
+      raise ArgumentError.new('no block given') unless block_given?
+      synchronize do
+        # If the executor is shut down, reject this task
+        return handle_fallback(*args, &task) unless running?
+        execute(*args, &task)
```
ianks: What does `execute` do exavtly? I'm curious why it is in a synchronized block.

lib/concurrent/executor/executor.rb
jdantonio: Any values needed within the task can be passed via `args`. Ruby has closures so any values needed within the task can be accessed via the enclosing context. Doing so, however, requires the runtime to retain the entire closure. This may be inefficient if the closure is large and only a few values are needed. Passing data as arguments instead will allow the runtime to release the closure.

lib/concurrent/executor/executor_service.rb
jdantonio: The `execute` method isn't very well named. This method is responsible for putting the task into the work queue. The `post` method checks for valid arguments and handles overflow when necessary. These operations are identical for all subclasses so we DRY them up but putting `post` in the base class. Each subclass has a different work queue so `execute` is delegated to the subclasses. Both `handle_fallback` and `execute` are synchronized because only one job can be placed into the work queue at any given time.

jdantonio:
## JRuby Executor Performance

See #280 

### Old Implementation

#### jruby-1.7.18

```
Rehearsal ------------------------------------------------------------------------
Concurrent::JavaCachedThreadPool      11.990000   1.380000  13.370000 (  8.653000)
Concurrent::JavaFixedThreadPool        8.400000   0.990000   9.390000 (  6.394000)
Concurrent::JavaSingleThreadExecutor   6.870000   0.100000   6.970000 (  4.003000)
Concurrent::JavaThreadPoolExecutor     6.590000   0.170000   6.760000 (  5.311000)
-------------------------------------------------------------- total: 36.490000sec

Concurrent::JavaCachedThreadPool       7.410000   1.140000   8.550000 (  5.772000)
Concurrent::JavaFixedThreadPool        7.250000   1.020000   8.270000 (  5.242000)
Concurrent::JavaSingleThreadExecutor   4.910000   0.430000   5.340000 (  3.919000)
Concurrent::JavaThreadPoolExecutor     3.850000   0.670000   4.520000 (  3.792000)
```

#### jruby-head

```
Rehearsal ------------------------------------------------------------------------
Concurrent::JavaCachedThreadPool      10.860000   1.270000  12.130000 (  6.242069)
Concurrent::JavaFixedThreadPool        8.030000   0.880000   8.910000 (  4.917759)
Concurrent::JavaSingleThreadExecutor   6.160000   0.290000   6.450000 (  3.971956)
Concurrent::JavaThreadPoolExecutor     4.220000   0.570000   4.790000 (  4.240146)
-------------------------------------------------------------- total: 32.280000sec

Concurrent::JavaCachedThreadPool       4.150000   0.840000   4.990000 (  4.100592)
Concurrent::JavaFixedThreadPool        3.860000   0.830000   4.690000 (  3.580397)
Concurrent::JavaSingleThreadExecutor   3.840000   0.630000   4.470000 (  3.355992)
Concurrent::JavaThreadPoolExecutor     3.760000   0.560000   4.320000 (  3.441682)
```

### New Implementation

#### jruby-1.7.18

```
Rehearsal ------------------------------------------------------------------------
Concurrent::JavaCachedThreadPool       2.980000   0.840000   3.820000 (  1.982000)
Concurrent::JavaFixedThreadPool        1.330000   0.450000   1.780000 (  0.919000)
Concurrent::JavaSingleThreadExecutor   0.570000   0.060000   0.630000 (  0.437000)
Concurrent::JavaThreadPoolExecutor     0.750000   0.270000   1.020000 (  0.761000)
--------------------------------------------------------------- total: 7.250000sec

Concurrent::JavaCachedThreadPool       1.080000   0.680000   1.760000 (  1.038000)
Concurrent::JavaFixedThreadPool        0.920000   0.620000   1.540000 (  0.915000)
Concurrent::JavaSingleThreadExecutor   0.590000   0.340000   0.930000 (  0.543000)
Concurrent::JavaThreadPoolExecutor     0.970000   0.250000   1.220000 (  0.693000)
```

#### jruby-head

```
Rehearsal ------------------------------------------------------------------------
Concurrent::JavaCachedThreadPool       3.460000   0.880000   4.340000 (  2.240732)
Concurrent::JavaFixedThreadPool        1.040000   0.260000   1.300000 (  0.666024)
Concurrent::JavaSingleThreadExecutor   0.760000   0.080000   0.840000 (  0.516168)
Concurrent::JavaThreadPoolExecutor     0.520000   0.150000   0.670000 (  0.429587)
--------------------------------------------------------------- total: 7.150000sec

Concurrent::JavaCachedThreadPool       0.750000   0.190000   0.940000 (  0.681543)
Concurrent::JavaFixedThreadPool        0.820000   0.710000   1.530000 (  0.902916)
Concurrent::JavaSingleThreadExecutor   0.390000   0.150000   0.540000 (  0.328649)
Concurrent::JavaThreadPoolExecutor     0.460000   0.300000   0.760000 (  0.602376)
```

jdantonio:
I believe this PR is ready to merge. The failing test is the `Condition` test that has been causing us problems on Rbx.

lib/concurrent/executor/executor.rb
ianks: Ah yes that's what I was imagining. Good call. We should document this behavior somewhere as well.

pitr-ch: There is also another danger. The access to closure's variables is not synchronized (on JRuby, based on documentation) which may lead to issues since the task is executed on different thread. The args should be protected against that to avoid this issue, e.g. Job object with final fields.

@jdantonio do you know about any sources talking about the closure releasing? I would like to know more. I always assumed that the closure is held because of following situation for example:
```txt
[1] pry(main)> a = 1
=> 1
[2] pry(main)> proc { local_variables }
=> #<Proc:0x007f21885d3850@(pry):2>
[4] pry(main)> proc { |s| eval s }.call 'local_variables'
=> [:s, :a, :__, :_, :_dir_, :_file_, :_ex_, :_pry_, :_out_, :_in_]
```

ext/com/concurrent_ruby/ext/SynchronizationLibrary.java
```diff
@@ -55,14 +55,14 @@ public IRubyObject initialize(ThreadContext context, IRubyObject[] args, Block b
             }
         }
 
-        @JRubyMethod(name = "synchronize", visibility = Visibility.PRIVATE)
+        @JRubyMethod(name = "synchronize", visibility = Visibility.PROTECTED)
```
pitr-ch: Just curious what is behind this change?

lib/concurrent/actor/core.rb
```diff
@@ -199,6 +150,60 @@ def build_context
...
+
+            initialized.success reference if initialized
+          rescue => ex
+            log ERROR, ex
+            @first_behaviour.terminate!
+            initialized.fail ex if initialized
+          end
+        end
+      end
+
```
pitr-ch: Thanks :)

pitr-ch:
:+1:

ext/com/concurrent_ruby/ext/SynchronizationLibrary.java
jdantonio: With those methods private the subclasses didn't work. Every call to `synchronize` on JRuby raised an exception.

lib/concurrent/executor/executor.rb
jdantonio: @pitr-ch The issue of closures came up a long time ago in a conversation another PR or Issue. Someone had written a (very favorable) blog post in which it was noted that arguments couldn't be passed into one of our abstractions. I can't remember which one (I want to say it was Future, but I honestly can't remember). At issue was relying on the closure vs. a scoped variable. I don't remember the details. The result of the conversation was that I added a feature to allow argument passing. I'll see if I can find the Issue/PR.

jdantonio: @ianks @pitr-ch This is the discussion where this came up: https://github.com/ruby-concurrency/concurrent-ruby/issues/223

jdantonio: @ianks @pitr-ch Now that I've read #223 again I recall that the issue wasn't related to releasing the closure, it was due to variable scope.

ext/com/concurrent_ruby/ext/SynchronizationLibrary.java
headius: Can you show me the synchronize calls that were failing? Sounds like we may have a visibility-checking bug somewhere.

jdantonio: @headius Change line 58 to `Visibility.PRIVATE` and run the specs. Or simply load up IRB and use any class that subclasses `Synchronization::Object`. You will get `NoMethodError: private method `synchronize' called for...` any time a subclass calls the `#synchronize` method. Which, from where I'm sitting, is the expected behavior. Private visibility should prevent subclasses from calling a method on a parent whereas protected visibility allows subclasses to access the method. The `Synchronization::Object` class [extends](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/synchronization/object.rb) the `Synchronization::JavaObject` class so protected *is* the correct visibility--given the way we structure things internally.

Since `Delay` is synchronized and we use it to lazy-load the global thread pools, you'll never get the gem to load if you change line 58 above. Commit 01810052ccfc04013035ac0345d13be57cad1d2a is a quick-and-dirty update which changes the visibility of `JavaObject` and changes the visibility of `synchronize

```
$ rspec spec/concurrent/delay_spec.rb
/Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/simplecov-0.9.2/lib/simplecov.rb:31 warning: tracing (e.g. set_trace_func) will not capture all events without --debug flag
NoMethodError: private method `synchronize' called for #<Concurrent::Delay:0x4abf3f0>
     set_deref_options at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/dereferenceable.rb:87
         ns_initialize at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/delay.rb:167
            initialize at com/concurrent_ruby/ext/SynchronizationLibrary.java:54
            initialize at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/delay.rb:72
            initialize at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/utility/processor_count.rb:8
            Concurrent at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/utility/processor_count.rb:143
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/utility/processor_count.rb:4
               require at org/jruby/RubyKernel.java:1071
               require at /Users/jdantonio/.rvm/rubies/jruby-1.7.19/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:54
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/configuration.rb:1
               require at org/jruby/RubyKernel.java:1071
               require at /Users/jdantonio/.rvm/rubies/jruby-1.7.19/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:54
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent/configuration.rb:6
               require at org/jruby/RubyKernel.java:1071
               require at /Users/jdantonio/.rvm/rubies/jruby-1.7.19/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:54
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent.rb:1
               require at org/jruby/RubyKernel.java:1071
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/lib/concurrent.rb:6
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/spec/spec_helper.rb:1
                  each at org/jruby/RubyArray.java:1613
                (root) at /Users/jdantonio/Projects/opensource/concurrent-ruby/spec/spec_helper.rb:23
                (root) at /Users/jdantonio/.rvm/rubies/jruby-1.7.19/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:1
                  each at org/jruby/RubyArray.java:1613
               require at /Users/jdantonio/.rvm/rubies/jruby-1.7.19/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:54
             requires= at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/rspec-core-3.2.3/lib/rspec/core/configuration.rb:1181
             requires= at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/rspec-core-3.2.3/lib/rspec/core/configuration.rb:1181
  process_options_into at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/rspec-core-3.2.3/lib/rspec/core/configuration_options.rb:110
  process_options_into at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/rspec-core-3.2.3/lib/rspec/core/configuration_options.rb:109
             configure at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/rspec-core-3.2.3/lib/rspec/core/configuration_options.rb:22
                 setup at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/gems/rspec-core-3.2.3/lib/rspec/core/runner.rb:96
                  load at org/jruby/RubyKernel.java:1087
                (root) at /Users/jdantonio/.rvm/gems/jruby-1.7.19@concurrent-ruby/bin/rspec:23
```

pitr-ch: @jdantonio it's counter intuitive but private methods can be actually called from sub-classes in Ruby
```ruby
A = Class.new { private def a; :a; end }; B = Class.new(A); B.new.instance_eval {a} # => :a
```
It seemed weird to me too, I should have thought about it more.

headius: The calls you point out are both from within a module, calling synchronize against @mutex, which is presumably the IVar above. I suspect JRuby has a problem with visibility from module methods. Thanks for the info.

headius: Actually I am not sure about this code now. I was unable to get a module to dispatch to a private method even on a hierarchy that had included it. Further, I was unable to get a dotted form (foo.bar) to dispatch to a parent's private methods. Only the bare version worked:

```
[] ~/projects/jruby $ rvm ruby-2.2 do ruby -e "class A; private def bar; end; end; class B < A; def foo; bar; end; end; B.new.foo"

[] ~/projects/jruby $ rvm ruby-2.2 do ruby -e "class A; private def bar; end; end; class B < A; def foo; self.bar; end; end; B.new.foo"
-e:1:in `foo': private method `bar' called for #<B:0x007fc681927300> (NoMethodError)
	from -e:1:in `<main>'

[] ~/projects/jruby $ rvm ruby-2.2 do ruby -e "class A; private def bar; end; end; class B < A; def foo; (a = self; a).bar; end; end; B.new.foo"
-e:1:in `foo': private method `bar' called for #<B:0x007fe82a0df218> (NoMethodError)
	from -e:1:in `<main>'
```

jdantonio: @pitr-ch All these years using Ruby and I never realize that subclasses could access private methods of superclasses--simply because I've never tried it. Your phrase "counter intuitive" is very polite. I like the word "dumb" better. That's not what `protected` is supposed to mean! I'll bring that up with Matz the next time he and I chat. :-) Seriously, though, I had to write some code and run it myself. Sadly, it worked exactly as you described...

@headius Ruby is as Ruby does, I guess. Thanks for taking the time to look into it but for now we'll just keep it protected. Behavior aside, I think `protected` is semantically and logically correct for what we're doing.

PR #284, opened 2015-04-28
Synchronization updates: final and volatile fields, immutable struct

pitr-ch:
- Add `ns_initialize` for ivars setting  inside synchronize block
- cleans up the Synchronization::Object space
- add `ensure_ivar_visibility!` for final fields
- add volatile attributes


pitr-ch:
After applying the new capabilities of `Synchronization::Object` to `Edge::Future` on the obvious places the performance improved. It now has better or same performance as old implementations.
```txt
           value-old    19.844k i/100ms
           value-new    21.473k i/100ms
           value-old    300.837k (±13.9%) i/s -      2.937M
           value-new    295.720k (±18.5%) i/s -      2.813M

Comparison:
           value-old:   300836.7 i/s
           value-new:   295720.0 i/s - 1.02x slower

           graph-old    59.000  i/100ms
           graph-new   231.000  i/100ms
           graph-old    777.004  (±32.9%) i/s -      6.431k
           graph-new      2.211k (±25.2%) i/s -     20.328k

Comparison:
           graph-new:     2211.0 i/s
           graph-old:      777.0 i/s - 2.85x slower

       immediate-old   924.000  i/100ms
       immediate-new   969.000  i/100ms
       immediate-old      9.124k (±23.4%) i/s -     84.084k
       immediate-new      9.134k (±24.7%) i/s -     82.365k

Comparison:
       immediate-new:     9133.6 i/s
       immediate-old:     9123.7 i/s - 1.00x slower

            then-old   402.000  i/100ms
            then-new   410.000  i/100ms
            then-old      4.306k (±28.2%) i/s -     36.984k
            then-new      4.506k (±27.1%) i/s -     40.590k

Comparison:
            then-new:     4505.6 i/s
            then-old:     4306.2 i/s - 1.05x slower
```
Before with only synchronization it was:
```txt
           value-old    17.623k i/100ms
           value-new    11.537k i/100ms
           value-old    293.436k (±15.9%) i/s -      2.837M
           value-new    254.683k (±13.1%) i/s -      2.504M

Comparison:
           value-old:   293436.4 i/s
           value-new:   254682.8 i/s - 1.15x slower

           graph-old    73.000  i/100ms
           graph-new   237.000  i/100ms
           graph-old    863.588  (±26.5%) i/s -      7.592k
           graph-new      2.502k (±18.0%) i/s -     23.700k

Comparison:
           graph-new:     2501.6 i/s
           graph-old:      863.6 i/s - 2.90x slower

       immediate-old   956.000  i/100ms
       immediate-new   678.000  i/100ms
       immediate-old      9.920k (±21.0%) i/s -     92.732k
       immediate-new      6.752k (±13.1%) i/s -     65.766k

Comparison:
       immediate-old:     9919.6 i/s
       immediate-new:     6752.0 i/s - 1.47x slower

            then-old   518.000  i/100ms
            then-new   519.000  i/100ms
            then-old      5.312k (±23.5%) i/s -     48.174k
            then-new      4.431k (±19.5%) i/s -     41.520k

Comparison:
            then-old:     5312.2 i/s
            then-new:     4431.2 i/s - 1.20x slower
```

lib/concurrent/ivar.rb
```diff
@@ -66,6 +66,8 @@ def initialize(value = NO_VALUE, opts = {})
       set(value) unless value == NO_VALUE
     end
 
+    public :synchronize
```
jdantonio: Why is this `public` and not `protected`?

pitr-ch: Good point, thanks. There should be `protected`.

pitr-ch:
Documentation added, see https://github.com/ruby-concurrency/concurrent-ruby/blob/synchronization/doc/synchronization.md

lib/concurrent/synchronization/immutable_struct.rb
```diff
@@ -0,0 +1,67 @@
...
+    #   Person.new 'John Doe', 15
+    #   Person['John Doe', 15]
+    #   Person['John Doe', 15].members # => [:name, :age]
+    #   Person['John Doe', 15].values  # => ['John Doe', 15]
+    class ImmutableStruct < Synchronization::Object
+      def self.with_fields(*names, &block)
+        Class.new(self) do
+          attr_reader(*names)
+
+          class_eval <<-RUBY, __FILE__, __LINE__ + 1
```
jdantonio: What is the advantage of `class_eval` over `define_method`?

lib/concurrent/synchronization.rb
```diff
@@ -6,23 +6,26 @@
...
-                           else
-                             MutexObject
-                           end
+    # @see AbstractObject
+    class Object < Implementation
     end
   end
 end
+
+require 'concurrent/synchronization/immutable_struct'
```
jdantonio: This feels very out of place. It isn't being used anywhere in this file but we're now forcing it downstream.

lib/concurrent/synchronization/immutable_struct.rb
```diff
@@ -0,0 +1,67 @@
+module Concurrent
+  module Synchronization
+    # Similar to Struct but the fields are immutable and always visible (like Java final fields)
+    # @example
+    #   Person = ImmutableStruct.with_fields :name, :age
+    #   Person.new 'John Doe', 15
+    #   Person['John Doe', 15]
+    #   Person['John Doe', 15].members # => [:name, :age]
+    #   Person['John Doe', 15].values  # => ['John Doe', 15]
+    class ImmutableStruct < Synchronization::Object
```
jdantonio: Are we using `ImmutableStruct` anywhere? I'm a big fan of the concept (and I've built similar constructs in other projects) but I can't seem to find a place where we use it.

lib/concurrent/synchronization/immutable_struct.rb
```diff
@@ -0,0 +1,67 @@
...
+      def self.with_fields(*names, &block)
```
jdantonio: I think construction and behavior of these should mirror Ruby's `Struct` class. Again, I'm a big fan of the concept but we should be aware of the user's expectations. This is similar to the conversation regarding `Agent`. `Struct` is part of the Ruby standard library so it's reasonable for a user to expect an `ImmutableStruct` to be "an immutable Struct."

lib/concurrent/synchronization/java_pure_object.rb
```diff
@@ -32,6 +33,10 @@ def ns_signal
           JRuby.reference0(self).notify
           self
         end
+
+        def ensure_ivar_visibility!
+          # relying on undocumented behavior of JRuby, ivar access is volatile
```
jdantonio: How much of a concern is this, that we are depending on undocumented behavior?

lib/concurrent/synchronization/immutable_struct.rb
jdantonio: PS: I'd be happy to update this class after we merge. Having built similar classes in other projects I think it would be fun to work on this one.

lib/concurrent/synchronization/java_pure_object.rb
chrisseaton: I thought we worked out that added a new ivar that increased the size of the object was volatile, but not accessing an existing one?

But whether or not that's true - how about we simply make it documented behaviour by agreeing it with the JRuby team and getting it written down somewhere?

lib/concurrent/synchronization.rb
```diff
@@ -6,23 +6,26 @@
...
 module Concurrent
+  # {include:file:doc/synchronization.md}
   module Synchronization
-    class Object < case
-                           when Concurrent.on_jruby?
-                             JavaObject
+    Implementation = case
+                     when Concurrent.on_jruby?
+                       JavaObject
+                     when Concurrent.on_cruby? && (RUBY_VERSION.split('.').map(&:to_i) <=> [1, 9, 3]) <= 0
```
jdantonio: I'd like to extract `(RUBY_VERSION.split('.').map(&:to_i) <=> [1, 9, 3]) <= 0` into a predicate method on `Engine`. You've done a great job making our code easier to reason about with methods like `Concurrent.on_jruby?`.

lib/concurrent/synchronization/java_pure_object.rb
thedarkone: > But whether or not that's true - how about we simply make it documented behaviour by agreeing it with the JRuby team and getting it written down somewhere?

Are you talking about making all Ruby i-vars volatile? Isn't it going to have terrible performance consequences?

chrisseaton: I'm specifically talking about the case where adding a new ivar to an object in JRuby which causes it to resize the object - that is I believe a volatile operation - I think we spent a while talking about that in the past.

jdantonio:
:+1:

lib/concurrent/synchronization/immutable_struct.rb
pitr-ch: Only performance improvements in some cases, otherwise define_method is cleaner. I wanted to avoid storing instance variables using reflection (which is optimized only on Truffle AFAIK).

lib/concurrent/synchronization.rb
pitr-ch: Made it little better in 4993b2c. It's required here following convention from `atomics.rb` file and similar, but I do not feel strongly about this one way or the other.

lib/concurrent/synchronization/immutable_struct.rb
pitr-ch: Not yet. But soon maybe in the AtomicMarkableReference and for other CAS operations.

lib/concurrent/synchronization/immutable_struct.rb
pitr-ch: One eval removed in 3dbc08f.

lib/concurrent/synchronization/immutable_struct.rb
pitr-ch: Yeah you are right I should have made it to match in the first place. I'll appreciate your help, I am not sure how to deal with the `new` method, where it in one case constructs classes and in the second instances.

lib/concurrent/synchronization/java_pure_object.rb
pitr-ch: I had a conversation on FOSDEM with @headius about the instance variables and it turned out that the access is in fact volatile, by default it uses https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/runtime/ivars/StampedVariableAccessor.java even though the documentation says that they are not. I got the feeling that the reason is to be able to go back to non-volatile ivars if needed. There are also experimental options like `REIFY_VARIABLES` which uses https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/runtime/ivars/VariableAccessorVar0.java where the access is not volatile. (I should probably check for it in the synchronization layer). Advantage of having this layer is that we can react to the changes in the implementations still providing unified behavior, without making changes in the rest of the `concurrent` library.

Just for completeness: ivars are volatile on MRI and JRuby but not on Rubinius. (AFAIK)

Exactly. I am already planning to finish this layer, write down documentation (including the assumptions it builds on) and then contact the implementors to be able to verify the assumptions and therefore soundness of this layer.

lib/concurrent/synchronization.rb
pitr-ch: Good idea done in fe0860e.

lib/concurrent/synchronization/immutable_struct.rb
jdantonio: I'll work on it tonight. I'll be in a hotel room looking for something productive to do (I'm attending [Stir Trek](http://stirtrek.com/) tomorrow).

lib/concurrent/synchronization/immutable_struct.rb
jdantonio: I saw the conversation in #281 after I made this comment. I think it's a good utility to have. Clojure and Erlang have made me greatly appreciate the benefits of immutability.

jdantonio:
:+1: I'm :ok_hand: with merging this. I'll rebase and continue working on #271, #283, and #280 afterwards.

pitr-ch:
Unfortunately there is an almost consistently failing test in rbx, I'll try to fix it first and then merge.

lib/concurrent/synchronization/java_pure_object.rb
headius: The reified variable logic is experimental and not enabled by default, but if we went forward with it we'd need to address the volatility difference.

Normally ivars are stored in an array, and opaque array accesses (same code, many different arrays) don't fold away on JVM because of the boundschecking and indirection involved. Updates of that array are done in a volatile way, as @chrisseaton mentioned. But writes *into* the array also go through the fence/volatile logic @pitr-ch mentioned, so all writes should be volatile. And I *believe* that traversing the volatile ivar array reference for reads forces happens-before visibility on the values (since the JVM can't optimistically optimize the access into a non-volatile cache hit).

We have not made this official for exactly the reason @pitr-ch mentions: no other implementation makes a hard volatility guarantee and we want to be able to pull back a bit on volatility later.

As for performance concerns...JRuby's current ivar strategy already injects so much overhead, making them volatile-by-default may not even be noticeable (we could test). But saying we're always volatile now means we can't improve ivar performance by removing volatility later.

It's a conundrum, to be sure.

pitr-ch:
The failing test is not using `Synchronized::Object` and I cannot get it reproduced :/ If It succeeds at least once on travis I'll merge this branch and add it to intermittently failing tests.

jdantonio:
I would prefer to merge this branch then work on the failing test afterwards:

1. This code needs to be in master for is to proceed with our other work
1. As noted, the failing test appears to be unrelated to the changes in this PR
1. I spent over an hour the other day trying to get Rbx to install on an Ubuntu laptop and wasn't successful, so fixing this test may be very difficult

lib/concurrent/synchronization/java_pure_object.rb
pitr-ch: @headius Is jruby9 bringing any changes in this behavior we should look out for? If so we can add different implementation of `Synchronization::Object` for jruby9.

pitr-ch:
Merging since travis gave a different intermittent error.

jdantonio:
:+1:

jdantonio:
I was able to install Rbx on my MacBook Pro and I've been unable to reproduce the failed test. I think it's safe to call this an intermittent failure and add it to the list.

pitr-ch:
@jdantonio added.

lib/concurrent/synchronization/java_pure_object.rb
headius: No changes at present. Most likely change we'd land after final would be the reified variables, but they'd have to get volatility treatment before release anyway.

thedarkone: > But saying we're always volatile now means we can't improve ivar performance by removing volatility later.

I'd say promising volatile for ivars would be a disastrous decision, that we are going to regret in the future.

PS: I'm not sure JRuby's i-vars are exactly volatile per JVM's spec. Yes, an i-var read always involves a volatile field read, but this is not true for writes, not every i-var write results in a proper Java volatile field write. In practice we don't run into trouble because of x86 and the current JRuby performance/overhead.

pitr-ch: @thedarkone where did you find the volatile field read on ivar read? I've found only https://github.com/pitr-ch/jruby/blob/master/core/src/main/java/org/jruby/runtime/ivars/VariableAccessor.java#L90-L115 where I could not find one, which would indicate that there is no relationship between reads and writes implying that JRuby ivars are not volatile.

Let me try to summarize what I've found out:
- JRuby ivar read does not read any volatile field (like `varTableStamp`). https://github.com/pitr-ch/jruby/blob/master/core/src/main/java/org/jruby/runtime/ivars/VariableAccessor.java#L90-L115
- Var table creation and expansion is protected by read and write to volatile varTableStamp so that is volatile https://github.com/pitr-ch/jruby/blob/master/core/src/main/java/org/jruby/runtime/ivars/StampedVariableAccessor.java#L106-138
- ivar update goes through https://github.com/pitr-ch/jruby/blob/master/core/src/main/java/org/jruby/runtime/ivars/StampedVariableAccessor.java#L151-163 which has fence or `putObjectVolatile` but there is no fence or `getObjectVolatile` on ivar read path. (The fence and `putObjectVolatile` might be useless here.)

It seems that the documentation https://github.com/jruby/jruby/wiki/Concurrency-in-jruby#volatility corresponds with this. cc @headius, could you help with deciphering, I do not know the code-base well I can easily overlook something.

thedarkone: @pitr-ch sorry, you are 100% right.

 * reads are plain Java field and array accesses,
 * writes are volatile read + volatile write.

Therefore for example JVM is allowed to "break" the following code:
```ruby
# EXAMPLE n. 1
class Foo
  def initialize
    @all_systems_go = false
  end
  
  def go!
    @all_systems_go = true
  end
  
  def await_the_go_ahead
    nil until @all_systems_go
    puts "We're good to go!"
  end
end

foo = Foo.new
Thread.new { foo.go! }
foo.await_the_go_ahead # this is allowed to spin infinitely
```

The loop in `nil until @all_systems_go` is allowed to cache and never re-read the value of `@all_systems_go`.

However, if used correctly, the __current implementation__ (so this in no way future proof and is not an official JRuby position) does give us "i-var proof" thread safe construction of objects:

```ruby
# EXAMPLE n. 2
class Foo
  attr_reader :val
  
  def initialize
    @val = :VALUE
  end
end

class Bar
  def setup_foo
    @foo = Foo.new
  end
  
  def foo_val
    nil until foo = @foo
    @foo.val
  end
end

bar = Bar.new
Thread.new { bar.setup_foo }
# might spin infinitely (see example n. 1), however if
# a value is produced it will be :VALUE and never `nil`
# or an exception etc.
bar.foo_val
```

The above example's `bar.foo_val` is allowed to go into an infinite loop (see example n. 1), but if it does return, the return value __will be__ `:VALUE` and not a `nil` and not an exception. This is because of `volatile` semantics combined with JMM's ban on out-of-thin-air values. Note that this applies only to i-vars, so if `Foo`'s constructor was something like this:

```ruby
class Foo
  def initialize # broken
    @val = []
    @val << :VALUE
  end
  
  def good_initialize # not broken
    val = []
    val << :VALUE
    @val = val
  end
end
```

then there are no guarantees about the state of `@val` array for the other thread in `Foo#foo_val`.

PR #286, opened 2015-05-03
MutexAtomicBoolean is now a Synchronization::Object

jdantonio:
## Do Not Merge

Before:

```
Testing with ruby 2.2.1
Testing with Concurrent::MutexAtomicBoolean...
Rehearsal ------------------------------------
   4.930000   0.010000   4.940000 (  4.967906)
--------------------------- total: 4.940000sec

   0.000000   0.000000   0.000000 (  1.021545)
Testing with Concurrent::CAtomicBoolean...
Rehearsal ------------------------------------
   4.330000   0.010000   4.340000 (  4.354824)
--------------------------- total: 4.340000sec

   0.000000   0.000000   0.000000 (  0.000272)
```

After:

```
Testing with ruby 2.2.1
Testing with Concurrent::MutexAtomicBoolean...
Rehearsal ------------------------------------
  20.630000  51.690000  72.320000 ( 54.480074)
-------------------------- total: 72.320000sec

   0.100000   0.000000   0.100000 (  0.100887)
Testing with Concurrent::CAtomicBoolean...
Rehearsal ------------------------------------
   1.260000   0.000000   1.260000 (  1.287014)
--------------------------- total: 1.260000sec

   0.000000   0.000000   0.000000 (  0.000239)
```

### DIFF
--- a/examples/benchmark_atomic_boolean.rb
+++ b/examples/benchmark_atomic_boolean.rb
@@ -6,6 +6,9 @@ require 'concurrent/atomics'
 require 'benchmark'
 require 'rbconfig'
 
+THREADS = 10
+TESTS = 1_000_000
+
 def atomic_test(clazz, opts = {})
   threads = opts.fetch(:threads, 5)
   tests = opts.fetch(:tests, 100)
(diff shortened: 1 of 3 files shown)


jdantonio:
@pitr-ch I've started working on migrating other abstractions to the common synchronization layer, but found some unexpected results. I ran the above comparison using my old 17" MBP (Early 2009). By these tests the version of `AtomicBoolean` that uses `synchronize` rather than `lock / unlock` is more than 10x slower. Can you verify, please?

pitr-ch:
@jdantonio which benchmark did you use?

jdantonio:
[examples/benchmark_atomic_boolean.rb](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/examples/benchmark_atomic_boolean.rb)

pitr-ch:
@jdantonio Thanks, I'll have a look.

jdantonio:
I have no idea what was going on in the above data, but the code in #287 does not exhibit this behavior.

pitr-ch:
Hm that sounds weird :/ I'll still try to do some benchmarks to see if I spot anything bad. We could try to find out if `Unsafe.enterMonitor` and `Unsafe.exitMonitor` are planned to stay (or in different form) in java9 or not, to be able to expose `lock` and `unlock` too (not just `synchronize`) removing the need for closure here. Alternatively we could experiment with `&method(:ns_value)`, it might be optimized better on MRI (I do not know). Personally if it does not bring significant performance boost I would rather stick with just `synchronize`.

Issue #260, opened 2015-03-04
Reconsider Condition usage at some places


pitr-ch:
There is lot of abstractions which are using `Condition` without using the added feature (ability to check if it timed out) which may be speeded up by using original `ConditionVariable`. It some cases the need for `Condition` could be eliminated by manually rechecking if the condition holds after a thread was signaled.

jdantonio:
Once #287 is merged the only remaining use of `Condition` will be in `MVar`. I'll take a look at that sometime before the 1.0 release.

jdantonio:
OK, I lied. I just updated `MVar` to use `ConditionVariable`. This Issue will be closed once we merge #287.

jdantonio:
Addressed in #287.

PR #287, opened 2015-05-03
More Synchronization

jdantonio:
Updated most classes using `Mutex` to be `Synchronization::Object` subclasses. A few classes have not been updated because they do not align neatly with the implementation of `Synchronization::Object`. Specifically:

* `ReadWriteLock` uses two mutexes and two condition variables
* `MutexAtomicReference` is a fallback class only used on runtimes without synchronization primitives
* `BufferedChannel` uses two condition variables with a single mutex
* `MVar` uses two condition variables with a single mutex
* `TVar` uses `#lock` and `#unlock` methods which are not available on `Synchronization::Object`

### DIFF
--- a/lib/concurrent/atomic/atomic_fixnum.rb
+++ b/lib/concurrent/atomic/atomic_fixnum.rb
@@ -35,7 +35,6 @@ module Concurrent
     #   @param [Fixnum] init the initial value
     #   @raise [ArgumentError] if the initial value is not a `Fixnum`
     def initialize(initial = 0)
-      raise ArgumentError.new('initial value must be a Fixnum') unless initial.is_a?(Fixnum)
       super(initial)
     end
 
@@ -58,8 +57,7 @@ module Concurrent
     #
     #   @raise [ArgumentError] if the new value is not a `Fixnum`
     def value=(value)
-      raise ArgumentError.new('value must be a Fixnum') unless value.is_a?(Fixnum)
-      synchronize { @value = value }
+      synchronize { ns_set(value) }
     end
 
     # @!macro [attach] atomic_fixnum_method_increment
@@ -68,7 +66,7 @@ module Concurrent
     #
     #   @return [Fixnum] the current value after incrementation
     def increment
-      synchronize { @value += 1 }
+      synchronize { ns_set(@value + 1) }
     end
 
     alias_method :up, :increment
@@ -79,7 +77,7 @@ module Concurrent
     #
     #   @return [Fixnum] the current value after decrementation
     def decrement
-      synchronize { @value -= 1 }
+      synchronize { ns_set(@value -1) }
     end
 
     alias_method :down, :decrement
@@ -107,7 +105,26 @@ module Concurrent
     protected
 
     def ns_initialize(initial)
-      @value = initial
+      ns_set(initial)
+    end
+
+    private
+
+    def ns_set(value)
+      range_check!(value)
+      @value = value
+    end
+
+    def range_check!(value)
+      if !value.is_a?(Fixnum)
+        raise ArgumentError.new('value value must be a Fixnum')
+      elsif value > MAX_VALUE
+        raise RangeError.new("#{value} is greater than the maximum value of #{MAX_VALUE}")
+      elsif value < MIN_VALUE
+        raise RangeError.new("#{value} is less than the maximum value of #{MIN_VALUE}")
+      else
+        value
+      end
     end
   end
 
--- a/lib/concurrent/atomic/atomic_fixnum.rb
+++ b/lib/concurrent/atomic/atomic_fixnum.rb
@@ -1,4 +1,5 @@
 require 'concurrent/native_extensions'
+require 'concurrent/synchronization'
 
 module Concurrent
 
@@ -21,7 +22,7 @@ module Concurrent
   #         4.520000   0.030000   4.550000 (  1.187000)
   #
   #   @see http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicLong.html java.util.concurrent.atomic.AtomicLong
-  class MutexAtomicFixnum
+  class MutexAtomicFixnum < Synchronization::Object
 
     # http://stackoverflow.com/questions/535721/ruby-max-integer
     MIN_VALUE = -(2**(0.size * 8 - 2))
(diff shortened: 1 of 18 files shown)


lib/concurrent/atomic/semaphore.rb
```diff
@@ -141,12 +145,13 @@ def try_acquire_now(permits)
     end
 
     def try_acquire_timed(permits, timeout)
-      remaining = Condition::Result.new(timeout)
-      while !try_acquire_now(permits) && remaining.can_wait?
-        @condition.signal
-        remaining = @condition.wait(@mutex, remaining.remaining_time)
-      end
-      remaining.can_wait? ? true : false
+      ns_wait_until(timeout) { try_acquire_now(permits) }
```
jdantonio: @pitr-ch Can you please verify that I'm using `ns_wait_until` correctly here? And if I am, I believe I can delete the tests [here](https://github.com/ruby-concurrency/concurrent-ruby/blob/more-synchronization/spec/concurrent/atomic/semaphore_spec.rb#L105) since spurious wakeups are now handled by the base class. Is that correct?

pitr-ch: Sorry I am travelling for few days, so I am slow in replies. Yes that's correct. `ns_wait_until` returns `true` only if the condition holds, or `false` on timeout if supplied. It should not return in any other circumstances.

pitr-ch:
I'll look at the rest on Wednesday.

jdantonio:
@pitr-ch No hurry. I think I've got a handle on it now.

lib/concurrent/atomic/atomic_fixnum.rb
```diff
@@ -74,10 +68,7 @@ def value=(value)
     #
     #   @return [Fixnum] the current value after incrementation
     def increment
-      @mutex.lock
-      @value += 1
-    ensure
-      @mutex.unlock
+      synchronize { @value += 1 }
```
pitr-ch: Shouldn't we put some boundary here based on the maximum `Fixnum` value?

lib/concurrent/executor/safe_task_executor.rb
```diff
@@ -6,17 +6,15 @@ module Concurrent
...
   # reason - the error risen by the callable if it has been executed with errors, nil otherwise
-  class SafeTaskExecutor
+  class SafeTaskExecutor < Synchronization::Object
 
     def initialize(task, opts = {})
-      @task = task
-      @mutex = Mutex.new
-      @exception_class = opts.fetch(:rescue_exception, false) ? Exception : StandardError
+      super(task, opts)
     end
```
pitr-ch: This class could use final fields and no synchronization instead, since it's immutable if I am not mistaken.

spec/concurrent/atomic/atomic_fixnum_spec.rb
```diff
@@ -116,50 +116,31 @@ module Concurrent
...
-    specify 'compare_and_set is synchronized' do
-      mutex = double('mutex')
-      allow(Mutex).to receive(:new).with(no_args).and_return(mutex)
-      expect(mutex).to receive(:lock)
-      expect(mutex).to receive(:unlock)
-      described_class.new(14).compare_and_set(14, 2)
+    context 'instance methods' do
+
+      before(:each) do
+        expect(subject).to receive(:synchronize).with(no_args).and_return(10)
```
pitr-ch: this might just use `and_call_original` to avoid unnecessary stubbing.

lib/concurrent/atomic/atomic_fixnum.rb
ianks: I know Java's int is max'd at 2^31 - 1, so it might make since to cap it at that for compatibility reasons.

On the contrary, if we do that we are re-interpreting what a Fixnum is, since we no longer roll over into BigNum automatically.

lib/concurrent/executor/safe_task_executor.rb
jdantonio: What do you mean by "final fields"?

ianks: I think he is referring to the fact that inside a synchornization object, ivars assigned in camel case are final, and will never be changed. i.e.:

```ruby
@Mutex = Mutex.new
```

lib/concurrent/atomic/atomic_fixnum.rb
jdantonio: I did some performance testing of `FixNum` and `BigDecimal` once and `BigDecimal` is *orders of magnitude* slower. I think it's OK for us to place a cap on this class and let the runtime throw an exception if it gets too big.

ianks: Yeah I agree @jdantonio. If that is the case, maybe we can name it someting other than atomic_fixnum to make it clear the the behavior doesn't mimic that of Fixnum?

lib/concurrent/executor/safe_task_executor.rb
pitr-ch: @ianks, @jdantonio: the `@Mutex` i's one part of it. (It's just my convention though so feel free to ignore.) Second part is that `ensure_ivar_visibility!` has to be called after the final fields are set to make them visible after initialization of the object. There is plenty of examples how it can be used in `edge/future.rb`.

lib/concurrent/atomic/atomic_fixnum.rb
pitr-ch: I depends how it is looked at. For me `AtomicFixnum` means it can only hold `Fixnums` so no overflow supported. If it was supported, the name would be `AtomicInteger`.

chrisseaton: I think the reason it was called `AtomicFixnum` is so that implementations in Java and C can use high performance primitives, which on those platforms do not support overflow.

jdantonio: +1 @pitr-ch and @chrisseaton

lib/concurrent/executor/safe_task_executor.rb
jdantonio: Done.

pitr-ch:
fixes #260 (autoclose)

lib/concurrent/atomic/atomic_fixnum.rb
ianks: My point is was mainly that someone from Ruby land might assume that AtomicFixnum might behave like exactly like a Fixnum that was atomic. This is not the case; they will be sad when their AtomicFixnum maxes out and raises an exception that they were not expecting. To me it just seems like the naming is asking for misinterpretation.

pitr-ch: Yeah I understand your concern, it should be probably mentioned in the documentation that it does not convert, there will always be some space for misunderstanding. Still this is the best naming we have. I guess this will be used by advanced users only so I am not worried that much. It's primarily an Atomic class with suffix describing what type does it supports. There would have to be AtomicBignum to suggest some conversion, but even then it is not compatible with what atomic does - a reference, it cannot auto-convert itself.

jdantonio:
Closes #278, #260, and #242.

PR #289, opened 2015-05-07
Fix broadcast spec bug from rubinius

ianks:
There was a test that kept failing and breaking Travis and I managed to pin down the issue. I noticed both threads we being awoken properly in the spec, however the timeout was causing it to never return. I plan on reporting this to rbx ASAP as well.

### DIFF
--- a/spec/concurrent/atomic/condition_spec.rb
+++ b/spec/concurrent/atomic/condition_spec.rb
@@ -272,6 +272,7 @@ module Concurrent
           it 'wakes up all threads' do
             mutex   = Mutex.new
             go      = CountDownLatch.new(2)
+
             threads = Array.new(2) do
               Thread.new do
                 mutex.synchronize do
@@ -280,12 +281,12 @@ module Concurrent
                 end
               end
             end
+
             go.wait
             mutex.synchronize { subject.broadcast }
+            joined = threads.map(&:join)
 
-            threads.each do |t|
-              expect(t.join(5)).to eq t
-            end
+            expect(joined).to eq threads
           end
         end
       end


pitr-ch:
Great! Thanks, this was really bugging me.

jdantonio:
Huh. I thought I had merged this last night when I merged the "rake task" PR.

@ianks Thanks for taking care of this!

brixen:
I'm looking at the issue filed for Rubinius but I'll comment on this change here.

There are 3 lines removed and 4 lines added (2 of them white space) in this diff. But the only functional change is replacing `Thread#join(5)` with `Thread#join`.

According to the documentation for `Thread#join`, it can return two values: 1. the Thread instance itself, or 2. nil. If a timeout value is provided, and the timeout expires, nil is return.

Since the code changed had a single expectation, and that was precisely that `expect(t.join(5)).to eq t`, the only way for this expectation to fail would be for `t.join(5)` to return nil instead of `t`. In other words, if the call to `t.join(5)` returned before the Thread `t` had finished.

This is a classic race *unless* there is synchronization between whatever `t` is running and the Thread calling `t.join(5)`. You simply *cannot* assume that `t.join(5)` is going to complete after the work in `t`, absent synchronization.

I'm not asserting that there is no synchronization to rule out undesirable interleaving in this code because I haven't inspected all the code, but on the face of it, this spec likely had a race condition.

I'll dissect the proposed reproduction in the Rubinius ticket, but the following illustrates the point I'm trying to make.

```
$ bin/rbx -v -e 'done = false; t = Thread.new { sleep 5; done = true }; s = Time.now; p t.join(5), Time.now - s; s = Time.now; Thread.pass until done; p Time.now - s, t.join'
rubinius 2.5.3.c29 (2.1.0 0e2ace24 2015-05-06 3.5.1 JID) [x86_64-darwin14.4.0]
nil
5.00541
2.2000000000000003e-05
#<Thread:0x1b4 id=8 dead>
```

jdantonio:
@brixen Thank you for the (as always) excellent insight. We know that many of our tests are brittle so it's likely the problem was with the test itself and not Rubinius. Testing concurrency is hard and we've intentionally avoided test doubles to ensure we're truly exercising our code concurrently. It's a daunting task, but we're getting progressively better.

ianks:
@jdantonio Yes I think there was a theoretical race condition with the test but in this case, the actual error we were experiencing was with Rubinius, which was promptly fixed thanks to @brixen!

PR #290, opened 2015-05-07
Concurrent::ConditionVariable

jdantonio:
* Created `Concurrent::ConditionVariable`
* Deprecated `Concurrent::Condition`

### DIFF
--- a/examples/benchmark_atomic_boolean.rb
+++ b/examples/benchmark_atomic_boolean.rb
@@ -6,6 +6,9 @@ require 'concurrent/atomics'
 require 'benchmark'
 require 'rbconfig'
 
+THREADS = 1
+TESTS = 10_000_000
+
 def atomic_test(clazz, opts = {})
   threads = opts.fetch(:threads, 5)
   tests = opts.fetch(:tests, 100)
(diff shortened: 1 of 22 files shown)


jdantonio:
@pitr-ch While updating `MVar` to use `ConditionVariable` from Ruby's standard library I had began thinking about leveraging the *intent* of the `Concurrent::Condition` class we are deprecating. Can you please give me your thoughts on this PR when you have a few minutes? I think this new class greatly increases the utility of `ConditionVariable`.

pitr-ch:
This can be useful for users already using mutexes. But for our library and our classes I would like to add `new_condition` method to `Synchronized::Object` (similar to https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html#newCondition()) which would allow to construct multiple conditions for one `Synchronized::Object`. The condition objects would be already tight to the `Synchronized::Object` instance so it would not need to be passed in as argument (probably less error-prone). It depends if you like this approach too.

jdantonio:
I definitely like the approach you suggest. We have a couple of classes that use multiple conditions (ReadWriteLock immediately comes to mind) that would be able to take advantage of that functionality. Is this something you were planning to do before the 1.0 release? I don't think it's a high priority--we're currently *much* better thanks to your work on the common synchronization layer. It's definitely something that can wait until later if you are busy with other work.

pitr-ch:
I've planned to have it in before it is needed which did not happen because you were amazingly fast with the conversion of the rest of the gem. So I'll do it soon, definitely for `0.9`.

jdantonio:
Cool. I'll close this.

Issue #279, opened 2015-04-24
Move Exchanger and Lazy Register to Edge Gem


jdantonio:
After #274 is merged.

* The current `Exchanger` implementation, which uses `MVar` has a race condition that cannot be fixed with the current approach. It must be rebuilt from scratch without using `MVar`.
* `LazyRegister` has never been documented and has no tests.

pitr-ch:
I would also suggest to move Agents.

jdantonio:
Agent is on of the oldest abstractions in this library. It preceded everything but Future and Promise. And Agent has no known bugs. I don't see how it is "experimental" or "cutting edge."

pitr-ch:
The reason is not bugs but rather that the implementation is incomplete, e.g. #70 which makes it hard to use. Also the integration with STM is missing.

chrisseaton:
Did we want to integrate agents with STM somehow?

jdantonio:
@pitr-ch I respectfully disagree that our Agent is an "incomplete" implementation. It is more limited in functionality than the Clojure abstraction which inspired it, but that doesn't make it incomplete or experimental. Everything in this library was *inspired* by something outside of Ruby, but that's not the same as 100% compatibility with the original source (something which has never been a goal of ours). The conversation from #70 is almost a year old and no one from the community has asked for that additional functionality. I believe that to be a very strong indication that we're talking about "nice to have" functionality and not "critical" functionality.

jdantonio:
@chrisseaton I believe that @pitr-ch is referring to the common synchronization layer (#284, #283, etc.) and not TVar. @pitr-ch Is that correct?

pitr-ch:
Having agents in the gem is great but I think they need some work to be able to go to 1.0, because of that I've suggested moving them to edge temporarily until it is done. There is ton of work to be done so I am trying to bring it down. 

Sorry, I meant TVar, `agent.post`s are postponed in clojure to the end of successful transaction commit. It something like distributed transactions where you are able to commit to DB and send messages as atomic operation. This is somewhat similar to be able to coordinate TVar changes in STM transaction with followup actions in agents. The STM integration is probably nice-to-have (just missing feature). (Even more nice-to-have is ability to integrate other abstractions like actors (postponing message sending).)

[...]

PR #291, opened 2015-05-07
Moved classes to Edge gem.

jdantonio:
Addresses #279

* Agent (moved to edge)
* Condition (deprecated)
* Exchanger (moved to edge)
* LazyRegister (moved to edge)


pitr-ch:
:+1:

jdantonio:
I've updated this PR with a "notravis" tag on the `Exchanger` tests and added an exclusion filter when running tests on Travis. Since we know Exchanger is broken there's no reason our Travis build should be red because of it.

PR #293, opened 2015-05-10
Added WeakKeyMap to ThreadLocalRubyStorage.

jdantonio:
We currently use `WeakKeyMap` from the [Ref](https://github.com/ruby-concurrency/ref) gem. This creates a downstream dependency. Rails requires its dependency gems to have no downstream dependencies. Although the Rails team has indicated that our dependency on Ref isn't an issue, we've heard comments from other users. The need for Ref is very narrow. WeakKeyMap is only used within `ThreadLocalVar` when not on JRuby. WeakKeyMap, in turn, depends on an internal implementation of `WeakReference`, but only on MRI 1.9.3.

Given the very narrow need for Ref I think we're better off copying the necessary code into the core gem as hidden, private, internal classes. In a future release (perhaps 2.0) we can merge the entire Ref gem into concurrent-ruby core.

### DIFF
--- a/lib/concurrent/atomic/thread_local_var.rb
+++ b/lib/concurrent/atomic/thread_local_var.rb
@@ -1,3 +1,5 @@
+require 'concurrent/atomic/thread_local_var/weak_key_map'
+
 module Concurrent
 
   # @!macro [attach] abstract_thread_local_var
@@ -33,12 +35,8 @@ module Concurrent
 
       protected
 
-      unless Concurrent.on_jruby?
-        require 'ref'
-      end
-
       def allocate_storage
-        @storage = Ref::WeakKeyMap.new
+        @storage = WeakKeyMap.new
       end
 
       def get
@@ -123,5 +121,4 @@ module Concurrent
       include ThreadLocalRubyStorage
     end
   end
-
 end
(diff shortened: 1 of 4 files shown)


lib/concurrent/atomic/thread_local_var.rb
```diff
@@ -1,3 +1,5 @@
+require 'concurrent/atomic/weak_key_map'
```
pitr-ch: What would you think about 'lib/concurrent/collection' placement?

pitr-ch: ah it's nested inside the AbstractThreadLocalVar, disregard previous comment. It might be better to keep it in one file though, or moved to sub-dir `concurrent/atomic/thread_local_var/weak_key_map`.

pitr-ch: We should probably copy the MIT license of `ref` at the top.

pitr-ch:
:+1:

lib/concurrent/atomic/thread_local_var.rb
jdantonio: I'll put it in a subdir. I didn't want to embed it directly in the class because it's a lot of code that isn't always activated. When it's in the same file as `AbstractThreadLocalVar` it makes the code for that class much harder to read.

I chose to nest it under `AbstractThreadLocalVar` because it's only needed on a few runtimes. I didn't want to make it a top-level, public class.

PR #294, opened 2015-05-10
Thread-safe Struct Variants

jdantonio:
Implements three thread-safe variations of the standard Ruby `Struct` class, along with benchmark tests.

* `MutableStruct` a thread-safe, mutable struct
* `ImmutableStruct` an immutable struct for which all values are set once at construction
* `SettableStruct` a write-once struct for which each value can be set no more than once, either at construction or any time thereafter


lib/concurrent/struct/abstract_struct.rb
```diff
@@ -0,0 +1,141 @@
...
+    # @!visibility private
+    def ns_equality(other)
+      self.class == other.class && self.values == other.values
+    end
+
+    # @!macro struct_each
+    #
+    # @!visibility private
+    def ns_each
+      values.each{|value| yield value }
```
pitr-ch: This should return enumerator when no method given, see `#to_enum`.

jdantonio: The subclasses do that since some of the subclasses are synchronized and others aren't. I assumed (perhaps incorrectly) that the `#to_enum` call should occur outside of `#synchronize`.

pitr-ch: Sorry never-mind, I've just skipped ahead.

lib/concurrent/struct/immutable_struct.rb
```diff
@@ -0,0 +1,87 @@
...
+  # A thread-safe, immutable variation of Ruby's standard `Struct`.
+  #
+  # @see http://ruby-doc.org/core-2.2.0/Struct.html Ruby standard library `Struct`
+  module ImmutableStruct
+    include AbstractStruct
+
+    # @!visibility private
+    def initialize(*values)
+      ns_initialize(*values)
+    end
```
pitr-ch: I think this is missing `ensure_ivar_visibility!`.

lib/concurrent/struct/final_struct.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    # @!macro struct_equality
+    def ==(other)
+      synchronize { ns_equality(other) }
+    end
+
+    # @!macro struct_each
+    def each
+      return enum_for(:each) unless block_given?
+      synchronize { ns_each(&Proc.new) }
```
pitr-ch: This is quite obscure, why not explicit `&block` argument? I think it would be more readable.

jdantonio: I read a blog post somewhere explaining that this performs slightly better. It had benchmarks to back up the explanation. I'll see if I can find that post. Basically, having `&block` in the function signature and then passing `&block` causes two `Proc` objects to be created, whereas this only creates one.

pitr-ch:
Thinking about the names, what about Mutable, Settable and Immutable ? The FinalStruct and ImmutableStruct are confusing to me.

lib/concurrent/struct/abstract_struct.rb
pitr-ch: yeah, it's perfectly fine. I was commenting on autopilot, I saw each without to_enum and commented without looking further down. Sorry about that.

lib/concurrent/struct/final_struct.rb
pitr-ch: Hm, I am not sure, this is the kind of stuff language implementation should optimize away itself. Unless it's a serious performance hit I would prefer clarity. AFAIK JRuby9 will optimize explicit `&block` and Truffle too.

edbond: See https://github.com/JuanitoFatas/fast-ruby#proccall-vs-yield-code

jdantonio:
@pitr-ch The name `FinalStruct` was intended as a reference to Java's `final` keyword, but I like `MutableStruct`, `SettableStruct`, and `ImmutableStruct` better. I think those names better reveal intent.

lib/concurrent/struct/immutable_struct.rb
jdantonio: Currently `ImmutableStruct` doesn't extend `Synchronization::Object` since it's immutable (and the `#synchronize` call in the initializer creates a significant performance penalty). The only reason it has an `ns_initialize` method is to be consistent with the other struct classes. I could add `ensure_ivar_visibility!` here but I would need to add a class method `ensure_ivar_visibility!` to the synchronization classes. That's easy enough to do if necessary.

jdantonio:
Changed the class names; added a new `#merge` method, and updated the documentation.

lib/concurrent/struct/immutable_struct.rb
pitr-ch: That sounds like I've written it badly. It should not force calling to `synchronize` in cases like this. The default call to `ns_initialize` should be probably removed and left on `Synchronization::Object` users. I would like to avoid moving `ensure_ivar_visibility!` to class because it would create requirement to be implemented on class level which may create complications across different implementations. If you agree I'll remove the auto `ns_initialize` call replacing it with a helper module or a subclass.

lib/concurrent/struct/final_struct.rb
pitr-ch: I did few benchmarks. MRI 2.2:
```txt
  def e1; yield; end    69.720k i/100ms
def e2(&block); block.call; end
                        43.656k i/100ms
  def e1; yield; end      3.782M (±20.5%) i/s -     34.581M
def e2(&block); block.call; end
                        869.372k (±17.0%) i/s -      8.295M

Comparison:
  def e1; yield; end:  3782497.8 i/s
def e2(&block); block.call; end:   869371.9 i/s - 4.35x slower

def e3; e1 &Proc.new; end
                        40.549k i/100ms
def e4(&block); e1 &block; end
                        40.700k i/100ms
def e3; e1 &Proc.new; end
                        806.050k (±16.8%) i/s -      7.704M
def e4(&block); e1 &block; end
                        884.341k (±16.0%) i/s -      8.547M

Comparison:
def e4(&block); e1 &block; end:   884341.4 i/s
def e3; e1 &Proc.new; end:   806049.8 i/s - 1.10x slower
```
JRuby 1.7.19
```txt
  def e1; yield; end   120.512k i/100ms
def e2(&block); block.call; end
                       116.150k i/100ms
  def e1; yield; end      4.408M (±15.4%) i/s -     42.420M
def e2(&block); block.call; end
                          3.349M (±15.1%) i/s -     32.290M

Comparison:
  def e1; yield; end:  4407918.3 i/s
def e2(&block); block.call; end:  3349074.7 i/s - 1.32x slower

def e3; e1 &Proc.new; end
                       100.244k i/100ms
def e4(&block); e1 &block; end
                       100.119k i/100ms
def e3; e1 &Proc.new; end
                          2.758M (±16.2%) i/s -     26.565M
def e4(&block); e1 &block; end
                          3.359M (±16.5%) i/s -     32.238M

Comparison:
def e4(&block); e1 &block; end:  3359376.3 i/s
def e3; e1 &Proc.new; end:  2757612.8 i/s - 1.22x slower
```
Results: There is overhead in MRI for creating procs (4.3x) as mentioned. In this particular case it creates the Proc anyway and `&Proc.new` is little slower in MRI and JRuby. JRuby is able to execute well both `yield` and `block.call` variants. I am inclining to following approach: not doing micro optimizations based on MRI. It would make the code more complicated and I am also assuming that users requiring performance will use JRuby anyway.

pitr-ch: @edbond Thanks for the links and your input.

jdantonio: @pitr-ch Before pushing the latest update I added `#each` calls to the [benchmark script](https://github.com/ruby-concurrency/concurrent-ruby/blob/safe-struct/examples/benchmark_structs.rb#L43) and tested `&block` vs. `&Proc.new`. The former was slower but the difference was minimal. The current update uses `&block` which, I agree, is more intuitive.

lib/concurrent/struct/immutable_struct.rb
jdantonio: If you change the way the constructors work in `Synchronization::Object` we'll have to update every class that extends it. I'm fine with that if you think that the call to `#synchronize` in the initializer should be optional. I'm going to merge this PR today since it works with `Synchronization::Object` as it currently exists. Once you finish your updates to the synchronization layer we can go back and update `ImmutableStruct` along with everything else.

pitr-ch: The problem is that the ImmutableStruct is not safe without `ensure_ivar_visibility!`. Second thread can see the object before its fully initialized. I will update all the child classes with the `Synchronization::Object` changes.

PR #296, opened 2015-05-13
Synchronization 3.update

pitr-ch:
not ready for merge


doc/synchronization.md
```diff
@@ -115,7 +115,52 @@ class Event < Synchronization::Object
...
+
+### Assumptions
+
+Following operations are **assumed** thread-safe, volatile and atomic on all implementations:
+
+-   Class definition
+-   Method definition
+-   Library requirement
+
+It's best practice thought to eager load before going into parallel part of an application.
```
jdantonio: Typo: `though`

doc/synchronization.md
```diff
@@ -115,7 +115,52 @@ class Event < Synchronization::Object
...
+
+### Issues to be aware of
+
+-   **Initialization** - Since instance variables are not volatile and a particular implementation may preinitialize values with nils, based on shapes it already saw, a second thread obtaining reference to newly constructed may still se old preinitialized values instead of values set in `initialize` method. To fix this `ensure_ivar_visibility!` can be used or the object can be safely published in a volatile field.
```
jdantonio: Typo: `may still see`

lib/concurrent/synchronization/condition.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      def wait(timeout = nil)
+        @Lock.synchronize { ns_wait(timeout) }
+      end
+
+      def ns_wait(timeout = nil)
+        synchronize { super(timeout) }
+      end
+
+      def wait_until(timeout = nil, &condition)
```
jdantonio: :+1:

jdantonio:
Have you had a chance to see the request from the Rails team regarding [enhancements to condition variables](https://github.com/rails/rails/pull/14938#discussion_r30153114)? Is that something you think we should add? If so I'd like to do it in v0.9 so that the Rails team can move forward. We can push back the release as long as necessary. Thoughts?

pitr-ch:
Sorry i know this is blocking pre release, but I still did not figure out why it is failing on 2.0. Please wait. I would also like to get the actor updates merged to edge for the pre release although I know it may not be that important.

jdantonio:
We can wait. I greatly appreciate the work you are doing. Can we aim for Monday for the first pre-release, or is that not enough time?

pitr-ch:
I'll have some time on weekend so I would like it to be ready on Monday.

pitr-ch:
@jdantonio so it looks like that the cause is `kill_rogue_threads` :/ I cannot easily remove it, what do we do?

jdantonio:
At LambdaConf with only my phone so I can't easily look at the code or test output, so I'm just throwing out ideas. Have you tried running the test suite without the "rogue threads" hack? I added that very early on because the test suite was weak. We've made a lot of improvements since then. We may be able to remove it.

If the issue is with the rest and not the code you can disable the individual test(s) on Travis. I added an RSpec tag for this. It's used in the Exchanger tests.

pitr-ch:
We had a race commenting, I was updating my comment with more information. I've tries not whole suite but just the failing file and it's fine without it, see https://github.com/ruby-concurrency/concurrent-ruby/pull/296#issuecomment-104917480, but there is still issue. I think best might be to remove the `kill_rogue_threads` completely. I'll probably postpone the pre-release further (fine by me).

jdantonio:
@pitr-ch Fixed.

It turns out that the problem was with the `reset_gem_configuration` helper and `spec/concurrent/configuration_spec.rb`. Once I fixed that I was able to completely remove the `kill_rogue_threads` function`.

Because I had to make changes to changes to `spec_helper.rb` this PR cannot automatically be merged with PR #302. I rebased on a different branch (PR #306) and the tests passed on everything but Rbx. For some reason Rbx hangs when running the `Actor` tests. I cannot reproduce that locally--tests pass for me on all platforms.

I believe we should merge this PR. I'll leave it to you to decide whether to merge or rebase.

PS:

With respect to kill_rogue_threads: I had created that function early on when I was spawning new threads in all the tests. The number of threads in the system would climb by *hundreds* because they weren't effectively being garbage collected. Now that we are using thread pools in the tests (and our thread pools are very stable) this doesn't appear to be a problem anymore.

pitr-ch:
Great, Thanks! I thought this would be much more work. I'll merge master here, I've already based another PR on this branch so I cannot easily rebase. Closing #306 I'll retest here.

jdantonio:
I thought it was going to be a lot more work, too. I honestly thought it would be the bulk of my work between 0.9 and 1.0. I'm *very* happy it ended up being so much easier!

jdantonio:
I know how to fix the tests that are failing on ruby-head. I will fix those after the merge.

pitr-ch:
Cool, the rbx seems happy without reconfiguration. So I think we are almost ready to merge, I'll just recheck the PR.

jdantonio:
Cool. Once you've merged this I'll rebase and merge #307 and #308. After that I'll create pre-release builds of all three gems, publish them, and return to work on #300 and #304.

pitr-ch:
There is only one failure https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/63966616#L4196 otherwise it's ready to merge. @jdantonio I'll leave the decision up to you, if you want to merge it to unblock rest of the work and fix later or fix first.

jdantonio:
That failing test ScheduledTask test is known to be buggy. It's on my list. I'm comfortable merging.

PR #298, opened 2015-05-14
Runtime-optimized Subclass Namespaces

jdantonio:
Addresses #261 

With this release we've developed a new convention for declaring runtime-optimized classes. This first class to follow this convention was `Synchronization::Object`. This updates the following classes to use this convention:

* AtomicBoolean
* AtomicFixnum
* Semaphore
* ThreadLocalVar
* CachedThreadPool
* FixedThreadPool
* SingleThreadExecutor
* ThreadPoolExecutor

NOTE: ExecutorService does *not* follow this convention because it does not follow this pattern. It is a completely different module/class configuration in which only the names are similar.

### DIFF
--- a/lib/concurrent/agent.rb
+++ b/lib/concurrent/agent.rb
@@ -21,7 +21,18 @@ module Concurrent
     #
     # @param [Object] initial the initial value
     #
-    # @!macro executor_and_deref_options
+    # @!macro [attach] executor_and_deref_options
+    #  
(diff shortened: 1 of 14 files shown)


jdantonio:
@pitr-ch I believe this is the last update we planned for v0.9, with the exception of the synchronization layer you are working on in #296.

pitr-ch:
:+1:

PR #305, opened 2015-05-22
Async now hides `new` and provides safe `create` factory method

jdantonio:


lib/concurrent/async.rb
```diff
@@ -42,12 +125,35 @@ def validate_argc(obj, method, *args)
...
+      base.extend(ClassMethods)
+      base.send(:private_class_method, :new)
+      super(base)
+    end
+
+    # @!visibility private
+    module ClassMethods
+      def create(*args, &block)
+        obj = self.send(:new, *args, &block)
+        obj.send(:init_synchronization)
```
pitr-ch: if user calls one of the async methods in initialize (even that it's a bad practice) this ca fail. Swapping the order: `allocate`, `init_synchronization`, `initialize` should fix it.

pitr-ch: Hiding the new method breaks API, the `create` method could just override the original `new`, without the name change.

jdantonio: I know it breaks the API, but I think that overriding `new` is a questionable practice. Using a factory method makes it clear that the object creation process has been modified. I'll add some deprecation warnings to and backward-compatable methods.

lib/concurrent/async.rb
```diff
@@ -42,12 +125,35 @@ def validate_argc(obj, method, *args)
...
+    def self.extended(base)
+      base.extend(ClassMethods)
+      base.send(:private_class_method, :new)
+      super(base)
+    end
+
+    # @!visibility private
+    module ClassMethods
+      def create(*args, &block)
+        obj = self.send(:new, *args, &block)
```
pitr-ch: btw just `new *args, &block` should work here.

PR #308, opened 2015-05-25
Atom

jdantonio:
A Ruby implementation of Clojure's [Atom](http://clojure.org/atoms). "Atoms provide a way to manage shared, synchronous, independent state."


lib/concurrent/atom.rb
```diff
@@ -0,0 +1,127 @@
...
+    # @!macro deref_options
+    # 
+    # @raise [ArgumentError] if the validator is not a `Proc` (when given)
+    def initialize(value, opts = {})
+      @validator = opts.fetch(:validator, ->(v){ true })
+      raise ArgumentError.new('validator must be a proc') unless @validator.is_a? Proc
+
+      @value = Concurrent::AtomicReference.new(value)
+      ns_set_deref_options(opts)
+    end
```
pitr-ch: The instance variable visibility should be enforced here, otherwise it's unsafe.

lib/concurrent/atom.rb
```diff
@@ -0,0 +1,136 @@
...
+    #     values. It must accept one and only one argument which will be the
+    #     intended new value. The validator will return true if the new value
+    #     is acceptable else return false (preferrably) or raise an exception.
+    #
+    #   @!macro deref_options
+    # 
+    #   @raise [ArgumentError] if the validator is not a `Proc` (when given)
+    def initialize(value, opts = {})
+      super()
+      synchronize{ ns_initialize(value, opts) }
```
pitr-ch: it can be either synchronized but then all later access has to be synchronized too, which is not what you want probably. `ensure_ivar_visibility!` will ensure visibility but later access synchronization is not required.

PR #310, opened 2015-05-26
Maybe

jdantonio:
Added a `Maybe` class based on the type of the same name from [Haskell](https://hackage.haskell.org/package/base-4.2.0.1/docs/Data-Maybe.html) and [PureScript](https://github.com/purescript/purescript-maybe/blob/master/docs/Data.Maybe.md).


pitr-ch:
Shouldn't any new feature go to edge first?

jdantonio:
Anything that is experimental, unstable, or with a fast-changing API should go into Edge. But not every feature meets these criteria. New features can still go directly into the core gem.

PR #314, opened 2015-05-31
Futures: passing args on construction, interoperability with channel, documentation

pitr-ch:


lib/concurrent/edge/future.rb
```diff
@@ -43,38 +44,66 @@ def future(default_executor = :io, &task)
...
+      # @return [Future]
+      def delay_on(default_executor, *args, &task)
+        Delay.new(default_executor, *args).future.then(&task)
       end
 
       # Schedules the block to be executed on executor in given intended_time.
       # @return [Future]
-      def schedule(intended_time, default_executor = :io, &task)
-        ScheduledPromise.new(intended_time, default_executor).future.chain(&task)
+      def schedule(intended_time, *args, &task)
```
ianks: Can we list what types can be used with IntendedTime? This may have some implications if used with something like ActiveSupport, for example.

lib/concurrent/edge/future.rb
```diff
@@ -100,25 +128,26 @@ def initialize(promise, default_executor = :io)
         ensure_ivar_visibility!
       end
 
+      # @return [:pending, :completed]
       def state
-        @State.get
+        @State.get.to_sym
```
ianks: Why do we opt to return a sym here? Could this potentially be a source of a memor

lib/concurrent/edge/future.rb
```diff
@@ -511,39 +583,64 @@ def pr_async_callback_on_completion(success, value, reason, executor, callback)
...
+      def apply_value(value, block)
+        block.call(*value)
+      end
+    end
+
+    # A Event which can be completed by user.
     class CompletableEvent < Event
-      # Complete the event
+      # Complete the Event, `raise` if already completed
       def complete(raise = true)
```
ianks: Since `raise` is already a defined method in Ruby, it may be wise to rename this arg.

lib/concurrent/edge/future.rb
pitr-ch: In this case it cannot. `@State.get` is set only internally and it can be only Symbol or special private `Success`of `Failed` objects which are overriding `to_sym` to return `:success` or `:failed` symbols.

lib/concurrent/edge/future.rb
pitr-ch: Thanks, good point.

lib/concurrent/edge/future.rb
pitr-ch: Agre.

lib/concurrent/edge/future.rb
ianks: Ah ok, so if I understand correctly this will only return either `:success` or `:failed`?

pitr-ch: see lines https://github.com/ruby-concurrency/concurrent-ruby/pull/314/files#diff-8a184cb7a6ea329fe4d978de8181531fR132 and https://github.com/ruby-concurrency/concurrent-ruby/pull/314/files#diff-8a184cb7a6ea329fe4d978de8181531fR366

jdantonio:
@pitr-ch Is this ready for merge? I would like to release pre2 this evening.

pitr-ch:
merging, and updating changelog branch. There is some work left, but that can wait after pre2.

PR #322, opened 2015-06-02
Marked private API abstractions.

jdantonio:
According to [Semantic Versioning 2.0.0](http://semver.org/), major and minor version numbers must only be changed when the *public* API changes. Those version numbers do not have to change when private, internal changes are made. According to the spec, software following Semantic Versioning "MUST declare a public API." Yardoc provides an [@api](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#api) tag which can be used to indicate which features are part of the public API and which are part of the private API.

The PR marks many abstractions with the `@api private` tag to clearly indicate which abstractions are being *excluded* from Semantic Versioning.

### DIFF
--- a/lib/concurrent/synchronization/object.rb
+++ b/lib/concurrent/synchronization/object.rb
@@ -1,7 +1,8 @@
 module Concurrent
   module Synchronization
 
-    # @api private
+    # @!visibility private
+    # @!macro internal_implementation_note
     Implementation = case
                      when Concurrent.on_jruby?
                        JavaObject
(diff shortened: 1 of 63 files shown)


pitr-ch:
There are two options:
* `@api private` will generate private warning and it stays in documentation
* `@!visibility private` will not be in documentation at all

I am not sure which one I prefer, just bringing it up for discussion.

jdantonio:
This PR is broken and needs revisited. I didn't look that closely at the generated documentation when I created this PR and I've now discovered a problem. Both `@api private` and `@!visibility private` are transitive. If we mark a superclass or a mixin with either `private` annotations, *every* inherited method is given a warning in the subclass. This isn't the intent.

With respect to the classes that have runtime-specific optimization (such as `AtomicBoolean` I can probably tag those with `@!visibility private` and let the Yardoc macros document the implementation classes. With respect to anything else we may want to consider part of the private API (such as `AbstractStruct`) we may need to use a simple `@note`, or just consider that part of the public API.

pitr-ch:
`@!visibility private` can be overridden children by `@!visibility public`. Same overriding does not work for `@api private`, `@api public` is just another tag. So maybe we could use the `@!visibility private` for this purpose, resulting in omitted documentation for private parts.

jdantonio:
The latest update uses `@!visibility private` more judiciously. It also uses method macros to provide documentation of classes that extend runtime-specific implementations (`AtomicReference` and `Synchronization::Object` are good examples). I think the documentation produced from this update is much better.

I still need to update the documentation for executors to follow the pattern in the latest update.

lib/concurrent/synchronization/object.rb
```diff
@@ -15,8 +18,61 @@ module Synchronization
...
+      #   @!macro synchronization_object_method_ns_signal
+
+      # @!method broadcast
+      #   @!macro synchronization_object_method_ns_broadcast
+
+      # @!method ensure_ivar_visibility!
+      #   @!macro synchronization_object_method_ensure_ivar_visibility
+
+      # @!method self.attr_volatile(*names)
+      #   @!macro synchronization_object_method_self_attr_volatile
```
pitr-ch: Nice! :+1:

pitr-ch:
I like the approach very much. Will `@api private` disappear completely?

jdantonio:
I have no strong preference regarding `@api private`. I see no reason we shouldn't use it in cases where it makes sense (I notice that you used it in Actor a few times). For the runtime-specific classes like `RubyThreadPoolExecutor` I like `@!visibility private` better so I plan to get rid of `@api private` for those.

pitr-ch:
Me neither. Only difference is that in our current configuration `@api private` will show up in the generated documentation, which might be better to keep clean without any implementation bits I think.

jdantonio:
Originally I considered everything public.  That's why the Yard config is the way it is. I've only recently begun considering parts of the gem private. Just before we merge this PR we should consider updating the Yard config.

pitr-ch:
So If I understand correctly we'll have in the generated documentation only methods and classes marked as public, yardopts'll be changed to hide private, protected and api-private.

jdantonio:
I'm not sure. I'm open to suggestions.

One of my goals for 1.0 is 100% documentation. I also want to explicitly mark everything that's part of the private API with either `@!visibility private` or `@api private`. I don't have a strong preference either way. I just want to make sure than anyone looking at the source code is 100% certain which parts of the gems are subject to semantic versioning and which aren't.

jdantonio:
The documentation probably isn't perfect, but it's significantly better than it was before. We'll need to review the docs a few more times before 1.0, but I'm going to merge this PR now then release 0.9.0.pre2.

PR #334, opened 2015-06-08
Cached/FixedThreadPool directly extend ThreadPoolExecutor

jdantonio:
`CachedThreadPool` and `FixedThreadPool` now directly extend `ThreadPoolExecutor` rather than having Ruby/JRuby implementations.


jdantonio:
Thoughts on this approach? It's results in the exact same implementation on every platform but with fewer files, less duplication, and fewer private classes. I'd like to do the same for `FixedThreadPool` and `SingleThreadExecutor`.

jdantonio:
The yardoc macros are not working quite right. The docs for `ExecutorService` aren't displaying correctly. But functionally this works. I'll figure out the yardoc later.

pitr-ch:
:+1:

PR #335, opened 2015-06-08
LockFreeLinkedSet: add initial implementation

ianks:
This PR adds a lock-free set, implemented as a link list. This data-structure is neccesary as a building block for a lock-free concurrent hash table/cache. The only difference is that in a concurrent hash-table, I was use split-ordering on the nodes in the linked list instead ordering the items based on their hash value.

As always, fire away on comments and potential refactorings! Always looking to improve my code.


jdantonio:
@ianks If you don't mind, I'd like to wait until the full release of 0.9 to merge this.

lib/concurrent/edge/lock_free_linked_set.rb
```diff
@@ -0,0 +1,131 @@
+require 'concurrent/edge/lock_free_linked_set/node'
+require 'concurrent/edge/lock_free_linked_set/window'
+
+module Concurrent
+  module Edge
+    class LockFreeLinkedSet
```
pitr-ch: Could you include short description of the class and if possible a link to your source of the algorithm?

lib/concurrent/edge/lock_free_linked_set/node.rb
```diff
@@ -0,0 +1,72 @@
+require 'concurrent/edge/atomic_markable_reference'
+
+module Concurrent
+  module Edge
+    class LockFreeLinkedSet
+      class Node < Synchronization::Object
+        include Comparable
+
+        attr_reader :Data, :SuccessorReference, :Key
```
pitr-ch: This does not feel right `pred.SuccessorReference.compare_and_set ...`. Could you create underscore readers for the CalmelCase ivars?

lib/concurrent/edge/lock_free_linked_set/window.rb
```diff
@@ -0,0 +1,47 @@
+module Concurrent
+  module Edge
+    class Window
```
pitr-ch: Shouldn't this be inside the Set's namespace?

pitr-ch:
And thank you for your great contributions!

lib/concurrent/edge/lock_free_linked_set/node.rb
ianks: Yeah I can do that. TBH I don't know how I like the uppercase convention for final fields. It makes the code reading awkward IMO. I wonder if there is a more explicit way of doing it that doesn't require semantic casing. What are you thoughts?

lib/concurrent/edge/lock_free_linked_set/window.rb
ianks: Yup, nice catch!

ianks:
@pitr-ch @jdantonio 

I have addressed the comments on the PR. Let me know if anything else is needed! :smile:

Issue #318, opened 2015-06-02
Buggy Tests


jdantonio:
- [x] [Concurrent::Promise fulfillment sets the promise state to :fulfilled if the block completes](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/65106242)
- [x] [Concurrent::Edge futures .schedule scheduled execution in graph](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/65112803)
- [x] [Concurrent::Edge futures .schedule scheduled execution](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/74747159)
- [x] [Concurrent::Promise rejection passes the reason to all its children](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/65822390)
- [x] [Concurrent::CyclicBarrier#wait without timeout should release all threads when their number matches the desired one](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/65833846)
- [x] [Concurrent::TimerTask dereferenceable it should behave like dereferenceable defaults :freeze_on_deref to false](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/66419811)
- [x] [Concurrent::ScheduledTask behavior it should behave like obligation #reason is set to error object of the exception when :rejected](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/66483570)
- [x] [Concurrent::Synchronization Concurrent::Synchronization::Object#wait waiting thread is sleeping](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/67094686)
- [x] [Concurrent::Edge futures .schedule scheduled execution](https://ci.appveyor.com/project/rubyconcurrency/concurrent-ruby/build/job/dpfr052yof8lxn89)
- ~~[ ] [Concurrent::Condition with one waiting thread timeout #wait should return 0 or negative number if timed out](https://ci.appveyor.com/project/rubyconcurrency/concurrent-ruby/build/28-master/job/gftyqvkg2qdt21r7)~~ *removed in 1.0*
- ~~[ ] [Executor.executor_from_options returns the global io executor when :operation is false](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/70941053)~~ *removed in 1.0*
[...]

PR #336, opened 2015-06-08
AtomicReference: update try_update API (resolves #330)

### LINKED ISSUE #330: Change AtomicReference API: `try_update` and `try_update!`
see https://github.com/ruby-concurrency/concurrent-ruby/pull/281#issuecomment-109407223

ianks:
Normal `try_update` now returns without raising an exception while `try_update!` will raise an exception.

### DIFF
--- a/lib/concurrent/atomic_reference/direct_update.rb
+++ b/lib/concurrent/atomic_reference/direct_update.rb
@@ -13,7 +13,7 @@ module Concurrent
     # Pass the current value to the given block, replacing it
     # with the block's result. May retry if the value changes
     # during the block's execution.
-    # 
+    #
     # @yield [Object] Calculate a new value for the atomic reference using
     #   given (old) value
     # @yieldparam [Object] old_value the starting value of the atomic reference
@@ -27,9 +27,28 @@ module Concurrent
     # @!macro [attach] atomic_reference_method_try_update
     #
     # Pass the current value to the given block, replacing it
+    # with the block's result. Return nil if the update fails.
+    #
+    # @yield [Object] Calculate a new value for the atomic reference using
+    #   given (old) value
+    # @yieldparam [Object] old_value the starting value of the atomic reference
+    #
+    # @return [Object] the new value, or nil if update failed
+    def try_update
+      old_value = get
+      new_value = yield old_value
+
+      return unless compare_and_set old_value, new_value
+
+      new_value
+    end
+
+    # @!macro [attach] atomic_reference_method_try_update!
+    #
+    # Pass the current value to the given block, replacing it
     # with the block's result. Raise an exception if the update
     # fails.
-    # 
+    #
     # @yield [Object] Calculate a new value for the atomic reference using
     #   given (old) value
     # @yieldparam [Object] old_value the starting value of the atomic reference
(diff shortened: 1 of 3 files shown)


jdantonio:
@ianks Thank for working on this! Can you please add a `@note` to the yardoc indicating the change in behavior? Don't worry about the failing tests. Those tests are on my list (#318).

lib/concurrent/atomic_reference/direct_update.rb
```diff
@@ -27,9 +27,28 @@ def update
...
+    #
+    # @return [Object] the new value, or nil if update failed
+    def try_update
+      old_value = get
+      new_value = yield old_value
+
+      return unless compare_and_set old_value, new_value
+
+      new_value
+    end
```
pitr-ch: We could consider returning `[success, new_value]` because nil cannot be distinguished from new value nil.

jdantonio: We should be somewhat consistent with the similar function in `AtomicMarkableReference`. In that class we return an immutable array with `[new_val, new_mark]`. We could have both methods return a list with true/false as the first element. Having worked in Erlang for a few years I'm very fond of this pattern. It's very common in Erlang to return a tuple of `{ok}` or `{error message}` from functions.

If we want to go in this direction we should consider using either `ImmutableStruct` or `Maybe`. I added the latter specifically for situations like this. Maybe (aka Option) is a common pattern in functional [languages](https://hackage.haskell.org/package/base-4.2.0.1/docs/Data-Maybe.html) and [libraries](https://functionaljava.googlecode.com/svn/artifacts/3.0/javadoc/fj/data/Option.html) that I'm very fond of but that I haven't seen much in Ruby.

ianks: I think that returning `[success, new_value]` might start adding unneccesary verbosity to the API. When you think about it, `AtomicMarkableReference` would have to then return `[success, [new_mark, new_value]]` which seems needlessly complex. 

Another alternative might be to initialize the value with some sort of private `EmptyObject` or something which is used as a placeholder to distinguish from `nil`.

ianks:
I added a note about the update API.

lib/concurrent/atomic_reference/direct_update.rb
jdantonio: In the end, the problem is that idiomatic Ruby abuses `nil` in ways that can make return values ambiguous. That's why I created the obligation API (#value and #reason) methods very early on. It removes the ambiguity. I don't have a strong opinion but I'm comfortable leaving it returning `nil`. For a couple of reasons. Mostly because that's the way the API previously worked. I believe that the baseline assumption is that we *do not* change APIs unless there is a compelling reason to do so. Secondly, returning `nil` is idiomatic Ruby, whether we like it or not. Finally, since the user is passing a block to the function it should be clear when looking at the code if `nil` is a valid return value, and the block can be refactored if necessary.

I'm going to merge this PR now and we can revisit later if we think there may be a better way to handle this return value.

pitr-ch: :+1: We could add a note to the documentation though to warn the users. If they need to distinguish `nil`, `compare_and_set` can be always used. From my side we've considered `[success, new_value]` and it's a bad idea.

PR #339, opened 2015-06-10
Fix call of disable_executor_auto_termination! in deprecated auto_terminate=

iNecas:
It was missed in 349c46ef466ebf1573e0c077b002e4a64fa9a685


jdantonio:
Thank you!

PR #340, opened 2015-06-11
Fix edge Event#wait timeout on ruby 1.9.3

iNecas:
The Monitor#synchronize method catches the break signals, causing the timeouts
not to work (just looping forever)

### DIFF
--- a/spec/concurrent/edge/future_spec.rb
+++ b/spec/concurrent/edge/future_spec.rb
@@ -171,6 +171,22 @@ describe 'Concurrent::Edge futures' do
       expect([queue.pop, queue.pop, queue.pop, queue.pop].sort).to eq [:async, :async, :sync, :sync]
     end
 
+    it 'supports setting timeout while waiting' do
+      start_latch = Concurrent::CountDownLatch.new
+      end_latch = Concurrent::CountDownLatch.new
+
+      future = Concurrent.future do
+        start_latch.count_down
+        end_latch.wait(1)
+      end
+
+      start_latch.wait(1)
+      future.wait(0.1)
+      expect(future).not_to be_completed
+      end_latch.count_down
+      future.wait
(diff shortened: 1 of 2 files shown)


spec/concurrent/edge/future_spec.rb
```diff
@@ -171,6 +171,12 @@
       expect([queue.pop, queue.pop, queue.pop, queue.pop].sort).to eq [:async, :async, :sync, :sync]
     end
 
+    it 'supports setting timeout while waiting' do
+      future = Concurrent.future { sleep 1 }
```
ianks: Can you use a shorter sleep here? We are trying to minimize our usage of sleep in the library, but since we are testing a timeout here using sleep is totally justified, but a shorter value will our test suite run slightly quicker. 😄

jdantonio: The sleep w/i the block shouldn't effect the speed of the tests. This test never waits for the future to finish. The test will end after the 0.1 second wait on line 176.

jdantonio: Another way to handle this more deterministically would be with `CountDownLatch`:

```ruby
start_latch = Concurrent::CountDownLatch.new
end_latch = Concurrent::CountDownLatch.new

future = Concurrent.future do
  start_latch.count_down
  end_latch.wait(1)
end

start_latch.wait(1)
future.wait(0.1)
expect(future).not_to be_completed
end_latch.count_down
```

iNecas: good point, I will update that

iNecas: Updated

jdantonio:
Thank you!

pitr-ch:
Nice catch :+1: Thanks

PR #346, opened 2015-06-12
Fix Edge::Future#value and others when specifying timeouts

iNecas:
The implementation was not expecting the state not being completed, and called
`value` on pending promise

### DIFF
--- a/lib/concurrent/edge/future.rb
+++ b/lib/concurrent/edge/future.rb
@@ -335,11 +335,11 @@ module Concurrent
 
       private
 
-      def wait_until_complete(timeout)
+      def wait_until_complete(timeout, &block)
         while true
-          last_waiter = @Waiters.peek # waiters' state before completion
           break if completed?
 
+          last_waiter = @Waiters.peek # waiters' state before completion
           # synchronize so it cannot be signaled before it waits
           synchronize do
             # ok only if completing thread did not start signaling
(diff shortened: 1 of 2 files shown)


pitr-ch:
Really glad you've caught it!

lib/concurrent/edge/future.rb
```diff
@@ -348,7 +348,9 @@ def wait_until_complete(timeout)
           end
           break
         end
-        self
+        if completed?
+          return block ? block.call : self
+        end
```
pitr-ch: This is too heavy. `ns_wait_until` already returns true/false this method should have been doing the same. There should be `return` before `ns_wait_until` and no break. My bad.

lib/concurrent/edge/future.rb
```diff
@@ -486,22 +488,25 @@ def rejected?
       # @return [Object] the value of the Future when success
       def value(timeout = nil)
         touch
-        wait_until_complete timeout
-        @State.get.value
+        wait_until_complete timeout do
+          @State.get.value
+        end
```
pitr-ch: this should have been just `@State.get.value if wait_until_complete timeout`, which removes the expensive closure here

pitr-ch:
This let me to realize that the return values of wait, etc functions has to be fixed. I'll include this into those fixes.

pitr-ch:
replaced by #349

Issue #350, opened 2015-06-15
MutexAtomic#compare_and_set broken contract


pitr-ch:
It may also return false if another thread is just reading the value, there should be lock not just try_lock:
https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/atomic_reference/mutex_atomic.rb#L45

jdantonio:
I think it has to be `#synchronize` and not `#lock`. The `#lock` method will raise an exception if the lock cannot be acquired. Neither the C nor the Java implementation will ever raise an exception. In order to keep the behavior of the method consistent across all platforms the pure-Ruby code cannot raise an exception.

PS: Thank you @pitr-ch for creating this issue. I saw the comment from @SebastianEdwards in Gitter and was going to create an issue myself. You beat me to it! :-)

iNecas:
I just want to confirm that this seems to cause unexpected issues even when not using the atomic reference directy (we use actors heavily): for example it seems to lead to

```
Concurrent::MultipleAssignmentError: multiple assignment
    /home/travis/.rvm/gems/ruby-2.1.5/bundler/gems/concurrent-ruby-7b78cf2eec5f/lib/concurrent/edge/future.rb:602:in `complete'
    /home/travis/build/Dynflow/dynflow/test/test_helper.rb:249:in `call'
    /home/travis/build/Dynflow/dynflow/test/test_helper.rb:249:in `block (3 levels) in <top (required)>'
    /home/travis/.rvm/gems/ruby-2.1.5/bundler/gems/concurrent-ruby-7b78cf2eec5f/lib/concurrent/edge/future.rb:714:in `complete'
    /home/travis/.rvm/gems/ruby-2.1.5/bundler/gems/concurrent-ruby-7b78cf2eec5f/lib/concurrent/edge/future.rb:797:in `complete'
    /home/travis/.rvm/gems/ruby-2.1.5/bundler/gems/concurrent-ruby-7b78cf2eec5f/lib/concurrent/edge/future.rb:831:in `success'
    /home/travis/.rvm/gems/ruby-2.1.5/bundler/gems/concurrent-ruby-7b78cf2eec5f/lib/concurrent/edge/future.rb:720:in `success'
``` 

or deadlocks:

```
fatal: No live threads left. Deadlock?
[...]

PR #355, opened 2015-06-16
Fixed compare_and_set on MutexAtomicReference.

jdantonio:
Addresses #350 

The pure-Ruby version would return false if the lock could not be obtained. This is distinctly different behavior from the C and Java implementations which only return false if the values did not match. This update brings parity to all implementations.

### DIFF
--- a/lib/concurrent/atomic_reference/mutex_atomic.rb
+++ b/lib/concurrent/atomic_reference/mutex_atomic.rb
@@ -42,14 +42,14 @@ module Concurrent
 
     # @!macro atomic_reference_method_compare_and_set
     def _compare_and_set(old_value, new_value)
-      return false unless @mutex.try_lock
-      begin
-        return false unless @value.equal? old_value
-        @value = new_value
-      ensure
-        @mutex.unlock
+      @mutex.synchronize do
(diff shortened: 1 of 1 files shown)


pitr-ch:
:+1:

ianks:
Shouldn't this comparison be object comparison? i.e. `.equal?`

lib/concurrent/atomic_reference/mutex_atomic.rb
```diff
@@ -42,14 +42,14 @@ def get_and_set(new_value)
...
     # @!macro atomic_reference_method_compare_and_set
     def _compare_and_set(old_value, new_value)
-      return false unless @mutex.try_lock
-      begin
-        return false unless @value.equal? old_value
-        @value = new_value
-      ensure
-        @mutex.unlock
+      @mutex.synchronize do
+        if @value == old_value
```
ianks: Shouldn't this comparison be object comparison? i.e. .equal?

jdantonio: Yes, it should. Nice catch. Thank you! I'll fix in today.

PR #356, opened 2015-06-24
Make sure the Monitor is require before using it

iNecas:
In production environment there is a posibility that the Monitor
is not present in vanila environment, which lead to `require 'concurrent'`
to fail on `Monitor` const missing.


jdantonio:
Thank you!

pitr-ch:
Thanks!

Issue #354, opened 2015-06-16
Synchronization::JavaObject bugs on Java7


pitr-ch:
Volatile write in ensure_ivar_visibility! is not paired with a read.

pitr-ch:
Another one is in the volatile fields, the volatile write is in wrong order.

jdantonio:
@pitr-ch Can this be closed?

PR #357, opened 2015-06-28
Synchronisation fixes

pitr-ch:
* Float assignment is no longer considered atomic.
* We have to discuss 4th commit (it's small please have a look). I do not know about a reliable way how to ensure that full memory barrier is executed on Java 7. Therefore I think it would be best to **support only Java8**.
* related to #354 

cc @jdantonio @chrisseaton

### DIFF
--- a/lib/concurrent/configuration.rb
+++ b/lib/concurrent/configuration.rb
@@ -11,37 +11,6 @@ module Concurrent
   extend Concern::Logging
   extend Concern::Deprecation
 
-  # Suppresses all output when used for logging.
-  NULL_LOGGER = lambda { |level, progname, message = nil, &block| }
-
-  # @!visibility private
-  GLOBAL_LOGGER = AtomicReference.new(NULL_LOGGER)
-  private_constant :GLOBAL_LOGGER
-
-  # @!visibility private
-  GLOBAL_FAST_EXECUTOR = Delay.new { Concurrent.new_fast_executor(auto_terminate: true) }
-  private_constant :GLOBAL_FAST_EXECUTOR
-
-  # @!visibility private
-  GLOBAL_IO_EXECUTOR = Delay.new { Concurrent.new_io_executor(auto_terminate: true) }
-  private_constant :GLOBAL_IO_EXECUTOR
-
-  # @!visibility private
-  GLOBAL_TIMER_SET = Delay.new { TimerSet.new(auto_terminate: true) }
-  private_constant :GLOBAL_TIMER_SET
-
-  # @!visibility private
-  GLOBAL_IMMEDIATE_EXECUTOR = ImmediateExecutor.new
-  private_constant :GLOBAL_IMMEDIATE_EXECUTOR
-
-  def self.global_logger
-    GLOBAL_LOGGER.value
-  end
-
-  def self.global_logger=(value)
-    GLOBAL_LOGGER.value = value
-  end
-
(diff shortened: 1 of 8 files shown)


jdantonio:
I don't know enough about the Java ecosystem to know what percentage of users are on Java 7 vs. Java 8. But I'm also comfortable with us simply providing a warning in the docs for Java 7 users and calling this "good enough." As much as I would love to provide absolute guarantees for everything, the reality is that Ruby--on all platforms--has limitations. It was never designed for concurrency. There is only so much we can do. I hope that, in time, Ruby will get better in these areas. Until then we may have to settle for "the best we can do" in some areas.

ianks:
Why are double (and I assume long) assignments in Java8 guaranteed to be atomic?

ianks:
i.e. Which APIs are available to implement the full memory fence in Java8 which are not available in Java7?

SebastianEdwards:
Inclined to agree with @jdantonio. Java 7 is past it's EOL.

pitr-ch:
@ianks double and long is not atomic on Java any version, as defined in Java memory model. Truffle is able to optimise Float to be just primitive double operations which implies `var = 1.1` won't be atomic assignment. Even though it is (best to my knowledge) atomic on other Ruby implementations I do not want to close this door for optimisation in Truffle backend. Hence I've relaxed the atomicity for Float.

@ianks This JEP describes what was added to Java8 http://openjdk.java.net/jeps/171.

@jdantonio I'll add runtime warning and I'll update documentation, then I think we can release. Sorry for delaying. I'll make the warning in form of deprecation that Java7 support will be removed in next 1.0 release.

jdantonio:
No worries on the delay. This is a very important release and we want to make sure to be thorough.

jdantonio:
Is this ready to merge?

pitr-ch:
Not yet, I am doing it asap. So we can release.

jdantonio:
OK. This weekend is a holiday weekend in the US (Independence Day) so I won't have much free time over the weekend. I may not have time to do the release until Sunday evening so no rush.

pitr-ch:
Sorry I still did not have time to finish it, I hope tomorrow evening.

pitr-ch:
@jdantonio Done. But I was in a hurry, sorry to bother you, but could you double check?

jdantonio:
I was out of town this past weekend, visiting family. I will take a look tonight.

lib/concurrent/configuration.rb
```diff
@@ -62,17 +31,48 @@ def self.create_stdlib_logger(level = Logger::FATAL, output = $stderr)
...
-      logger.add level, message, progname, &block
-    end
+    Concurrent.global_logger = create_stdlib_logger level, output
   end
 
+  # Suppresses all output when used for logging.
+  NULL_LOGGER   = lambda { |level, progname, message = nil, &block| }
+
+  # @!visibility private
+  GLOBAL_LOGGER = AtomicReference.new(create_stdlib_logger(Logger::WARN))
```
jdantonio: Why the change from `NULL_LOGGER`? Logging is a side-effect. In general, I prefer not to create unnecessary side-effects and, instead, let the user decide for themselves if they want to log.

pitr-ch: If the logger is disabled by default than we cannot deliver warnings and errors to our users. I am also not big fan of side-efects, but for logging I think it's acceptable. We log at only very few places anyway.

PR #359, opened 2015-07-07
Add rake-compiler-dock for building Windows binary gems.

larskanis:
There is already a section for cross building the extension gem for windows, but no further information how this is done. I would like to contribute this in form of the new rake task 'build:windows'.

The intention of [rake-compiler-dock](https://github.com/larskanis/rake-compiler-dock) is to make cross compiling windows gems simple and reliable.

I tested building the fat binary concurrent-ruby-ext gems successfully on Linux and Windows. I also tested the resulting gems on RubyInstaller versions 2.2.1-x86 and 2.2.1-x64 on Windows-7 successfully.

I was unsure whether 'build:windows' should be part of 'build'. And I noted that the 'pkg' directory isn't part of the clean list. In my experience the 'pkg' directory is sometimes messed and should be cleared (at least when cross building). Can I change 'pkg/classes' to 'pkg' ?

PR #360, opened 2015-07-07
Add CI-tests on Windows.

larskanis:
The github repository is not yet enabled for CI-tests on appveyor. This can be done by adding concurrent-ruby as a new project on http://appveyor.com .

The SVG image URL must be replaced by the one of the owner the ruby-concurrency account - it shows the status for my repository, currently.

I did not enable all possible Ruby targets, because appveyor is slow...

In theory all tests should pass, but there are some timing dependent specs, that seems to fail like [here](https://ci.appveyor.com/project/larskanis/concurrent-ruby)


ianks:
Do appveyor builds run concurrently?

larskanis:
Not on the free accounts.

nirvdrum:
For anyone else confused, it looks like the correct URL is http://www.appveyor.com/.

larskanis:
Yes, its .com, sorry!

jdantonio:
I didn't know that AppVeyor existed, so thank you @larskanis for the info. I've started experimenting with it over [here](https://ci.appveyor.com/project/jdantonio/concurrent-ruby) using PR #361. If it works as expected I'll integrate once I've looked over and merged #359.

jdantonio:
I created an AppVeyor account for the `ruby-concurrency` org and configured the necessary webhook. I also manually ran a job against @larskanis fork. It seems to [work very well](https://ci.appveyor.com/project/rubyconcurrency/concurrent-ruby). Unfortunately AppVeyor drops the dash from "ruby-concurrency" when it creates the org URL, so the org path is "rubyconcurrency."

I expect to merge this later this evening, once I've left work.

ianks:
Should we set an environment variable to prevent Appveyor from using Coveralls?

jdantonio:
We can. I'm fairly ambivalent, to be honest. I've given up on Coveralls for this gem. I removed the badge from the README. Because of the platform-specific code our test coverage reports are always artificially low. Until we can figure out how to get a more accurate coverage score out of Coveralls, it doesn't have much more value than simply running the tests locally using simplecov.

jdantonio:
@larskanis My Internet access is down because my provider sent me new, faulty equipment. I only have my phone right now. I'll work on this as soon as I get my Internet issue resolved.

larskanis:
@jdantonio My internet provider took me offline for almost half a year. I'll keep my fingers crossed and hope that your issue will be solved faster...

PR #364, opened 2015-07-09
Build updates

jdantonio:
Updated rake tasks, spec_helper, etc. to better support AppVeyor and docker-based builds. See #359 and #360.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -42,8 +42,7 @@ Dir.glob('tasks/**/*.rake').each do |rakefile|
 end
 
 def has_docker?
-  system("docker version > /dev/null 2>&1")
-  0 == $?.exitstatus
+  system("docker version > /dev/null 2>&1 || boot2docker version > /dev/null 2>&1")
 end
 
 if Concurrent.on_jruby?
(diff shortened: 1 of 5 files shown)


jdantonio:
@larskanis Whenever AppVeyor runs it triggers two builds. I probably did something wrong in the setup. Any idea what I may have done wrong? I'd be happy to give you access to the AppVeyor settings if you would be willing to help.

Rakefile
```diff
@@ -41,6 +41,11 @@ Dir.glob('tasks/**/*.rake').each do |rakefile|
   safe_load rakefile
 end
 
+def has_docker?
+  system("docker version > /dev/null 2>&1")
+  0 == $?.exitstatus
+end
+
```
larskanis: `docker version` fails, when boot2docker is installed, but not started. In this case rake-compiler-dock will use boot2docker to start the VM on it's own. The `system` method already returns true or false. So you might use:
``` ruby
def has_docker?
  system("docker version > /dev/null 2>&1 || boot2docker version > /dev/null 2>&1")
end
```

larskanis:
@jdantonio I didn't had such an effect so far. But I noticed, that it's possible to add one and the same project twice in the [project view](https://ci.appveyor.com/projects) . If this is the case you can remove the second project in the settings.

jdantonio:
@larskanis Somehow I managed to setup the webhook twice in GitHub. That looks like it was the problem.

PR #371, opened 2015-07-14
Removed unnecessary extension warnings.

jdantonio:
The two warnings removed in this PR would only show once them gem was loaded, and not during installation. They informed the user of possible performance improvements even though the gem was fully functional. I'm not certain it's appropriate to force this warning onto projects that elect to use this gem (such as [Rails](https://github.com/rails/rails/pull/20866)).

@pitr-ch Thoughts?

/cc @tenderlove

### DIFF
--- a/lib/concurrent/utility/native_extension_loader.rb
+++ b/lib/concurrent/utility/native_extension_loader.rb
@@ -25,9 +25,6 @@ module Concurrent
             # may be a Windows cross-compiled native gem
             require "concurrent/#{RUBY_VERSION[0..2]}/extension"
             @c_ext_loaded = true
-          end,
-          lambda do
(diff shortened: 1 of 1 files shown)


tenderlove:
Yes, please!

pitr-ch:
Yeah makes sense. README is probably enough.

ianks:
Should we consider putting a message about it in the gemspec? I think most of the times those messages are annoying, but in this case it does actually serve a purpose. Mostly I am concerned about the average user not caring enough to rtfm for concurrent-ruby, and this might help with that. Thoughts?

jdantonio:
@ianks The Rails team specifically requests that gems used by Rails do *not* generate a post-install messages.

PR #372, opened 2015-07-17
ReadWriteLock feature parity with Rails ShareLock

jdantonio:
We have a potential opportunity to replaces the [ShareLock](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/concurrency/share_lock.rb) class in Rails with our `ReadWriteLock`. Before we can do so, we need to add features and benchmark against the Rails version. Since the Rails version makes extensive use of locks it's likely we will perform well.

The PR adds all necessary features except "loose upgrades", which may be a bit challenging.

Specific Updates:

* `AtomicFixnum` now supports `AtomicDirectUpdate` (some of those methods were needed by `ReadWriteLock`)
* The counter in `ReadWriteLock` has been changed from `AtomicReference` to `AtomicFixnum`
* `ReadWriteLock` is now reentrant
* Attempts to acquire a write lock can, optionally, return `false` immediately if the write lock is being held.

### DIFF
--- a/lib/concurrent/atomic/read_write_lock.rb
+++ b/lib/concurrent/atomic/read_write_lock.rb
@@ -1,5 +1,5 @@
 require 'thread'
-require 'concurrent/atomic/atomic_reference'
+require 'concurrent/atomic/atomic_fixnum'
 require 'concurrent/errors'
 require 'concurrent/synchronization'
 
@@ -54,7 +54,7 @@ module Concurrent
 
     # Create a new `ReadWriteLock` in the unlocked state.
     def initialize
-      @Counter   = AtomicReference.new(0) # single integer which represents lock state
+      @Counter   = AtomicFixnum.new(0) # single integer which represents lock state
       @ReadLock  = Synchronization::Lock.new
       @WriteLock = Synchronization::Lock.new
       ensure_ivar_visibility!
@@ -125,11 +125,11 @@ module Concurrent
             if running_writer?(c)
               @ReadLock.wait_until { !running_writer? }
             else
-              return if @Counter.compare_and_swap(c, c+1)
+              return if @Counter.compare_and_set(c, c+1)
             end
           end
         else
-          break if @Counter.compare_and_swap(c, c+1)
+          break if @Counter.compare_and_set(c, c+1)
         end
       end
       true
@@ -141,7 +141,7 @@ module Concurrent
     def release_read_lock
       while true
         c = @Counter.value
-        if @Counter.compare_and_swap(c, c-1)
+        if @Counter.compare_and_set(c, c-1)
           # If one or more writers were waiting, and we were the last reader, wake a writer up
           if waiting_writer?(c) && running_readers(c) == 1
             @WriteLock.signal
@@ -169,8 +169,8 @@ module Concurrent
 
         if c == 0 # no readers OR writers running
           # if we successfully swap the RUNNING_WRITER bit on, then we can go ahead
-          break if @Counter.compare_and_swap(0, RUNNING_WRITER)
-        elsif @Counter.compare_and_swap(c, c+WAITING_WRITER)
+          break if @Counter.compare_and_set(0, RUNNING_WRITER)
+        elsif @Counter.compare_and_set(c, c+WAITING_WRITER)
           while true
             # Now we have successfully incremented, so no more readers will be able to increment
             #   (they will wait instead)
@@ -187,7 +187,7 @@ module Concurrent
             # Then we are OK to stop waiting and go ahead
             # Otherwise go back and wait again
             c = @Counter.value
-            break if !running_writer?(c) && !running_readers?(c) && @Counter.compare_and_swap(c, c+RUNNING_WRITER-WAITING_WRITER)
+            break if !running_writer?(c) && !running_readers?(c) && @Counter.compare_and_set(c, c+RUNNING_WRITER-WAITING_WRITER)
           end
           break
         end
--- a/lib/concurrent/atomic/read_write_lock.rb
+++ b/lib/concurrent/atomic/read_write_lock.rb
@@ -1,5 +1,6 @@
 require 'thread'
 require 'concurrent/atomic/atomic_fixnum'
+require 'concurrent/atomic/atomic_reference'
 require 'concurrent/errors'
 require 'concurrent/synchronization'
 
@@ -54,9 +55,10 @@ module Concurrent
 
     # Create a new `ReadWriteLock` in the unlocked state.
     def initialize
-      @Counter   = AtomicFixnum.new(0) # single integer which represents lock state
-      @ReadLock  = Synchronization::Lock.new
-      @WriteLock = Synchronization::Lock.new
+      @Counter    = AtomicFixnum.new(0) # single integer which represents lock state
+      @ReadLock   = Synchronization::Lock.new
+      @WriteLock  = Synchronization::Lock.new
+      @WriterLock = AtomicReference.new(nil)
       ensure_ivar_visibility!
       super()
     end
@@ -162,6 +164,7 @@ module Concurrent
     # @raise [Concurrent::ResourceLimitError] if the maximum number of writers
     #   is exceeded.
     def acquire_write_lock(no_wait = false)
+      return true if @WriterLock.value == Thread.current
       while true
         c = @Counter.value
         return false if no_wait && running_writer?(c)
@@ -192,6 +195,7 @@ module Concurrent
           break
         end
       end
+      @WriterLock.set(Thread.current)
       true
     end
 
@@ -200,6 +204,7 @@ module Concurrent
     # @return [Boolean] true if the lock is successfully released
     def release_write_lock
       c = @Counter.update { |c| c-RUNNING_WRITER }
+      @WriterLock.set(nil)
       @ReadLock.broadcast
       @WriteLock.signal if waiting_writers(c) > 0
       true
--- a/lib/concurrent/atomic/read_write_lock.rb
+++ b/lib/concurrent/atomic/read_write_lock.rb
@@ -82,6 +82,9 @@ module Concurrent
 
     # Execute a block operation within a write lock.
     #
+    # @param [Boolean] no_wait When true will return immediately if the lock cannot
+    #   be acquired without waiting. When false will wait for the lock.
+    #
     # @yield the task to be performed within the lock.
     #
     # @return [Object] the result of the block operation.
@@ -89,9 +92,9 @@ module Concurrent
     # @raise [ArgumentError] when no block is given.
     # @raise [Concurrent::ResourceLimitError] if the maximum number of readers
     #   is exceeded.
-    def with_write_lock
+    def with_write_lock(no_wait = false)
       raise ArgumentError.new('no block given') unless block_given?
-      acquire_write_lock
+      return nil unless acquire_write_lock(no_wait)
       begin
         yield
       ensure
@@ -151,13 +154,17 @@ module Concurrent
 
     # Acquire a write lock. Will block and wait for all active readers and writers.
     #
-    # @return [Boolean] true if the lock is successfully acquired
+    # @param [Boolean] no_wait When true will return immediately if the lock cannot
+    #   be acquired without waiting. When false will wait for the lock.
+    #
+    # @return [Boolean] true if the lock is successfully acquired else false
     #
     # @raise [Concurrent::ResourceLimitError] if the maximum number of writers
     #   is exceeded.
-    def acquire_write_lock
+    def acquire_write_lock(no_wait = false)
       while true
         c = @Counter.value
+        return false if no_wait && running_writer?(c)
         raise ResourceLimitError.new('Too many writer threads') if max_writers?(c)
 
         if c == 0 # no readers OR writers running
--- a/lib/concurrent/atomic/read_write_lock.rb
+++ b/lib/concurrent/atomic/read_write_lock.rb
@@ -203,7 +203,7 @@ module Concurrent
     #
     # @return [Boolean] true if the lock is successfully released
     def release_write_lock
-      c = @Counter.update { |c| c-RUNNING_WRITER }
+      c = @Counter.update { |count| count-RUNNING_WRITER }
       @WriterLock.set(nil)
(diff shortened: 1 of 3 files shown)


jdantonio:
## ReadWriteLock Benchmark

### 0.9.0 master

```
READ INTENSIVE (80% read, 20% write):
Testing Concurrent::ReadWriteLock with 80 readers and 20 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
  0.480000   0.910000   1.390000 (  1.492259)
WRITE INTENSIVE (80% write, 20% read):
Testing Concurrent::ReadWriteLock with 20 readers and 80 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
  0.630000   1.050000   1.680000 (  5.434377)
BALANCED (50% read, 50% write):
Testing Concurrent::ReadWriteLock with 50 readers and 50 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
  0.720000   1.390000   2.110000 (  3.460515)
```

### With Rails-supporting Updates

```
READ INTENSIVE (80% read, 20% write):
Testing Concurrent::ReadWriteLock with 80 readers and 20 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
  0.470000   0.920000   1.390000 (  1.488184)
WRITE INTENSIVE (80% write, 20% read):
Testing Concurrent::ReadWriteLock with 20 readers and 80 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
  0.630000   1.050000   1.680000 (  5.451514)
BALANCED (50% read, 50% write):
Testing Concurrent::ReadWriteLock with 50 readers and 50 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
  0.720000   1.410000   2.130000 (  3.454151)
```

jdantonio:
@pitr-ch @alexdowad -- If you have an opportunity, can you please take a look at this? I'm 99% confident in the updates, but this is a complex class so a review would be very helpful.

lib/concurrent/atomic/read_write_lock.rb
```diff
@@ -192,7 +203,8 @@ def acquire_write_lock
     #
     # @return [Boolean] true if the lock is successfully released
     def release_write_lock
-      c = @Counter.update { |c| c-RUNNING_WRITER }
+      c = @Counter.update { |count| count-RUNNING_WRITER }
+      @WriterLock.set(nil)
```
ianks: Does this @WriterLock need to be set before the Counter is updated? From what I can tell, setting the counter represents your linearization point; meaning that once that is set another thread could conceivably acquire the lock, set the @WriterLock to Thread.current, then *after* that we set @WriterLock to be nil. This could cause a deadlock if the thread attempted to acquire the lock multiple times without releasing it.

ianks: EDIT: When I said "another thread could conceivably acquire to lock", I meant that another thread could conceivably "pass the readers writers test once he counter." 

The flow would be:

1. Thread A  acquires writers lock
2. Thread B attempts to acquire writers locks, gets stuck waiting for Thread A
3. Thread A unsets WAITING_RUNNER
4. Thread B sets @WriterLock to Thread.current
5. Thread A sets @WriterLock to nil
6. Thread B calls acquire lock again, compares @Writer lock to Thread.current (which is nil), then spins waiting *forever* for the RUNNING_WRITER bit to be unset. Thread B can never progress.

alexdowad: Good catch, @ianks! What you say is absolutely right.

(Your first comment is an example of what I said about this PR: the name `WriterLock` is confusing. Something like `WriteLockHolder` would be more accurate.)

alexdowad:
Sorry, I just realized my comment about the reader lock was inaccurate. I believe the reader lock is already re-entrant (though I haven't looked at the code for a while and would have to confirm this).

jdantonio:
@alexdowad @ianks Thank you both for the feedback. This is extremely helpful. A few comments, in no specific order:

* Tracking the number of times the writer thread acquires the lock makes sense (re-entrant). The Rails lock class simply stores the Thread object, which is why I took that same approach. But as I think through the problem I see why it needs to be done the way you suggest.
* I agree that we can use better names now that we are adding additional functionality. I'll use the suggestions above.
* I'll update the locking algorithm per the suggestion above.
* I'm considering making re-entrant a constructor option. I'm going to benchmark the performance of every change. I'm going to do my best to ensure we're as fast as possible.

jdantonio:
@pitr-ch Now it's my turn to push back a release :-( Now that the Rails team has [formally announced](http://weblog.rubyonrails.org/2015/7/17/this-week-in-rails-goodbye-rack-lock-concurrent-ruby-better-ac-parameters-and-more/) that concurrent-ruby is being used by Rails, I'd like to continue making PRs to Rails. I'd like this lock in the next release so that we can make another PR to Rails. I'm going to work on this over the weekend, with hopes to have it ready by Monday.

alexdowad:
I was a bit incredulous that Rails' "re-entrant" lock could really be that buggy, so I took a look at their implementation. It is, in fact, re-entrant, and it does achieve this by counting the number of times it has been acquired. Look at how it uses `@exclusive_depth`.

I've noticed that they also allow upgrading the read lock to a write lock. They don't require a special method call to do this; you just call `#start_exclusive` while holding a read lock. This is a very useful feature; @jdantonio, while you are doing this work, I would suggest that you add a spec to ensure that `ReadWriteLock` allows the same. (Right now, I don't think it does.) This is important if `ReadWriteLock` is to replace `ShareLock`.

They have a feature called "loose upgrades" which makes a thread temporarily drop its read lock while attempting to upgrade to a write lock. At first look, this seems ill-advised, and I wonder if Rails actually relies on this feature. If it does, that is probably symptomatic of something else which can be improved.

alexdowad:
About automatically upgrading read locks to write locks: this is only possible if `ReadWriteLock` *knows* which threads are holding read locks. Right now it knows the *number* of read locks which are held, but not *who* is holding them. Making it keep a record of which threads are holding read locks will of necessity increase its memory footprint and slow it down.

The alternative is to add explicit "upgrade" and "downgrade" methods. "Upgrade" will be very similar to the current `#acquire_write_lock`, with the difference that the write lock can be acquired if the number of readers is 1, and additionally, the same atomic swap which "grabs" the write lock should simultaneously "release" the previously held read lock. (So just adjust the counter by `+RUNNING_WRITERS-RUNNING_READERS`.) "Downgrade" will be similar to `#release_write_lock`, but will use an atomic swap to adjust the counter by `-RUNNING_WRITERS+RUNNING_READERS`, and waiting writers will *not* be released by a downgrade (waiting readers will).

If you want these "upgrade" and "downgrade" methods, let me know and I can send a PR.

jdantonio:
@alexdowad Any help you wish to provide will be greatly appreciated! Your feedback so far has been invaluable. Any code you wish to contribute will be a huge boon.

alexdowad:
@jdantonio Do you like the idea of explicit "upgrade" and "downgrade" methods? How will those interact with nested acquires/releases? (Tricky question!)

Or do you like the idea of tracking all the threads which hold a read lock, so we can make acquire/release of a write lock work "automatically" *even when* nested inside a region which uses a read lock?

Nesting acquisition/release of a read lock inside a region which uses a write lock seems easy enough -- those simply become no-ops.

alexdowad:
I think I have the answer to that "tricky question" -- in any case where a write lock is already held, a nested "upgrade" is the same as a nested acquisition (it just increments what Rails calls the `@exclusive_count`). A nested "downgrade" is the same as a nested release (it just decrements the `@exclusive_count`).

That will make methods which use these locks composable in a reasonably intuitive manner.

We are still left with the issue that regions which use a read lock can be nested inside those which use a write lock, but not the other way around. When you are already holding a read lock, you have to be aware of that and "upgrade" it rather than trying to acquire a write lock in the normal manner. Hopefully that is something which people can deal with?

jdantonio:
@alexdowad I trust your judgement on this. What you suggest sounds perfectly reasonable to me. Once you update the code with your suggestions I'll start working on a PR for Rails that replaces their lock with ours and run the full test Rails test suite. If everything passes then we'll explain the behavioral differences in the PR to Rails.

Thanks again for your help with this! I'm participating in a hackathon all weekend (Cleveland GiveCamp) so I don't have a lot of time to spend with the code right now. I cannot understate how beneficial it is having you jump in and provide this help.

alexdowad:
@jdantonio Would you like to commit some or all of the code in this PR, so that I will then build on it? Or would you like me to start with the current `master`?

pitr-ch:
@jdantonio sure I'll wait.

ianks:
How much of a performance hit would keeping a record of who is holding read locks actually be? I intuit that it wouldn't be bad in typical Rails workloads, but I could be long. Also, I have implemented a non-blocking set data structure for concurrent-ruby which would minimize sequential bottlenecking when checking if an item already holds a read lock.

alexdowad:
@ianks It might not be much of a big deal at all. @jdantonio, what is this read-write lock actually used for in Rails?

jdantonio:
@alexdowad I'm not entirely sure exactly how Rails is using the read/write lock. The code is [several turtles deep](https://en.wikipedia.org/wiki/Turtles_all_the_way_down) and I haven't had an opportunity to follow it all the way. Or to run the full test suite.

The only place in Rails I can find that references `ShareLock` is [ActiveSupport::Dependencies::Interlock](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/dependencies/interlock.rb). That, in turn, seems to be only referenced from within the [ActiveSupport::Dependencies](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/dependencies.rb) module. The `#load_interlock` function only seems to be called from within this module, and the `#run_interlock` module doesn't seem to be called from *anywhere* within Rails. So as far as I can tell the lock in only has that one use.

If I'm correct that this is the only place where `ShareLock` is being used, the best approach may be to add only the functionality needed by that module, rather than trying to have full feature parity. When I started working on this I assumed, perhaps incorrectly, that updating our lock would be fairly simple. If I was wrong then we should step back and reconsider.

As to which branch, would you mind working from this branch (`shared-lock`) and submitting your PR to it? Unless you feel that my updates on this branch are ill-advised, in which case please work from `master`.

jdantonio:
@ianks If the performance hit ends up being high then we'll look at other options. Possibly a subclass which supports re-entrant locks, lock upgrading, etc.. Or possibly constructor options. In the end, we *definitely* want to build more hooks from Rails!

jdantonio:
Because I've been busy all day I've haven't had a chance to do comprehensive testing, but it appears as though the code in its given state (understanding that there is a bug with the re-entrant write lock) may work for Rails.

I forked Rails and created a `shared-lock` branch over [here](https://github.com/jdantonio/rails/tree/shared-lock) and updated it to use the updated ReadWriteLock code on this branch. I've run the tests for ActiveSupport and I get 0 failures. I haven't had a chance to run the full test suite yet.

I'm going to fix the aforementioned bug and run all the tests as soon as I have time.

Interestingly, when I point the Gemfile to a the local path for concurrent-ruby I get one failing test (below), but that may be a false negative. It may have something to do with using a local path for the gem (the failing test is related to loading dependencies).

```
1) Failure:
LoadPathsTest#test_uniq_load_paths [/home/jerry/Projects/rails/activesupport/test/load_paths_test.rb:14]:
{"/home/jerry/Projects/ruby-concurrency/concurrent-ruby/lib"=>3, "/home/jerry/Projects/ruby-concurrency/concurrent-ruby/support"=>2}

3270 runs, 381929 assertions, 1 failures, 0 errors, 48 skips
```

alexdowad:
I just took at look at the use of Rails' `ShareLock`. You know how Rails auto-loads source files when you refer to an undefined constant? It's related to that. Basically, they don't want any requests to be running while source files are being (lazily) loaded. So auto-loading of source files is done with a write (exclusive) lock held, and some middleware is used to acquire the read (non-exclusive) lock before handling a request. The write lock is also acquired in a few places which mutate auto-loading-related state.

I can see why they made the RW-lock reentrant. It looks like trying to make it work with a non-reentrant lock would be quite hairy.

Upgrading/downgrading of locks isn't needed for this use case.

alexdowad:
I'm interested in using `ruby-concurrent`'s `ThreadLocalVar` in making `ReadWriteLock` reentrant. Any comments on how expensive (or not) it in terms of CPU/memory use?

jdantonio:
@chrisseaton Do you have any insight for @alexdowad, re: `ThreadLocalVar`?

chrisseaton:
I commented on his PR.

jdantonio:
Thanks for the feedback @chrisseaton!

jdantonio:
@alexdowad Thanks again for your help with this. Do you have all the information you need?

matthewd:
FYI https://github.com/rails/rails/pull/20928

alexdowad:
Now that we have fast `ThreadLocalVar`s, I'm going to look at adding a `ReentrantReadWriteLock` which uses them. I'd like to start from a fresh slate, though I may consult the code in this PR for ideas.

alexdowad:
OK, I have a reentrant RW-lock implementation. On my PC, an acquire+release of a non-reentrant RW-lock takes about 2 microsecs. Acquire+release of a reentrant RW-lock takes about 3 microsecs. I don't see any way to squeeze more perf out of it right now, aside from rewriting in C (which I can do if desired).

Is that difference significant to keep BOTH the non-reentrant and reentrant versions around? Or should we ditch the non-reentrant one, and just keep the reentrant one?

One good thing is that nested acquires/releases are MUCH faster than the outermost pair.

pitr-ch:
I think it's better to keep both reentrant and non-reentrant. Users may want to use non-reentrant to describe that it's intended behaviour. Is the non-reentrant lock failing on subsequent entering?

alexdowad:
If a thread tries to acquire the non-reentrant lock twice, it will deadlock.

jdantonio:
Superseded by #381.

PR #373, opened 2015-07-20
Fix some MRI 2.2.2 warnings

repinel:
```
lib/concurrent/atomic/atomic_fixnum.rb:84: warning: `-' after local variable or literal is interpreted as binary operator
lib/concurrent/atomic/atomic_fixnum.rb:84: warning: even though it seems like unary operator
lib/concurrent/atomic/read_write_lock.rb:195: warning: shadowing outer local variable - c'`
```


pitr-ch:
Thank you! Merging.

PR #375, opened 2015-07-20
Reduce pointer chasing/object allocation in code using WeakReference

alexdowad:
Please have a look and see if you think this is a good idea.

### DIFF
--- a/lib/concurrent/atomic/thread_local_var/weak_key_map.rb
+++ b/lib/concurrent/atomic/thread_local_var/weak_key_map.rb
@@ -28,20 +28,20 @@ module Concurrent
       require 'weakref'
 
       # @!visibility private 
-      class WeakReference
+      class WeakReference < ::WeakRef
 
         # The object id of the object being referenced.
         attr_reader :referenced_object_id
 
-        # This implementation of a weak reference simply wraps the standard WeakRef implementation
-        # that comes with the Ruby standard library.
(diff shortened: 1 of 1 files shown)


jdantonio:
@alexdowad This update is [broken on Rbx](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/71819709), but I'm quite sure why. I'll pull the code later tonight and take a look. I definitely want to merge this PR so I'll try to figure out what's going on in Rbx.

jdantonio:
@alexdowad The `WeakRef` class in Rbx doesn't seem to exhibit identical behavior to the MRI and JRuby class. Honestly, I've given up trying to figure out why it doesn't work. If you ate interested, the source is [here](https://github.com/rubinius/rubinius/blob/master/kernel/bootstrap/weakref.rb). I think the best approach is to detect the platform and use the original implementation on Rbx. This [gist](https://gist.github.com/jdantonio/5614d327665ca3dcfa3e) passes the tests on all platforms.

alexdowad:
OK, I found out why it does this. Have a look at https://github.com/rubinius/rubinius/blob/master/kernel/bootstrap/weakref.rb#L6.

Note that on RBX, `WeakRef.new` does *not* return an instance of `self`. It is hard-coded to return an instance of `WeakRef`, even if you call it on a subclass of `WeakRef`.

PR #376, opened 2015-07-20
Experimental thread-local variable implementation

alexdowad:
WARNING! NOT PRODUCTION-READY! DO NOT MERGE! This is just for others to look at and comment on.

The included benchmark creates 100 thread-local counters, spawns 100 threads, and makes each thread increment a randomly-chosen thread-local 10000 times. On this benchmark, I measure `AlexThreadLocalVar` as outperforming the existing `RubyThreadLocalVar` by a factor of more than 14.


lib/concurrent/atomic/thread_local_var.rb
```diff
@@ -115,6 +115,45 @@ def set(value)
...
+
+    def allocate_storage
+      @index = @@lock.synchronize do
+        @@free.pop || begin
+          result = @@next
+          @@next += 1
+          result
+        end
+      end
+      ObjectSpace.define_finalizer(self, RubyThreadLocalVar.finalizer(@index))
```
ianks: Isn't ObjectSpace disabled in JRuby by default?

jdantonio: @ianks It is, but JRuby provides its own `ThreadLocal` class, so the suggested `AlexThreadLocalVar` would replace only `RubyThreadLocalVar` and `JavaThreadLocalVar` would remain untouched. (@alexdowad Please correct me if I am wrong.)

jdantonio:
@alexdowad To be perfectly honest, this low-level stuff is outside my wheelhouse. @chrisseaton and @pitr-ch can provide you with much better feedback than I can. They are both very busy but hopefully we'll get feedback from one of them fairly soon.

lib/concurrent/atomic/thread_local_var.rb
chrisseaton: `ObjectSpace#each_object` with a non-`Module` subclass is disabled by default on JRuby, but other `ObjectSpace` methods, including `#define_finalizer` work fine.

chrisseaton:
To summarise the implementation, there is a single real thread local variable `:__tls_storage__`, which refers to a hash. The actual thread local variables are stored in that hash with keys that are `Fixnum`. In order to not run out of `Fixnum` values, they are 'garbage collected' by returning them to a freelist, which is synchronised.

I think it's possible that there is an ABA race condition on reading from `Thread.current`. Thread A could read it, find it empty, then thread B reads it, finds it empty and creates it with its value. Thread A then continues, and overwrites what thread B has already done.

```ruby
    def get
      (Thread.current[:__tls_storage__] ||= [])[@index]
    end

    def set(value)
      (Thread.current[:__tls_storage__] ||= [])[@index] = value
    end
```

That may therefore need to be synchronised - which I think is where your 14x is unfortunately going to  be lost.

Another problem is that who is clearing the `Thread.current[:__tls_storage__]` hash when thread local variables are no longer needed? Their index is returned to the free list, but when about the entry in the hash? Does that therefore need to be weak? Again that's going to seriously erode your 14x. Or does the finaliser also need to remove the entry from the hash.

alexdowad:
Thanks for looking at the code. To clarify a bit, `Thread.current` always returns the current thread. For information on `Thread#[]`, see the official docs: http://ruby-doc.org/core-2.2.2/Thread.html#method-i-5B-5D. (It accesses what is effectively a per-thread `Hash`. This is how Ruby provides "thread-local storage".)

An ABA race is not possible when reading `Thread.current[:key]`, because `Thread.current` is a completely different object in thread A and thread B.

Another clarification: in this code, `Thread.current[:__tls_storage__]` is an `Array`, not a `Hash`. You are right, the finalizer doesn't clear the entries which are used in these arrays. For this to be workable for production, it definitely needs to do so, not just so the objects stored in the arrays can be GC'd, but also so values from GC'd `ThreadLocalVar`s don't pass over when the indices are reused.

To do that, it will be necessary to either 1) track threads which are using `ThreadLocalVar`s, or 2) find some way to retrieve *all* threads. Or 3) references could be kept to the per-thread arrays, rather than the threads themselves. You are right, those references would have to be weak.

chrisseaton:
Ah right yeah I confused `Thread#[]` and `Thread.[]`.

If you can solve the other problems and still maintain performance this could work well - but I'd also be surprised if that's possible.

Why can't we just create a unique object for each thread local variable and use that as a key in `Thread.current`?

alexdowad:
Absolutely, we could create a unique object for each TLV, and use it as a key in `Thread.current[]`. The only thing is that we don't want to prevent the TLVs from being GC'd, so `Thread.current[]` would have to hold weak references. And when those TLVs are GC'd, we would also need to find all the threads which were holding the weak references, and delete them from the `Thread.current[]` hashes.

It would be very interesting to benchmark such an implementation against what I have here!

alexdowad:
OK, I think I found a way to overcome the problems mentioned in the above discussion, at a relatively small performance cost. At the same time, I found that the chaining from `#value=` to `#bind` to `#set` (and likewise on the read side) is quite expensive performance-wise. Flattening that call chain out buys back the lost performance and adds some more. Right now I am measuring a speed boost of about 15-20x.

Please have a look!

lib/concurrent/atomic/thread_local_var.rb
```diff
@@ -115,6 +115,104 @@ def set(value)
...
+
+    def self.threadlocal_finalizer(index)
+      lambda do
+        LOCK.synchronize do
+          FREE.push(index)
+          # The cost of GC'ing a TLV is linear in the number of threads using TLVs
+          # But that is natural! More threads means more storage is used per TLV
+          # So naturally more CPU time is required to free more storage
+          ARRAYS.each_key do |array|
+            array[index] = nil
```
thedarkone: You're changing a `Hash` key (`array`) here, its `hash` value might change, so after the `each_key` loop you need to "rehash the hash": `ARRAY.rehash`.

alexdowad: Aha, nice catch. So I really want something like ARRAYS[array.object_id] --
in other words, this was intended to be an identity set.

lib/concurrent/atomic/thread_local_var.rb
```diff
@@ -115,6 +115,104 @@ def set(value)
...
+    #   held in the now-unused slots (both to avoid blocking GC of those objects,
+    #   and also to prevent "stale" values from being passed on to a new TLV
+    #   when the index is reused)
+    # Because we need to null out freed slots, we need to keep references to
+    #   ALL the thread-local arrays -- ARRAYS is for that
+    # But when a Thread is GC'd, we need to drop the reference to its thread-local
+    #   array, so we don't leak memory
+
+    FREE = []
+    LOCK = Monitor.new
```
thedarkone: Looks like `LOCK` doesn't have to be re-entrant, lets switch it to `Mutex` for perf reasons?

alexdowad: Awesome, thanks for the suggestion! Coming right up!

alexdowad:
Despite the "experimental" name, this is now approaching something which could actually be used in production. I tested whether it leaks memory, and found a *devilishly* tricky bug which I couldn't figure out without using GDB to single-step through the Ruby interpreter code.

In short, finalizer procs which are passed to `Object.define_finalizer` are invoked with a single argument -- the object ID of the destroyed object. I was using *lambdas* rather than procs for the finalizers. Procs are "loose" in the arguments they accept, but lambdas are strict, and my 0-arg lambdas could not be invoked with 1 arg. Rather than crashing or throwing an exception or something, the Ruby interpreter just silently ignored the finalizers.

After changing `lambda` to `proc`, I was able to confirm that there are no memory leaks.

lib/concurrent/atomic/thread_local_var.rb
```diff
@@ -33,169 +33,131 @@ module Concurrent
...
+    #   ALL the thread-local arrays -- ARRAYS is for that
+    # But when a Thread is GC'd, we need to drop the reference to its thread-local
+    #   array, so we don't leak memory
 
     # @!visibility private
     NIL_SENTINEL = Object.new
-    private_constant :NIL_SENTINEL
+    FREE = []
+    LOCK = Mutex.new
+    ARRAYS = {} # used as a hash set
```
jdantonio: Do we want any of these to be private?

alexdowad: Up to you!

jdantonio: NIL_SENTINEL should be public for external comparison (:+1: to allowing nil to be a valid value in TLS). The others are mutable so keeping them public could lead to problems. Hopefully no one would try to use them directly, but making them private is safer.

This is a little pedantic, but in other places in this gem we use NO_VALUE in the same way that NIL_SENTINEL is used here. Probably best to use NO_VALUE here as well.

jdantonio:
Does the pure-Ruby version perform better than the JRuby version?

jdantonio:
This looks very good to me. Again, I'm not the expert on this but the design is very simple and the code is very concise. As far as I can tell, it does exactly what it needs to do and does it very optimally.

alexdowad:
On my dev PC, testing with JRuby 1.7.9 and JDK 1.8.0, yes, it does. Please try the benchmark yourself and see what results you get.

lib/concurrent/atomic/thread_local_var.rb
alexdowad: NO_VALUE (NIL_SENTINEL) should never be used externally to this class. Internally, it indicates "this variable has been explicitly set to `nil`". `nil` indicates "this variable has never been explicitly set yet, so its value is still the default". When you call `#value` or `#value=`, it converts accordingly, so only `nil` is seen by external callers.

I will change NIL_SENTINEL to NO_VALUE, and make the constants private.

lib/concurrent/atomic/thread_local_var.rb
```diff
@@ -33,169 +33,131 @@ module Concurrent
...
-
-      if value.nil?
-        @default
-      elsif value == NIL_SENTINEL
-        nil
+      if array = Thread.current[:__threadlocal_array__]
+        value = array[@index]
+        if value.nil?
+          @default
+        elsif value.equal?(NIL_SENTINEL)
```
jdantonio: Ah, I see it now. This is what I get for reviewing a PR while on a conference call. :-) I saw NIL_SENTINEL here and--in my mind--registered that it was the return value. We use NO_VALUE as a return value in other places in this gem. Sorry for my confusion.

alexdowad: So do you think NO_VALUE would be a more appropriate name? Or stick with
NIL_SENTINEL?

jdantonio: I'm fine either way. I just realized that `NIL_SENTINEL` is what the original implementation used. I mistakenly assumed that `NIL_SENTINEL` was just added. Again, sorry for the confusion.

thedarkone:
> In short, finalizer procs which are passed to `Object.define_finalizer` are invoked with a single argument -- the object ID of the destroyed object. I was using lambdas rather than procs for the finalizers. Procs are "loose" in the arguments they accept, but lambdas are strict, and my 0-arg lambdas could not be invoked with 1 arg. Rather than crashing or throwing an exception or something, the Ruby interpreter just silently ignored the finalizers.

Oh god!

jdantonio:
@alexdowad I'm comfortable merging this if you feel it's ready. Is it ready? (You still have the "WARNING! NOT PRODUCTION-READY! DO NOT MERGE!" message in the PR description, so I wanted to ask, just to be sure.)

alexdowad:
I think it's ready.

alexdowad:
THOUGH, if more people can benchmark under JRuby and ensure that I haven't
needlessly given up perf by removing the Java-specific implementation, that
would be better.

jdantonio:
I've done some testing on my MacBook Pro and I've had mixed results. The new version runs faster on JRuby 1.7.19 but slower on JRuby 9000 pre-release 1. I haven't had a chance to test it on the full release of JRuby 9000 yet because it isn't installable through RVM yet. I'll install it later and do more testing.

I'm going to merge the PR now because it will be easy to add the JRuby-specific implementation back later if necessary.

pitr-ch:
@alexdowad Very nice! Thanks.

schmurfy:
I admit I have not read everything but be careful with Thread#[], as stated in the documentation (and I was the one to add that patch) it will not access thread local variables but the current fiber's variables (or root fiber if none created) !

In most case as you don't use directly fiber it will not matter but you better use http://ruby-doc.org/core-2.2.2/Thread.html#method-i-thread_variable_get and http://ruby-doc.org/core-2.2.2/Thread.html#method-i-thread_variable_set to really use thread locals and avoid some weird side effects.

jdantonio:
@schmurfy Like this: 4851e5ea125f5901f77c9b20e572f1d5e561018b?

alexdowad:
@schmurfy and @jdantonio, thanks!

pitr-ch:
If we change this behaviour we should also probably add `FiberLocalVariable`.

alexdowad:
@pitr-ch, the previous `ThreadLocalVar` was indeed thread-local and not fiber-local, as it uses a map keyed off `Thread.current`. If someone has a use case for `FiberLocalVar`, then certainly we could add it, but it's not clear whether that is so.

jdantonio:
Do we really need FiberLocalVar? One of the main goals of this library is to prevent Ruby devs from every having to work with threads or fibers directly. ThreadLocalVariable is mainly an internal tool. I appreciate the symmetry, but I don't know if there is a strong use case. And, if memory serves, JRuby and Rbx just map Fiber to Thread, don't they?

pitr-ch:
@alexdowad Ah you are right, so we should just make a issue for future addition.

@jdantonio The use case is not a strong one, I have to admit. I can image libs which would wish to abstract fiber context information to a FiberLocalVar as other do with ThreadLocalVar. However Fibers are not performant on JRuby as you've also said so they are not really that much interesting for concurrent-ruby. TL;DR I think it's just nice to have, you can ignore my suggestion.

jdantonio:
@pitr-ch I'll never ignore your suggestions. Your insight and help are invaluable! :-)

schmurfy:
@jdantonio yes like that, thanks ^^
it is unfortunate to see Thread#[] not accessing thread local variables but since they did a, in my opition, a stupid choice when MRI 1.9 was released they are now stuck with it... At least now we have real thread local variables.

I currently use fibers and I think they are a nice tool to have in your toolbox but as far as concurrent-ruby is concerned I think they can safely be ignored for now, celluloid-io use them for... IO while keeping the number of threads low.
The way I currently use them is with eventmachine to prevent a callback hell and keep our asynchronous code "linear".

jdantonio:
@schmurfy I used EventMachine heavily at my last job. That's when I grew to really dislike Fiber. In fact, my frustration that the two most popular concurrency choices seemed to be EventMachine and Celluloid was the impetus for creating this gem.

schmurfy:
I remember you don't like them and your are not the only one ^^
As for me I have no issues with them and they help a lot keeping our code readable, callback calling callbacks is just an horrible way of doing things.

But the reason I am keeping a close look on concurrent-ruby is to switch to threads when I get the chance now that multi core processors are the norm, when we started our project (ruby) threads meant more problems than anything else, knowing when a switch can occurs with fibers allow to nullify some concurrency problems you have with threads.

PS: I tried Celluloid and I am not a big fan of it either, they elevated black magic to an art... There are some old discussions about that from me on this issue tracker and on the mailing list if I remember well ;)

pitr-ch:
Since we do not have any IO integration so far, I see an advantage in EventMachine, that it can be easily used to do IO ops efficiently in one thread (in its loop) and easily delegating the data to any ConcurrentRuby abstraction our user may choose based on a given problem. I think Fibres are good idea but since there is no supported on JVM, they are not usable in ConcurrentRuby.

PR #377, opened 2015-07-20
Add requires to support standalone `atom` require

philipmw:
The readme says that we should be able to `require` individual components of Concurrent-Ruby gem, like so:

    require 'concurrent/atom'

However, this particular `require` fails because of un-modeled dependencies internal to the gem.

This PR adds the missing dependencies. Now `require 'concurrent/atom'` succeeds.


jdantonio:
@philipmw Nice catch! Thank you very much. We plan to release an update this week so your fix should be live very soon.

PR #380, opened 2015-07-24
updated patch and jruby

ferdinandrosario:
Since jruby 9k released adding that into the travis matrix and update ruby-2.2 patch version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
 language: ruby
 
 rvm:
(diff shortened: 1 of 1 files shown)


ferdinandrosario:
If possible we can remove 2.1.2 & 2.1.5 and use 2.1.6 to help travis by avoid more ruby version.

jdantonio:
@ferdinandrosario Thank you!

PR #381, opened 2015-07-26
New ReentrantReadWriteLock class

alexdowad:
See PR #372 for the background on this one.

I am opening this PR to allow for code review. It is NOT ready to merge yet.

### DIFF
--- a/lib/concurrent/atomic/reentrant_read_write_lock.rb
+++ b/lib/concurrent/atomic/reentrant_read_write_lock.rb
@@ -1,0 +1,375 @@
+require 'thread'
+require 'concurrent/atomic/atomic_reference'
+require 'concurrent/errors'
+require 'concurrent/synchronization'
+require 'concurrent/atomic/thread_local_var'
+
+module Concurrent
+
+  # Re-entrant read-write lock implementation
+  #
+  # Allows any number of concurrent readers, but only one concurrent writer
+  # (And while the "write" lock is taken, no read locks can be obtained either.
+  # Hence, the write lock can also be called an "exclusive" lock.)
+  #
+  # If another thread has taken a read lock, any thread which wants a write lock
+  # will block until all the readers release their locks. However, once a thread
+  # starts waiting to obtain a write lock, any additional readers that come along
+  # will also wait (so writers are not starved).
+  #
+  # A thread can acquire both a read and write lock at the same time. A thread can
+  # also acquire a read lock OR a write lock more than once. Only when the read (or
+  # write) lock is released as many times as it was acquired, will the thread
+  # actually let it go, allowing other threads which might have been waiting
+  # to proceed.
+  #
+  # If both read and write locks are acquired by the same thread, it is not strictly
+  # necessary to release them in the same order they were acquired. In other words,
+  # the following code is legal:
+  #
+  # @example
+  #   lock = Concurrent::ReentrantReadWriteLock.new
+  #   lock.acquire_write_lock
+  #   lock.acquire_read_lock
+  #   lock.release_write_lock
+  #   # At this point, the current thread is holding only a read lock, not a write
+  #   # lock. So other threads can take read locks, but not a write lock.
+  #   lock.release_read_lock
+  #   # Now the current thread is not holding either a read or write lock, so
+  #   # another thread could potentially acquire a write lock.
+  #
+  # This implementation was inspired by `java.util.concurrent.ReentrantReadWriteLock`.
+  #
+  # @example
+  #   lock = Concurrent::ReentrantReadWriteLock.new
+  #   lock.with_read_lock  { data.retrieve }
+  #   lock.with_write_lock { data.modify! }
+  #
+  # @see http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html java.util.concurrent.ReentrantReadWriteLock
+  class ReentrantReadWriteLock < Synchronization::Object
+
+    # Implementation notes:
+    #
+    # A goal is to make the uncontended path for both readers/writers mutex-free
+    # Only if there is reader-writer or writer-writer contention, should mutexes be used
+    # Otherwise, a single CAS operation is all we need to acquire/release a lock
+    #
+    # Internal state is represented by a single integer ("counter"), and updated
+    #   using atomic compare-and-swap operations
+    # When the counter is 0, the lock is free
+    # Each thread which has one OR MORE read locks increments the counter by 1
+    #   (and decrements by 1 when releasing the read lock)
+    # The counter is increased by (1 << 15) for each writer waiting to acquire the
+    #   write lock, and by (1 << 29) if the write lock is taken
+    #
+    # Additionally, each thread uses a thread-local variable to count how many times
+    #   it has acquired a read lock, AND how many times it has acquired a write lock.
+    # It uses a similar trick; an increment of 1 means a read lock was taken, and
+    #   an increment of (1 << 15) means a write lock was taken
+    # This is what makes re-entrancy possible
+    #
+    # 2 rules are followed to ensure good liveness properties:
+    # 1) Once a writer has queued up and is waiting for a write lock, no other thread
+    #    can take a lock without waiting
+    # 2) When a write lock is released, readers are given the "first chance" to wake
+    #    up and acquire a read lock
+    # Following these rules means readers and writers tend to "take turns", so neither
+    #   can starve the other, even under heavy contention
+
+    # @!visibility private
+    READER_BITS    = 15
+    # @!visibility private
+    WRITER_BITS    = 14
+
+    # Used with @Counter:
+    # @!visibility private
+    WAITING_WRITER = 1 << READER_BITS
+    # @!visibility private
+    RUNNING_WRITER = 1 << (READER_BITS + WRITER_BITS)
+    # @!visibility private
+    MAX_READERS    = WAITING_WRITER - 1
+    # @!visibility private
+    MAX_WRITERS    = RUNNING_WRITER - MAX_READERS - 1
+
+    # Used with @HeldCount:
+    # @!visibility private
+    WRITE_LOCK_HELD = 1 << READER_BITS
+    # @!visibility private
+    READ_LOCK_MASK  = WRITE_LOCK_HELD - 1
+    # @!visibility private
+    WRITE_LOCK_MASK = MAX_WRITERS
+
+    # Create a new `ReentrantReadWriteLock` in the unlocked state.
+    def initialize
+      @Counter    = AtomicFixnum.new(0)       # single integer which represents lock state
+      @ReadQueue  = Synchronization::Lock.new # used to queue waiting readers
+      @WriteQueue = Synchronization::Lock.new # used to queue waiting writers
+      @HeldCount  = ThreadLocalVar.new(0)     # indicates # of R & W locks held by this thread
+      ensure_ivar_visibility!
+    end
+
+    # Execute a block operation within a read lock.
+    #
+    # @yield the task to be performed within the lock.
+    #
+    # @return [Object] the result of the block operation.
+    #
+    # @raise [ArgumentError] when no block is given.
+    # @raise [Concurrent::ResourceLimitError] if the maximum number of readers
+    #   is exceeded.
+    def with_read_lock
+      raise ArgumentError.new('no block given') unless block_given?
+      acquire_read_lock
+      begin
+        yield
+      ensure
+        release_read_lock
+      end
+    end
+
+    # Execute a block operation within a write lock.
+    #
+    # @yield the task to be performed within the lock.
+    #
+    # @return [Object] the result of the block operation.
+    #
+    # @raise [ArgumentError] when no block is given.
+    # @raise [Concurrent::ResourceLimitError] if the maximum number of readers
+    #   is exceeded.
+    def with_write_lock
+      raise ArgumentError.new('no block given') unless block_given?
+      acquire_write_lock
+      begin
+        yield
+      ensure
+        release_write_lock
+      end
+    end
+
+    # Acquire a read lock. If a write lock is held by another thread, will block
+    # until it is released.
+    #
+    # @return [Boolean] true if the lock is successfully acquired
+    #
+    # @raise [Concurrent::ResourceLimitError] if the maximum number of readers
+    #   is exceeded.
+    def acquire_read_lock
+      if (held = @HeldCount.value) > 0
+        # If we already have a lock, there's no need to wait
+        if held & READ_LOCK_MASK == 0
+          # But we do need to update the counter, if we were holding a write
+          #   lock but not a read lock
+          @Counter.update { |c| c + 1 }
+        end
+        @HeldCount.value = held + 1
+        return true
+      end
+
+      while true
+        c = @Counter.value
+        raise ResourceLimitError.new('Too many reader threads') if max_readers?(c)
+
+        # If a writer is waiting OR running when we first queue up, we need to wait
+        if waiting_or_running_writer?(c)
+          # Before going to sleep, check again with the ReadQueue mutex held
+          @ReadQueue.synchronize do
+            @ReadQueue.ns_wait if waiting_or_running_writer?
+          end
+          # Note: the above 'synchronize' block could have used #wait_until,
+          #   but that waits repeatedly in a loop, checking the wait condition
+          #   each time it wakes up (to protect against spurious wakeups)
+          # But we are already in a loop, which is only broken when we successfully
+          #   acquire the lock! So we don't care about spurious wakeups, and would
+          #   rather not pay the extra overhead of using #wait_until
+
+          # After a reader has waited once, they are allowed to "barge" ahead of waiting writers
+          # But if a writer is *running*, the reader still needs to wait (naturally)
+          while true
+            c = @Counter.value
+            if running_writer?(c)
+              @ReadQueue.synchronize do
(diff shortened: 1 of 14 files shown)


alexdowad:
Note the new `#try_write_lock` and `#try_read_lock` methods. Also, this class allows "upgrading" and "downgrading" of locks -- to do this, simply acquire a read lock before releasing your write lock (or vice versa). It doesn't care if you don't follow a strict ordering of write-acquire, read-acquire, read-release, write-release.

alexdowad:
Specs are coming soon...

alexdowad:
I cherry-picked the 2 commits from @jdantonio's PR which still seem useful, so if/when this PR is ready to merge, those will go in at the same time.

lib/concurrent/atomic/reentrant_read_write_lock.rb
```diff
@@ -0,0 +1,353 @@
...
+        return true
+      end
+
+      while true
+        c = @Counter.value
+        raise ResourceLimitError.new('Too many writer threads') if max_writers?(c)
+
+        if c == 0 # no readers OR writers running OR writers waiting
+          # if we successfully swap the RUNNING_WRITER bit on, then we can go ahead
+          if @Counter.compare_and_set(0, RUNNING_WRITER)
```
jdantonio: @alexdowad If I follow the code correctly, this is the bit that allows a thread with a read lock to "upgrade" to a write lock (as in the Rails lock class). Am I correct?

jdantonio:
@alexdowad This looks fantastic! Thank you very much for working on this!

The tests are failing on all JRuby build because of the Java update to `AtomicFixnum`:

```
ext/com/concurrent_ruby/ext/JavaAtomicFixnumLibrary.java:80: error: cannot find symbol
        public IRubyObject update(ThreadContext context, Block block) {
                                                         ^
  symbol:   class Block
  location: class JavaAtomicFixnum
```

This looks like it should be easy to fix. I can look into it if need be.

The tests are also failing on Windows because of a `Bignum`/`Fixnum` issue. I will look into that for you.

With respect to the tests, it seems to me that we should be able to extract most of the tests for `ReadWriteLock` into a set of shared specs, `behaves_as :read_write_lock`. I can do that myself, so if you only write tests for the new re-entrant functionality that should be plenty.

lib/concurrent/atomic/reentrant_read_write_lock.rb
alexdowad: @jdantonio, wow, you just made me realize that this code does not handle that "read lock upgrading to write lock" scenario properly! Thanks, I'll fix it!

alexdowad:
I'll fix the Java code. Sorry, this PR is still a bit rough. Regarding `Bignum`/`Fixnum`, would that be on a 32-bit machine by any chance? It's been a while since I looked at the relevant MRI code, but I suspect that they may be using 2 bits for tagging, leaving only 30 bits available for fixnums. Could you adjust `WRITER_BITS` to 14 rather than 15, and re-run the test on Windows?

jdantonio:
The Windows build will run automatically every time you update the PR. Right now we're testing both 32 and 64 bit versions of Ruby on Windows. There is a `Concurrent.on_windows?` helper function that might be helpful. Perhaps something like this?

```
WRITER_BITS = (Concurrent.on_windows? ? 14 : 15)
```

alexdowad:
You can see that the specs are underway now (and have already flushed out a number of bugs). Still lots more tests to add.

A fun diversion has come up, though: I wrote a spec which causes MRI to segfault! To my trusty debugger...

alexdowad:
Sorry, just re-ordered some commits.

alexdowad:
We have a pretty good little suite of specs for `ReentrantReadWriteLock` spec now. More bugs flushed out.

I think a thorough code review by at least one either developer is needed before this goes into production.

jdantonio:
The tests hang with `Concurrent::ReentrantReadWriteLock #try_read_lock` on all versions of Windows (see [here](https://ci.appveyor.com/project/rubyconcurrency/concurrent-ruby/build/79-master/job/c0knlp6d3dunbaat)). I'll pull the code to my Windows test machine later tonight and see if I can figure out what's going on.

alexdowad:
OK, I'm on it.

alexdowad:
If anyone is interested in the MRI segfault which was discovered while working on this PR, here is what it turned out to be: https://bugs.ruby-lang.org/issues/11404

alexdowad:
The problem on Windows is sorted out. Unfortunately, it looks like `Rubinius.unlock` is buggy. Have a look at this:

```ruby
        def ns_wait(timeout = nil)
          wchan = Rubinius::Channel.new

          begin
            @__Waiters__.push wchan
            puts "Waiting.... #{self.inspect}"
            puts "Locked = #{Rubinius.locked?(self)}"
            puts Rubinius.unlock(self).inspect
            puts "Locked = #{Rubinius.locked?(self)}"
            signaled = wchan.receive_timeout timeout
          ensure
            Rubinius.lock(self)

            if !signaled && !@__Waiters__.delete(wchan)
              # we timed out, but got signaled afterwards,
              # so pass that signal on to the next waiter
              @__Waiters__.shift << true unless @__Waiters__.empty?
            end
          end

          self
        end
```

That's a modified version of `Concurrent::Synchronization::Lock#ns_wait`, with some debug print statements. Output from RSpec test:

```
Waiting.... #<Concurrent::Synchronization::Lock:0x2260 @__Waiters__=[#<Rubinius::Channel
:0x2268>]>
Locked = true
nil
Locked = true
```

Looking at the RBX source, it looks like `Rubinius.unlock` is supposed to return `nil` if successful, or some special "primitive failure" value if unsuccessful. You can see that it is returning `nil`, but the object is still not unlocked!

jdantonio:
I'm glad you were able to get things to work on Windows! I didn't have a chance to take a look last night as I had hoped. The one test that failed on one Windows build doesn't concern me. Tests that check the status of a thread are incredibly difficult to make work consistently.

The Rbx bug is unfortunate, but I don't think this should stop is from merging this PR. There is nothing we can do about a buggy runtime. We can submit a bug to Rbx and add an Rbx-specific warning to the code.

alexdowad:
@jdantonio, of course you can do whatever you like, but I would suggest holding onto this one for just a few days to see if we can really make it right.

jdantonio:
@alexdowad I was under the impression that we couldn't do anything about this. If there is a way to fix it then by all means, we should.

pitr-ch:
@alexdowad thanks for letting me know, I'll look asap. :+1: for waiting until fixed, we cannot released it if it's broken on Rubinius anyway, we should at least give them a chance to fix it if it's indeed bug in the runtime.

lib/concurrent/atomic/reentrant_read_write_lock.rb
```diff
@@ -0,0 +1,374 @@
...
+    # @!visibility private
+    WRITE_LOCK_MASK = MAX_WRITERS
+
+    # Create a new `ReentrantReadWriteLock` in the unlocked state.
+    def initialize
+      @Counter    = AtomicFixnum.new(0)       # single integer which represents lock state
+      @ReadQueue  = Synchronization::Lock.new # used to queue waiting readers
+      @WriteQueue = Synchronization::Lock.new # used to queue waiting writers
+      @HeldCount  = ThreadLocalVar.new(0)     # indicates # of R & W locks held by this thread
+    end
```
pitr-ch: it's more safe to have `ensure_ivar_visibility!` here as in `ReadWriteLock`.

alexdowad: The question is, if we put a memory barrier there on Rubinius, are we also going to put a memory barrier at the beginning of every method which uses the ivars? If not, then is the memory barrier in the constructor actually worth anything?

Regarding the current problem with `Rubinius.unlock` -- I have tried peppering the code with memory barriers and it didn't make any difference.

alexdowad:
OK, interesting. I just checked out what is happening inside RBX. Their object locks are reentrant. In the above code sample, when `Rubinius.unlock` is called, `self` has already been locked *twice*, not just once. That is why it is still locked after calling `Rubinius.unlock(self)`.

alexdowad:
And the reason **why**? I'm calling `@WriteQueue.wait` from *inside* a `synchronized` block. `Synchronization::Lock#wait` contains *another* `synchronized` block.

So it turns out that `Rubinius.unlock` is working just fine. The bug is in `Synchronization::Lock`. It doesn't behave consistently across Ruby implementations. Script for repro:

```ruby
$:.unshift(Dir.pwd+'/lib')
require 'concurrent'
include Concurrent

lock = Synchronization::Lock.new
Thread.new { sleep(0.1); lock.signal }
lock.synchronize { lock.wait }
puts "Finished!"
```

Run that under MRI, it finishes. Run it under RBX, it hangs.

alexdowad:
I have worked around the problem with `#wait` by just calling `#ns_wait` directly. Also discovered that RSpec `let` blocks are not protected against simultaneous access from multiple threads... I may send a PR to RSpec to try fixing that. In the meantime, I have worked around it.

lib/concurrent/atomic/reentrant_read_write_lock.rb
pitr-ch: The reasoning for the barrier is as follows: the object is being constructed so no other thread has a reference to it (assuming proper construction, where reference does not escape during construction). The barrier is here so final field assignments of the constructed object cannot be reordered with an assignment of the constructed object to a publicly accessible variable where it can be read by other threads. If the barrier is omitted other threads can see the object instance before it's fully initialised because of the possible reorderings on thread performing the construction.

alexdowad: OK, makes sense!

I'd like to make this improvement *without* inheriting from `Synchronization::Lock`. `Lock` adds extra instance variables which will not be useful for `ReentrantReadWriteLock`. This would increase the memory footprint of each instance.

What would you think of `Rubinius.memory_barrier if Concurrent.on_rbx?`?

pitr-ch:
@alexdowad Calling ns_wait is actually the intended use to avoid double locking. Never the less, what you have described is a problem on rbx. I think I have a fix I'll push it soon.

alexdowad:
@pitr-ch, I also have a fix and can send a PR.

pitr-ch:
@alexdowad :)
```ruby
        def synchronize(&block)
          if Rubinius.locked?(self)
            block.call
          else
            Rubinius.synchronize(self, &block)
          end
        end
```
or something else ?

pitr-ch:
I did not tested it yet, if you did please send a PR.

alexdowad:
...except I used `yield`. :)

pitr-ch:
I think it's fine if you put it here in this PR, that's the easiest.

alexdowad:
I'm getting a really fascinating failure with the above code. The `'allows only one thread to execute count'` spec for `#synchronize` is failing. Putting `Rubinius.memory_barrier` before and after `yield` in `#synchronize` does **not** make the spec pass. But this does:

```ruby
        def count
          synchronize { @count += 1; Rubinius.memory_barrier; @count }
        end
```

How crazy is that??

pitr-ch:
Quite a lot :) Ah I do not have rbx on this computer, installing to take a look.

alexdowad:
You can pull the "testing_rbx" branch from github.com/alexdowad/concurrent-ruby to try the exact same code that I am trying.

alexdowad:
OK, I think this might be a simple mistake. `Rubinius.locked?(self)` checks whether `self` is **locked by any thread**, not whether it's locked by **this** thread.

alexdowad:
@pitr-ch, I'm not sure if you'll like the way I've fixed `RbxObject#synchronize`, please have a look. It would be nice if RBX had a primitive to check if an object was locked *by the current thread*, but it doesn't, as far as I can see.

lib/concurrent/atomic/reentrant_read_write_lock.rb
pitr-ch: For now i think its better to inherit so it's used standardly. However I am aware of this issue and I'll fix it later for all classes using `Synchronisation::Object`. If here is some custom code here I can easily miss it.

pitr-ch:
The approach is fine, Thanks! Could you wrap `unlock` and `owner = nil` in finalise section?

pitr-ch:
Hm one more thought: isn't testing `locked?` unnecessary if we check `__owner__` anyway?

pitr-ch:
Could you create AbstractReentrantLock, it looks like there are some duplications between the two locks.

alexdowad:
@pitr-ch, what parts would you suggest pulling out into `AbstractReentrantReadWriteLock`?

pitr-ch:
Ah sorry I meant `AbstractReadWriteLock`. `atomic/read_write_lock.rb` and `atomic/reentrant_read_write_lock.rb` seem almost the same. Could you make it more DRY? Either by creating common Abstract parent or by making reentrant a child of the other.

lib/concurrent/synchronization/rbx_object.rb
```diff
@@ -8,13 +8,26 @@ module Synchronization
       class RbxObject < AbstractObject
         def initialize
           @__Waiters__ = []
+          @__Owner__   = nil
```
pitr-ch: sorry for nitpick: this should be `@__owner__` since it's mutated.

alexdowad: Thanks, I didn't know your convention was caps for "final". Fixed.

alexdowad:
@pitr-ch, they do seem similar, because the algorithm used for `ReentrantReadWriteLock` was derived from the one used for `ReadWriteLock`, but there are numerous differences in the details of the algorithm. It might be possible to find some little random bits of code which are the same, but not conceptually "cohesive" operations which make sense on their own.

DRY is generally a good thing, but can be detrimental when taken too far. Why? Because removing duplication often means adding another layer of indirection. Too much indirection makes code harder to read, understand, and maintain.

At the same time, higher LOC count also makes a project harder to understand and maintain. And unifying duplicated code has the advantage that bug fixes and performance optimizations on de-duplicated code can benefit all the places that use it. On the other hand, de-duplicating also means that you can *break* class A while trying to fix class B. You are creating dependencies where none existed before.

There is more that could be said about pros and cons on both sides, but in short, I believe that readability, maintainability, and performance of this code will all be enhanced if we just keep the reentrant RW-lock classes separate.

However, I would love to be proved wrong, if you (or anybody) can show a nice way to break out the pieces of the algorithms, which reduces duplication while enhancing readability and maintaining good performance.

pitr-ch:
Ok at first glance it looked similar a lot, but as you say it just appears that way. So lets keep it, no need to delay this PR. I may look at it later.

jdantonio:
Unfortunately, while @alexdowad was doing his excellent work to create a highly-performant lock class that was compatible with the dependency loader lock in Rails, they [added new features](https://github.com/rails/rails/commit/bd31aec9c33453e0ba96aec614e56958784e6b8d) to their lock. At this point it looks like their lock is highly specialized and probably not worth our time to support. :-(

alexdowad:
It is quite specialized indeed. I wonder how much they gain from those "loose upgrades". It seems to me the purpose could probably be achieved in a much simpler way, though I'm not deep in the Rails code so I don't know.

matthewd:
@alexdowad they prevent us from deadlocking as soon as you have two concurrent requests, which both need an opportunity to do a load

jdantonio:
@alexdowad Did you get my email?

jdantonio:
On the bright side, this new lock class is an excellent addition! And we have an improved TLV, too!

alexdowad:
@matthewd, why do requests need to do a load? Autoload of Rails classes/other gem classes?

matthewd:
Rails application classes (e.g., the UserController or Post classes) -- the lock is for the development mode, (re)load-on-demand, behaviour.

alexdowad:
Why does that reloading have to be done in the middle of a request? Can't you complete all outstanding requests, reload, then start accepting requests again?

matthewd:
We unload between requests, and we *load* each class when it is referenced (via `const_missing`); we can't pre-load the entire application every time because performance.

alexdowad:
OK, fair enough.

jdantonio:
@matthewd Thank you for all your feedback!

pitr-ch:
We could generalise our ReadWrite lock implementation to be able to cover the Rails use-case. A factory class could be introduced to CR, which would generate a Lock classes based on given rules. The rules would be a small set of purposes and compatibility rules. Based on that it should be able to generate performant implementation. For ReadWriteLock the rules would be: `{read: [:read], write: []}` (where key is name of purpose and value is set of compatible purposes). For the Rails use-case it would be: `{load: [:load], unload: [:load, :unload]}`. The factory class would divide the bits in Fixnum to track the counts for given purposes, failing when there is not enough space or fall-backing to an immutable struct of Fixnums.

This is just an idea for improvement, since the lock is not used in Rails production, there will be little gain if Rails migrate to this more performant implementation.

alexdowad:
@pitr-ch It's an interesting idea, but do you think such an interface would be useful for anyone else aside from Rails?

alexdowad:
Another thing: we can cover the Rails use case much more simply:

```ruby
class SharedLock < Concurrent::ReentrantReadWriteLock
  def with_write_lock_loose(&block)
    if (@HeldCount.value & READ_LOCK_MASK) > 0
      # we are already holding a read lock
      release_read_lock
      with_write_lock(&block)
      acquire_read_lock
    else
      with_write_lock(&block
    end
  end
end
```

pitr-ch:
Yeah that sounds more practical. I cannot think of another example. I mainly wanted to put the idea somewhere if anybody is interested looking into it.

PR #382, opened 2015-07-27
Fix: Concurrent.zip value! logs errors and raise Concurrent::Error

dueckes:
Corrects a minor typo - when futures are zipped, any errors that occur were not being logged correctly.


pitr-ch:
Good catch, thanks!

PR #385, opened 2015-07-29
Cleared all interpreter warnings.

jdantonio:
* Added parenthesis when using splat operator
* Used _ for unused variables
* Renamed variables when block echoed surrounding scope
* Configuration now autoloads lazily.
* Cleared all circular references.

### DIFF
--- a/lib/concurrent/configuration.rb
+++ b/lib/concurrent/configuration.rb
@@ -1,9 +1,13 @@
 require 'thread'
-require 'concurrent/atomics'
+require 'concurrent/delay'
 require 'concurrent/errors'
-require 'concurrent/executors'
+require 'concurrent/atomic/atomic_reference'
 require 'concurrent/concern/deprecation'
 require 'concurrent/concern/logging'
+require 'concurrent/executor/timer_set'
+require 'concurrent/executor/immediate_executor'
+require 'concurrent/executor/fixed_thread_pool'
+require 'concurrent/executor/thread_pool_executor'
 require 'concurrent/utility/at_exit'
 require 'concurrent/utility/processor_counter'
 
@@ -32,8 +36,8 @@ module Concurrent
              formatted_message
     end
 
-    lambda do |level, progname, message = nil, &block|
-      logger.add level, message, progname, &block
+    lambda do |loglevel, progname, message = nil, &block|
+      logger.add loglevel, message, progname, &block
     end
   end
 
--- a/lib/concurrent/edge/future.rb
+++ b/lib/concurrent/edge/future.rb
@@ -113,7 +113,7 @@ module Concurrent
       # post job on executor
       # @return [true, false]
       def post_on(executor, *args, &job)
-        Concurrent.executor(executor).post *args, &job
+        Concurrent.executor(executor).post(*args, &job)
       end
 
       # TODO add first(futures, count=count)
@@ -307,7 +307,7 @@ module Concurrent
       # @!visibility private
       def complete_with(state, raise_on_reassign = true)
         if @State.compare_and_set(PENDING, state)
-          (state)
+          #(state)
           # go to synchronized block only if there were waiting threads
           synchronize { ns_broadcast } if @Waiters.clear
           call_callbacks
@@ -323,7 +323,7 @@ module Concurrent
       # @return [Array<AbstractPromise>]
       def blocks
         @Callbacks.each_with_object([]) do |callback, promises|
-          promises.push *callback.select { |v| v.is_a? AbstractPromise }
+          promises.push(*(callback.select { |v| v.is_a? AbstractPromise }))
         end
       end
 
@@ -464,7 +464,7 @@ module Concurrent
       # @!visibility private
       class SuccessArray < Success
         def apply(block)
-          block.call *value
+          block.call(*value)
         end
       end
 
@@ -520,7 +520,7 @@ module Concurrent
         end
 
         def apply(block)
-          block.call *reason
+          block.call(*reason)
         end
       end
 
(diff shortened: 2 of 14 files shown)


lib/concurrent/configuration.rb
```diff
@@ -1,16 +1,21 @@
...
 module Concurrent
   extend Concern::Logging
   extend Concern::Deprecation
 
+  autoload :AtExit,             'concurrent/utility/at_exit'
+  autoload :AtomicReference,    'concurrent/atomic/atomic_reference'
+  autoload :Delay,              'concurrent/delay'
+  autoload :ProcessorCounter,   'concurrent/utility/processor_counter'
+  autoload :TimerSet,           'concurrent/executor/timer_set'
+  autoload :ImmediateExecutor,  'concurrent/executor/immediate_executor'
```
ianks: Why the switch to autoload? I remember a post from Matz about it having thread safety issues. Is that still true?

jdantonio: That's the question at hand. None of the changes in this PR are necessary, just a good opportunity for discussion.

The issue is that we create a lot of circular references to `concurrent/configuration`. So far this hasn't been a problem but it's a source of concern. During my #386 experimentation I noticed that `thread_safe` uses `autoload` so I thought I'd give it a try. All the tests pass and the interpreter warnings have gone away.

lib/concurrent/edge/future.rb
```diff
@@ -775,14 +775,14 @@ def call_callback(method, state, *args)
         self.send method, state, *args
       end
 
-      def pr_async_callback_on_success(state, executor, callback)
-        pr_with_async(executor, state, callback) do |state, callback|
+      def pr_async_callback_on_success(_state, _executor, _callback)
+        pr_with_async(_executor, _state, _callback) do |state, callback|
           pr_callback_on_success state, callback
```
pitr-ch: I think these names suggests that the `_state, _executor, _callback` is unused, could you do `pr_async_callback_on_success(state, executor, callback)` and `pr_with_async(executor, state, callback) { |st, cb| pr_callback_on_success st, cb }` instead?

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -296,8 +296,8 @@ def kill
 
       private
 
-      def create_worker(queue, pool, idletime)
-        Thread.new(queue, pool, idletime) do |queue, pool, idletime|
+      def create_worker(_queue, _pool, _idletime)
+        Thread.new(_queue, _pool, _idletime) do |queue, pool, idletime|
```
pitr-ch: Not sure what to do here, but the method signature is more important I think, so it should keep the names without `_`.

lib/concurrent/configuration.rb
pitr-ch: I would rather shuffle the file content to avoid the circularity if possible than using autoload. Not sure if it can be easily done though :/

lib/concurrent/edge/future.rb
jdantonio: Done.

lib/concurrent/executor/ruby_thread_pool_executor.rb
jdantonio: Done.

lib/concurrent/configuration.rb
jdantonio: According to [this](https://bugs.ruby-lang.org/issues/921) it *seems* that `autoload` is now thread-safe. I've reached out to @headius and @tenderlove over Twitter for clarification.

jdantonio: Here's the response: "autoload itself is safe but class definition happening in parallel can still expose incomplete classes."

https://twitter.com/headius/status/626797364066578432

lib/concurrent/edge/future.rb
```diff
@@ -803,8 +803,8 @@ def pr_callback_notify_blocked(state, promise)
         super(promise)
       end
 
-      def pr_async_callback_on_completion(state, executor, callback)
-        pr_with_async(executor, state, callback) do |state, callback|
+      def pr_async_callback_on_completion(_state, _executor, _callback)
+        pr_with_async(_executor, _state, _callback) do |state, callback|
```
pitr-ch: same here pls

lib/concurrent/configuration.rb
jdantonio: The big problem is `Delay`. We use Delay extensively to lazy-load resources--especially in `configuration.rb`. Delay itself is relatively uncoupled from everything else. The rabbit-hole is `Executor.executor_from_options(opts)`. *Many* of our abstractions, including Delay, use that function to initialize an executor. That function, in turn, uses the global executor declarations from configuration.rb, most of which are lazy-loaded with Delay.

Since Delay doesn't need configuration data until an object is constructed we could potentially lazy-load configuration w/i Delay and possibly avoid the circular dependency.

lib/concurrent/edge/future.rb
jdantonio: Done.

lib/concurrent/configuration.rb
pitr-ch: We could also clear this up by creating a boot (or core) file which will always load the minimal required files for the concurrent-ruby to function it would load these problematic files. All the other abstractions would load boot first and then any other required abstractions. This way the circular dep. would be always handled by boot.

jdantonio:
This update gets rid of the autoload from configuration. The only remaining warnings are the circular references. Since there have not been a problem thus far I don't think it's a critical thing to change right now.

I like the idea from @pitr-ch of having a "boot" file. That's something that can probably wait until the 2.0 release.

PR #386, opened 2015-07-30
Import `thread_safe` gem

jdantonio:
Once I've merged PR #391 I'll rebase this branch and we should be ready to merge and release. I won't release for a few days so that we can have further discussion. Most notably, the consolidation of `attr_volatile` which existed in different forms in both `concurrent-ruby` and `thread_safe`.

### Things Done

* Initial import of the `thread_safe` gem.
* ~~Replaced internal CHM with Java version.~~
* Renamed `Concurrent::Cache` to `Concurrent::Map`.
* ~~Consolidated `attr_volatile` from `concurrent-ruby` and `thread_safe`.~~
* Moved `VolatileTuple` to top level and renamed `Tuple`.
* Replaced `autoload` with `require`.

### Notes

In no particular order:

* The current build of thread_safe includes several Java source files which are copies of files "intended" to be a part of a future version of Java. The necessary classes--`ConcurrentHashMap` specifically--are now available in Java 7 and Java 8. I replaced the internal implementation with the JVM version.
* Using the JVM version of `ConcurrentHashMap` required that I rewrite one API method that depended on missing functionality. This method is `public IRubyObject key(IRubyObject value)`.
* I renamed the Cache class to Map because "map" is a much more common data structure ([Erlang](http://www.erlang.org/doc/man/maps.html), [Go](https://blog.golang.org/go-maps-in-action), and [Clojure](https://clojuredocs.org/clojure.core/map)). It seems to me that map conveys the purpose of the object better than cache, but that's personal opinion.
* Rails uses Cache in several places so we'll need to make the necessary change when submitting a PR to Rails to replace thread_safe.
* Both concurrent-ruby and thread_safe had implementations of `attr_volatile`. Rather than keep them both around I consolidated by keeping the version from thread_safe.
  - The concurrent-ruby version had Java and Rbx specific implementations
  - But the c-r version had fewer features
  - And the c-r version depended on undocumented features of MRI
  - Whereas the t-s version was consistent across all platforms
  - And provided CAS semantics
* I uplifted the Volatile module to `Concurrent::Volatile` and made is a part of the public API.
* I also uplifted VolatileTuple to `Concurrent::Tuple` and documented it. It compliments our thread-safe structs very well. I removed "Volatile" because it might confuse those without understanding of the term in this context and its thread safety can be assumed even without that term.
* The only other possible change is `CheapLockable`. It looks fairly redundant with our `Concurrent::Synchronization::Lock` class. Since it's part of the private API I've left it alone for now, but we should look at it more closely.


jdantonio:
@pitr-ch Please give me your thoughts on `attr_volatile` and `CheapLockable` (see notes above).

pitr-ch:
**Volatile field**: Even though our implementation does not support CAS on fields directly I think we should keep it. It does not introduce indirections by creating AtomicReferences which introduces overhead. Advantage is that it is than really only about volatile field allowing better implementations on given Ruby runtimes. CAS can be easily added by keeping final reference to a AtomicReference (as we do now) which is more explicit and I think good at this level, since the abstractions will be used much more frequently than volatile fields.

We could introduce a utility module for `Synchronisation::Object` which would use final field and AtomicReference to create a CAS accessors though.

**CheapLockable**: We should replace it with `Synchronisation::Object` to avoid duplication, it's a less performant (extra indirection, java interop) subset of what we have. 

Performance statements are my best guess I'll verify during the migration. @jdantonio If agreed and if you like, I'll do the suggested changes.

jdantonio:
@pitr-ch I'm comfortable with both your suggestions. I've rebased this PR agains the latest master (after merging PR #391). I'll leave it alone so that you can work on it. Thank you!

thedarkone:
@jdantonio I'm sorry for taking this long to respond (next time feel free to ping me multiple times :) ).

#### bundled `ConcurrentHashMap.java`

There are 2 reasons for this:
 
* Java 6 and Java 7 support. The new version of `CHM` (the so called `CHMv8`) was introduced in JDK8. At the time that I originally wrote the code, I didn't event attempt to detect Java 8, because then it was in so early stages of development that the `CHMv8` wasn't even part of JDK8, so there was nothing to version-check for. Now, if we're running on JDK8+ we can version check and use `java.util.concurrent.ConcurrentHashMap`.

  But (!) on Java 6 and Java 7 it won't work, because their versions of `CHM` do not provide the `compute` methods, while at the same time it is impossible to "wrap" the earlier versions of `CHM` and implement the `compute` methods atop.

  I see that officially only JDK8+ is supported, so this is a non-issue, unless of course this is an issue for Rails (afterall `thread_safe` works even on Java 6).

* Bundled `CHM` knows how to "properly" compare Ruby objects via `<=>`, but this is a [mixed blessing](https://github.com/headius/thread_safe/issues/19) though. In a grand scheme of things this is a largely irrelevant perf optimization.

#### Cache -> Map rename

I'm fine with this. It just shouldn't ever be `Concurrent::Hash`, since there are very slight API differences and most importantly: no insertion order guarantees.

#### `autoload`

I'm pretty certain I saw a discussion about `autoload` somewhere here on `concurrent-ruby`, but I can't find it now.

Anyways, `autoload` should be thread safe and a very good way of keeping memory footprint of a library low.

Some older discussions:
https://github.com/headius/thread_safe/issues/15
https://github.com/rails/rails/issues/17091

Other than that, I obviously don't care if `autoload`s are replaced by manual `require`s.

pitr-ch:
I think we can merge it with autoload and decide use or not to use later in separate issue. @thedarkone thanks for the links (TL;DR: https://github.com/rails/rails/issues/17091#issuecomment-62419427, https://github.com/headius/thread_safe/issues/15#issuecomment-36153742)

jdantonio:
@thedarkone Thanks for the feedback!

@pitr-ch A few notes:

* Rails 5 requires Ruby 2.2+, which means JRuby 9000. According to @headius JRuby 9000 [runs on Java 7+](https://twitter.com/headius/status/631299675098759168) so we'll have to revert commit 36fd26980f6d6b16d5a51747336c78a60a2fb254 and use the internal implementation of CHM for now.
* We need to update `MriMapBackend` (formerly `MriCacaheBackend`) to have a per-instance mutex rather than a class-level mutex (see discussion [here](https://github.com/ruby-concurrency/thread_safe/issues/22#issuecomment-130116493)).
* At first glance it looks like it should be fairly easy to remove autoloading from the thread_safe code. It doesn't look like there will be any circular requires.
* If we decided to revert to the c-r version of `attr_volatile` we'll need to be on the lookout for `cas_*` methods in the classes that use it. A quick grep of the codebase shows that `AtomicReferenceMapBackend`, `Adder`, `CheapLoclable

jdantonio:
@pitr-ch You mentioned a couple of days ago that you wouldn't mind working on this, but the scope has grown a little. I will have some time to work on it this weekend if need be. Do you want to do it all, or would you like to split the work? I'm cool either way, I'd just like to get this done as soon as reasonably possible so that we can release 1.0.0.pre1 and start testing it.

pitr-ch:
I'll take the last 2, comment updated.

pitr-ch:
I am planning to look at adding CAS for instance variables to get rid of the indirection. After looking at the code, it would a lot of work to migrate now to our current synchronisation and than again after the ivar CAS is added. So for this PR I would suggest to keep current implementation and I address this in following PRs.

jdantonio:
@pitr-ch I think that makes a lot of sense. There is no harm in having two `attr_volatile` implementations for now. We'll keep the one from `thread_safe` private/internal and not publicize it. We can remove it later once we've evolved the other implementation.

pitr-ch:
Hm after thinking about it for a while, the ivar CAS is too ambitious for now (let's aim this for 2.0 or later). Anyway lets merge this after this is ready and I'll start migrating thread_safe to our tools slowly after that.

jdantonio:
### Concurrent::Map vs ThreadSafe::Cache Benchmark Results

```
[12:12:31 jerry ~/Projects/ruby-concurrency/concurrent-ruby (thread_safe)]$ ruby -v
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-linux]

[12:12:48 jerry ~/Projects/ruby-concurrency/concurrent-ruby (thread_safe)]$ bundle exec examples/bench_map.rb
Rehearsal --------------------------------------------------
Hash#[]          3.300000   0.000000   3.300000 (  3.301365)
Map#[]           7.950000   0.000000   7.950000 (  7.950419)
Hash#each_pair   3.750000   0.010000   3.760000 (  3.755541)
Map#each_pair   26.180000   0.050000  26.230000 ( 26.220769)
---------------------------------------- total: 41.240000sec

Hash#[]          4.180000   0.000000   4.180000 (  4.179995)
Map#[]          11.160000   0.060000  11.220000 ( 11.219914)
Hash#each_pair   4.960000   0.000000   4.960000 (  4.959219)
Map#each_pair   26.820000   0.020000  26.840000 ( 26.831965)
```

```
[12:10:35 jerry ~/Projects/ruby-concurrency/thread_safe (master)]$ ruby -v
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-linux]

[12:14:56 jerry ~/Projects/ruby-concurrency/thread_safe (master)]$ bundle exec examples/bench_cache.rb
Rehearsal ---------------------------------------------------
Hash#[]           3.570000   0.010000   3.580000 (  3.564060)
Cache#[]          8.010000   0.000000   8.010000 (  8.015879)
Hash#each_pair    3.640000   0.000000   3.640000 (  3.640242)
Cache#each_pair  23.200000   0.060000  23.260000 ( 23.252722)
----------------------------------------- total: 38.490000sec

Hash#[]           4.740000   0.000000   4.740000 (  4.745926)
Cache#[]          9.760000   0.070000   9.830000 (  9.831186)
Hash#each_pair    4.420000   0.000000   4.420000 (  4.424815)
Cache#each_pair  22.020000   0.040000  22.060000 ( 22.054609)
```

```
[12:20:37 jerry ~/Projects/ruby-concurrency/concurrent-ruby (thread_safe)]$ ruby -v
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar
jruby 1.7.19 (1.9.3p551) 2015-01-29 20786bd on Java HotSpot(TM) 64-Bit Server VM 1.8.0_51-b16 +jit [linux-amd64]

[12:20:41 jerry ~/Projects/ruby-concurrency/concurrent-ruby (thread_safe)]$ bundle exec examples/bench_map.rb
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar
Rehearsal --------------------------------------------------
Hash#[]          2.840000   0.040000   2.880000 (  2.237000)
Map#[]           4.960000   0.570000   5.530000 (  4.910000)
Hash#each_pair   3.430000   0.390000   3.820000 (  3.431000)
Map#each_pair    1.860000   0.030000   1.890000 (  1.739000)
---------------------------------------- total: 14.120000sec

Hash#[]          2.270000   0.000000   2.270000 (  2.157000)
Map#[]           4.400000   0.080000   4.480000 (  4.382000)
Hash#each_pair   3.350000   0.010000   3.360000 (  3.228000)
Map#each_pair    2.850000   0.010000   2.860000 (  2.764000)
```

```
[12:23:49 jerry ~/Projects/ruby-concurrency/thread_safe (master)]$ ruby -v
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar
jruby 1.7.19 (1.9.3p551) 2015-01-29 20786bd on Java HotSpot(TM) 64-Bit Server VM 1.8.0_51-b16 +jit [linux-amd64]

[12:23:53 jerry ~/Projects/ruby-concurrency/thread_safe (master)]$ bundle exec examples/bench_cache.rb
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar
Rehearsal ---------------------------------------------------
Hash#[]           3.390000   0.000000   3.390000 (  2.925000)
Cache#[]          4.850000   0.190000   5.040000 (  4.442000)
Hash#each_pair    5.860000   0.360000   6.220000 (  5.714000)
Cache#each_pair   4.170000   0.000000   4.170000 (  3.757000)
----------------------------------------- total: 18.820000sec

Hash#[]           5.810000   0.000000   5.810000 (  5.794000)
Cache#[]          9.030000   0.350000   9.380000 (  9.141000)
Hash#each_pair    7.130000   0.210000   7.340000 (  7.023000)
Cache#each_pair   2.840000   0.000000   2.840000 (  2.719000)
```

jdantonio:
The failure on Travis for JRuby 9000 isn't really a failure. All tests pass. I don't know why it's showing as a failure. The failing tests on Appveyor are unrelated to this PR. I'm merging.

pitr-ch:
Where could I find bench_cache? Also the benchmarks does not seem to test the collections under contention which I think would show quite different numbers (better for Map).

jdantonio:
It's been renamed to "examples/bench_map.rb"

thedarkone:
> Also the benchmarks does not seem to test the collections under contention which I think would show quite different numbers (better for Map).

No, not necessarily, this is more about correctness under concurrent load. Think about any part of `concurrent-ruby`, what happens if you remove locks or guaranteed volatile reads/writes? Why would anything get slower? In general I would expect things to break, but not really to get slower.

As for slowness of `Map#[]` (vs `Hash#[]`), it stems from this Ruby code https://github.com/ruby-concurrency/concurrent-ruby/blob/d58679c325cd370286bde049636c29c82bb35585/lib/concurrent/map.rb#L51-L63. I opted out for shared pure Ruby (JRuby + MRI + Rubinius) code, it could be optimized by "un-sharing" it (ie write it Java for JRuby etc).

As for slowness of `Map#each_pair` on MRI, this is because of [needed to `dup` the underlying `@backend` `Hash`](https://github.com/ruby-concurrency/concurrent-ruby/blob/d58679c325cd370286bde049636c29c82bb35585/lib/concurrent/thread_safe/non_concurrent_map_backend.rb#L99-L104). If we don't dup and somebody writes into it (during `each_pair`) MRI throws an error (`RuntimeError: can't add a new key into hash during iteration`). This sucks and I don't a way around it.

pitr-ch:
@thedarkone ah sure, I thought it's the synchronised version of Hash not the core one.

PR #396, opened 2015-08-13
New Exchanger Implementation

jdantonio:
Brand new implementation based on [OpenJDK 8](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/concurrent/Exchanger.java?av=f)


jdantonio:
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/concurrent/Exchanger.java

jdantonio:
I *believe* this new implementation is functionally correct. Both the original condition variable implementation and the later MVar implementation had known bugs, which is why we moved Exchanger into Edge. So I don't think there is any harm in merging it, even if we discover bugs later.

@mighe @chrisseaton @pitr-ch I know you are all very busy, but if any of you have a chance to review this PR I would appreciate it.

chrisseaton:
The crucial method for anyone else reviewing this: https://github.com/ruby-concurrency/concurrent-ruby/pull/396/files#diff-8a5aa6ad7f52e71c4ce9a64573329694R163

I don't understand the high-level algorithm here so it's hard to verify that it's a correct algorithm, or that it's correctly implemented. Can you describe the algorithm in a comment, or reference the code where it is from?

jdantonio:
@chrisseaton Done. The algorithm is described (in detail) in the source. Once I wrote it all out I discovered a bug so that was a bonus.

mighe:
I've read carefully both the algorithm and the code, I've also run the code "by hand" looking for race conditions.
It seems ok, I could not find any issue, but I'd like to review again everything on Monday to be sure that it is really ok :)

jdantonio:
This PR has been rebased against the latest master (after thread_safe was merged). I would like to include this PR in the 1.0.0.pre1 release but that is not essential. I'll leave it open for a couple of days to get more review/feedback. We'll release 1.0.0.pre1 next week with or without this PR.

PR #397, opened 2015-08-16
More thread_safe cleanups

pitr-ch:


lib/concurrent.rb
```diff
@@ -121,4 +125,9 @@
 # * Be small, lean, and loosely coupled
 module Concurrent
 
+  # Various classes within allows for +nil+ values to be stored,
+  # so a special +NULL+ token is required to indicate the "nil-ness".
+  # @!visibility private
+  NULL = Object.new
```
jdantonio: :+1:

lib/concurrent/collection/array.rb
```diff
@@ -0,0 +1,38 @@
...
+    # non-thread-safe structures should usually work fine.
+
+    # @!macro [attach] concurrent_array
+    #
+    #   A thread-safe subclass of Array. This version locks against the object
+    #   itself for every method call, ensuring only one thread can be reading
+    #   or writing at a time. This includes iteration methods like `#each`.
+    #
+    #   @see http://ruby-doc.org/core-2.2.0/Array.html Ruby standard library `Array`
+    class Array < ::Array;
```
jdantonio: In the 0.9 release I reconfigured a bunch of classes/modules to conform to more idiomatic namespacing. This change seems like a step back. I have no issues moving files from the `thread_safe` directory into the `collection` directory, but why move the top-level classes into there as well? Is there something I'm missing?

Unless there is a compelling reason to do otherwise, I'd like to keep array.rb, hash.rb, tuple.rb, and map.rb at the top level. We should be able to move all of the backend code from map.rb into a new file w/i the collection folder and namespace those classes idiomatically.

jdantonio:
On a completely unrelated note, I'm going to refactor PriorityQueue today or tomorrow night. I need to add a synchronized, thread-safe version for a feature I want to submit to Rails. If I get that done by Tuesday morning I'd like to slip that into the next release, but I won't hold up the next release any longer than that.

lib/concurrent/collection/array.rb
pitr-ch: Ah sorry I got that confused, I thought we sticked with the previous name-spacing without aligning directories and modules. I am for alignment. The number of collections has quite a potential for growth so to avoid too many classes in Concurrent namespace I wanted to move them from the start. I'll move it back.

I am sorry I know we've discussed our name-spacing policy in the past but I do not remember clearly what the conclusion was. 

1) Any implementation or utility classes should not be in main space. Main space only contains Public classes. Same could be used for requirable files, only `concurrent/*` are intended for requiring by users. Clear, strict, but bigger main namespace.

2) also use sub spaces and use private constants to hide things. Requirable files are those which are documented as such. Less clear, more flexibility, less things in main space, more work with maintaining because of the fuzzier rules.

Let me know what you think and I'll update readme and also synchronisation layer which ended up non-standard.

jdantonio:
I apologize for the confusion on this--I have done a poor job if communicating amongst the team.

For 2.0 I agree that we should be 100% idiomatic in module/file alignment. I'm even open to moving some of the public classes to submodules if that makes sense (`Concurrent::Atomic::Reference` is a possibility I'm open to, for example). The only question is how much do we do *before* the 1.0 release.

:+1: to the guidelines you have suggested.

For 1.0, all of the public classes are in the root namespace so we need to leave them there. I don't have strong feelings about file placement so long as we don't flood the root folder with a ton of files. The `atomic` and `executor` folders are the most cluttered with public classes/modules so we'll need to decide what to do about them in 2.0.

Definitely, non-public classes/modules should be in subfolders and namespaced idiomatically.

As much as possible in 1.x we want to reorganize toward idiomatic class/module/file alignment so long as we don't change the namespaces of public classes, but we don't want to hold back releases solely for the purpose of rearranging files.

jdantonio:
@pitr-ch Since this is just internal reorganization and doesn't change functionality or the public API, I would like to release v1.0.0.pre1. We can continue the reorganization en route to the full 1.0 release. Any objections?

pitr-ch:
I've pushed update for this PR, it can be merged.

If we decide to go with schema 1) I would be strict and use AtomicReference, AtomicFixnum, etc. it depends if we can do similar prefixing in all other cases. Anyway lets figure this out after 1.0 release :)

jdantonio:
@thedarkone What does TS:Cache undef the `freeze` method? The question was [asked](https://gitter.im/ruby-concurrency/concurrent-ruby?at=55d3a7da4f6c9d301831852e) in Gitter. My best [guess](https://gitter.im/ruby-concurrency/concurrent-ruby?at=55d3a7da4f6c9d301831852e) (and probably completely wrong) is that it has something to do with the multiple backend implementations. The [original commit](https://github.com/ruby-concurrency/thread_safe/commit/6aca615675c0714f42d4665f87d2df2f06a04c54) didn't provide much detail. I'm just curious.

thedarkone:
`TS::Cache` attempts to provide strong guarantees about its methods, this allows it to be used as a concurrency primitive. For example: if you call `delete` and a non-`nil` value is returned, you guaranteed not only that the value has been removed, but that it the current thread that did it, the same applies for host of all other methods `compute_if_absent|present`, `replace_pair` etc.

However this wouldn't hold for a `freeze` method, if you `freeze` a `cache` you wouldn't be guaranteed that there are no other threads inside its methods. So, for example this would be allowed:

```ruby
@cache[:foo] = :foo
# other thread starts @cache.delete(:foo)
@cache.freeze
@cache.key?(:foo) # => true
# other thread's @cache.delete(:foo) finishes
@cache.key?(:foo) # => false
```

This of course can be a documented behavior, but I really wanted to avoid having methods that lack strong "assert-ness" about them:
 * "I inserted/swapped/deleted a key, I am guaranteed that it was me (current thread) that did it, all other threads failed",
 * "I failed to insert/swap/deleted a key, this means that exactly one other thread was successful"

Then one might expect:
 * "I froze the cache, no elements will be inserted/deleted/swapped beyond this point"

to also hold true, but this wouldn't be the case.

PS: "strong" `freeze` is unimplementable without perf/scalability consequences.

jdantonio:
@thedarkone Thanks for the response! As always it was very informative.

thedarkone:
A couple more things...

We could add `eventually_freeze`, thus clearly impacting on the users, that this might not do what they expect of normal `freeze`.

I mighty be mistaken about the impossibility of "strong" `freeze` (I now kinda see a away to implement it without perf or scalability issues), but this would mean bundling our own custom `CHM.java` forever. And this also goes against one of the main philosophies of `TS::Cache`: do exactly as Doug Lea does (ie follow the direction and implementation of Java's CHM as closely as possible).

In general, `freeze` (not to be confused with immutability) is a "very MRI" thing that doesn't translate all to well into the concurrent world (same as with `Thread#raise` etc).

jdantonio:
I don't think there's any need for freeze. Projects have been using thread_safe for years without needing that method.

PR #398, opened 2015-08-16
Implemented synchronized PriorityQueue.

jdantonio:
The original priority queue implementation was added as an internal implementation detail of `TimerSet`. It was not thread-safe because TimerSet handled the synchronization. I recently began experimenting with a Rails ActiveJob adapter [built on concurrent-ruby](https://github.com/rails/rails/pull/21257). I began thinking about adding prioritization to the adapter and realized that I might be able to use our priority queue, so I took a stab at creating a thread-safe wrapper around our current implementation.

This is a low-priority item since it doesn't look like my Rails PR is going to be merged. But this update may still be a nice addition to our gem.

This PR does a few things:

* Renames the original priority queue implementation to better reflect that it is not thread-safe.
* Removes the unnecessary method aliases from the non-thread-safe priority queue implementations.
* Implements a synchronized `Concurrent::PriorityQueue` which leverages `Concurrent::Synchronization::Object`.
* Implements the original method aliases on `Concurrent::PriorityQueue`.


jdantonio:
I'm going to close this PR until we have an actual use case. I'm going to cherry-pick commit 677ab3e onto master. The classes need renamed to better reflect their not-thread-safe status.

jdantonio:
Superseded by PR #399

PR #401, opened 2015-08-18
Every class/module in own file

jdantonio:
Early on I started putting multiple classes/modules in one file when the classes/modules were small. Now it's difficult to find things. I'm now correcting that by moving everything into its own file. This is a reorganization only. No new/changed features will be on this PR.

I've covered everything in the `lib/executor` and `lib/atomic` directories.


sschepens:
This is great and should be done for the remaining modules!

pitr-ch:
:+1:

jdantonio:
This was purely selfish. I was working on PR #400 and got lost in the executor files. I needed to reorganize them so that I could work on #404. :-)

PR #404, opened 2015-08-19
Prioritized Thread Pools

jdantonio:
This is an experiment in supporting prioritization in our thread pools. Because Java's thread pools support injectable queues, Java thread pools can be prioritized. This only makes a difference when the number of threads is constrained. When the number of threads can grow without bound nothing will ever be in the queue. It was an interesting experiment and I learned a lot, but I don't feel strongly one way or the other about merging this update.

Does anyone have any thoughts?

I'll do some benchmarking and post the results. It will be interesting what impact this update has had on performance, if any.


lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -269,17 +287,20 @@ class Worker
 
       def initialize(pool)
         # instance variables accessed only under pool's lock so no need to sync here again
-        @queue  = Queue.new
         @pool   = pool
+        # I'm not sure that we need a prioritized queue here
+        # Queued jobs should be in the pool's queue, this is mainly for synchronization
+        #@queue  = pool.prioritized? ? Concurrent::PriorityBlockingQueue.new(order: :max) : Queue.new
+        @queue  = Queue.new
```
jdantonio: @pitr-ch If my understanding is correct, we do not need a prioritized queue here, do we? This queue is mainly used to wake up an idle thread. If jobs are backed up in the queue they will be backed up in the pool's queue. Is this correct?

lib/concurrent/executor/abstract_executor_service.rb
```diff
@@ -10,10 +10,39 @@ module Concurrent
...
+    # @!visibility private
+    #Job = Struct.new(:priority, :args, :task) do
+      #def run
+        #task.call(*args)
+      #end
+      #def <=>(other)
+        #priority <=> other.priority
+      #end
+    #end
+    class Job
```
jdantonio: For those interested in such things, defining this as a class rather than a struct improved performance by about 10 i/100ms in the IPS benchmark.

jdantonio:
I resurrected an old benchmark script from @pitr-ch and updated it for this change. The gist is [here](https://gist.github.com/jdantonio/b8c67dcaff22312995ff). If my question about the per-worker queue is correct and we can use a Ruby Queue even when the pool is prioritized, then the performance of the non-prioritized queues is virtually unchanged. Which is a good thing. Here is a sample run under Ruby 2.2.2:

```
                 old   195.000  i/100ms
                 new   193.000  i/100ms
            priority   109.000  i/100ms
                 old      1.943k (± 2.6%) i/s -     19.500k
                 new      1.917k (± 1.3%) i/s -     19.300k
            priority      1.083k (± 4.0%) i/s -     10.900k

Comparison:
                 old:     1942.6 i/s
                 new:     1917.5 i/s - 1.01x slower
            priority:     1082.6 i/s - 1.79x slower
```

lib/concurrent/executor/ruby_thread_pool_executor.rb
pitr-ch: Yes, exactly.

pitr-ch:
btw what was the rails PR? I've lost the link.

jdantonio:
Rails PR 21257 "Initial implementation of ActiveJob AsyncAdapter." I'm building an in-memory background job processor to be a part of Rails. It will complement the built-in `:inline` adapter. It will be functionally equivalent (or greater than) the sucker_punch gem but built into Rails itself.

jdantonio:
I'm going to close this PR as it has been superseded by #405. Since RubyThreadPoolExecutor does not need PriorityBlockingQueue internally I have removed that latter class from PR #405. Instead, RubySingleThreadExecutor creates its own internal wrapper around NonConcurrentPriorityQueue (the internal collection class used by TimerSet).

PriorityBlockingQueue isn't a necessary part of the public API and it is horribly slow. Better to not have it at all.

PR #405, opened 2015-08-19
Priority thread pools

jdantonio:
The thread pool classes listed below now have a `:prioritize` constructor option that allows jobs waiting in queue to be prioritized. When `false` (the default) the pools work the way they previously did. When `true` the internal queue is prioritized based on an integer value given when the job is post.

This is a continuation of the spikes done in PRs #400 and #404

The `#prioritized?` predicate methods indicates whether the thread pool is working in prioritized mode or not.

The `#post` method works the way it always did. When the pool is prioritized the `#post` method sets the priority to zero.

The new `#prioritize` method is a copy of the `#post` method except that it takes an integer `priority` parameter. When the pool is in non-priority mode the `priority` parameter is ignored.

When operating in non-priority mode the thread pools have nearly identical performance as they did before.

When in priority mode, thread pool performance is degraded slightly, but this is to be expected. Prioritizing the queue involves overhead.

The Java implementation uses `java.util.concurrent.PriorityBlockingQueue` internally so performance is still very good,

### DIFF
--- a/lib/concurrent/executor/abstract_executor_service.rb
+++ b/lib/concurrent/executor/abstract_executor_service.rb
@@ -10,22 +10,32 @@ module Concurrent
   class AbstractExecutorService < Synchronization::Object
     include ExecutorService
 
-    # @!visibility private
-    Job = Struct.new(:priority, :args, :task) do
-      def execute
-        task.call(*args)
+    class Job
+      attr_reader :priority
+
+      def initialize(priority, args, task)
+        @priority = priority
+        @args = args
+        @task = task
       end
+
+      def run
+        @task.call(*@args)
+      end
+
+      def <=>(other)
+        @priority <=> other.priority
+      end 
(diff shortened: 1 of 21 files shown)


jdantonio:
Now that this is working I like it. It's a feature supported by Java. When I googled for information on how to build this there were many hits, indicating it's a popular topic. Since the performance of the pools in non-priority mode isn't degraded it doesn't hurt anything. It doesn't involve very much code since we leveraged classes that already existed (NonConcurrentPriorityQueue, java.util.concurrent.PriorityBlockingQueue) and our internal implementation didn't have to change much.

Unless someone has a strong objection against, I'd like to merge this.

jdantonio:
Travis is suddenly very unhappy with this branch: https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/76307024

PR #408, opened 2015-08-24
Fix if statement in atomic_reference_spec.rb

akihiro17:
Running the test after executing ``` rake clean clobber ``` shows the following errors if we use mri ruby,(ruby 2.1.0, 2.2.2) 

```ruby
> bundle exec rake clean clobber
> bundle exec rspec spec/concurrent/atomic/atomic_reference_spec.rb

1) Concurrent::AtomicReference inherits from CAtomicReference
     Failure/Error: expect(AtomicReference.ancestors).to include(Concurrent::CAtomicReference)
     NameError:
       uninitialized constant Concurrent::CAtomicReference
     # ./spec/concurrent/atomic/atomic_reference_spec.rb:177:in `block (2 levels) in <module:Concurrent>'

  2) Concurrent::AtomicReference inherits from CAtomicReference
     Failure/Error: expect(AtomicReference.ancestors).to include(Concurrent::CAtomicReference)
     NameError:
       uninitialized constant Concurrent::CAtomicReference
     # ./spec/concurrent/atomic/atomic_reference_spec.rb:215:in `block (2 levels) in <module:Concurrent>'
```

No error occurs if we execute ``` rake compile ``` before running the test.

```ruby
> bundle exec rake compile
> bundle exec rspec spec/concurrent/atomic/atomic_reference_spec.rb

52 examples, 0 failures
```

In this commit, the `'Concurrent::AtomicReference inherits from CAtomicReference'` test cases are run only if `@c_ext_loaded` is true.

### DIFF
--- a/lib/concurrent/utility/native_extension_loader.rb
+++ b/lib/concurrent/utility/native_extension_loader.rb
@@ -14,6 +14,10 @@ module Concurrent
       def allow_c_extensions?
         Concurrent.on_cruby?
       end
+   
+      def loaded_c_extensions?
+        NativeExtensionLoader.instance_variable_get(:@c_ext_loaded)
+      end
 
       if Concurrent.on_cruby? && !@c_ext_loaded
         tries = [
--- a/spec/concurrent/atomic/atomic_reference_spec.rb
+++ b/spec/concurrent/atomic/atomic_reference_spec.rb
@@ -172,7 +172,7 @@ module Concurrent
       it 'inherits from JavaAtomicReference' do
         expect(AtomicReference.ancestors).to include(Concurrent::JavaAtomicReference)
       end
-    elsif Concurrent.allow_c_extensions?
+    elsif Concurrent.loaded_c_extensions?
       it 'inherits from CAtomicReference' do
         expect(AtomicReference.ancestors).to include(Concurrent::CAtomicReference)
       end
@@ -210,7 +210,7 @@ module Concurrent
       it 'inherits from JavaAtomicReference' do
         expect(AtomicReference.ancestors).to include(Concurrent::JavaAtomicReference)
       end
-    elsif Concurrent.allow_c_extensions?
+    elsif Concurrent.loaded_c_extensions?
       it 'inherits from CAtomicReference' do
         expect(AtomicReference.ancestors).to include(Concurrent::CAtomicReference)
       end


lib/concurrent/utility/native_extension_loader.rb
```diff
@@ -14,6 +14,10 @@ module NativeExtensionLoader
       def allow_c_extensions?
         Concurrent.on_cruby?
       end
+   
+      def loaded_c_extensions?
+        NativeExtensionLoader.instance_variable_get(:@c_ext_loaded)
```
ianks: Is the instance_variable_get necessary here? I noticed a few lines below we were not using it.

h-izu: I think instance_variable get might be needed because ``` @c_ext_loaded ``` is nil in the context of the ``` Concurrent ``` module and it is true in the context of the ``` 
NativeExtensionLoader ``` module.
```ruby
module Concurrent
  module Utility
    module NativeExtensionLoader
      def c_ext_loaded
        @c_ext_loaded
      end
      def c_ext_loaded_with_iv_get
        NativeExtensionLoader.instance_variable_get(:@c_ext_loaded)        
      end
    end
  end
 ～～～～～～～～
 ～～～～～～～～～
end

Concurrent.c_ext_loaded? # => nil
Concurrent.c_ext_loaded_with_iv_get # => true
```

akihiro17: I think instance_variable get might be needed because ``` @c_ext_loaded ``` is nil in the context of the ``` Concurrent ``` module and it is true in the context of the ``` 
NativeExtensionLoader ``` module.
```ruby
module Concurrent
  module Utility
    module NativeExtensionLoader
      def c_ext_loaded
        @c_ext_loaded
      end
      def c_ext_loaded_with_iv_get
        NativeExtensionLoader.instance_variable_get(:@c_ext_loaded)        
      end
    end
  end
 ～～～～～～～～
 ～～～～～～～～～
end

Concurrent.c_ext_loaded? # => nil
Concurrent.c_ext_loaded_with_iv_get # => true
```

jdantonio:
@akihiro17 Thank you very much for taking the time to submit this PR. Unfortunately, your suggested change doesn't fit with the needs of the changed tests.

When the C-extension gem is installed the core gem will automatically load the extensions. It is possible for a bug to allow the extensions to compile but prevent them from loading. Should a bug of this nature occur (which has happened in the past) those tests will fail on Travis and Appveyor, alerting us to the bug. With your change, a bug like that would cause the tests to be *skipped*, thus masking the bug.

I am definitely open to changes that would allow the test suite, when run locally, to skip the C extension tests if the extensions have not been compiled. On Travis and Appveyor, however, it is critical that tests fail if the C extensions do not load.

akihiro17:
I see. I'll close this PR.

Thanks.

jdantonio:
@akihiro17 After looking at this PR I discovered that the tests for our other C classes (such as `CAtomicBoolean`) don't work the way the ones for AtomicReference do. Those tests will mask the potential load problems I described above. I wouldn't have noticed that without your PR so this was very helpful. Thank you!

PR #409, opened 2015-08-26
Fix interpreter warnings

jdantonio:
I'm still trying to get rid of all the warnings that display when we load the gem. Most notable, the "circular reference" warnings. This is my next attempt. This one removes all warnings but one and all the tests still pass. It needs much more testing.

To display warnings, run this:

```
bundle exec ruby -v -e "require 'concurrent'"
```

Issue #395, opened 2015-08-12
Do we embrace Kernel#autoload ?


pitr-ch:
Discussed here https://github.com/ruby-concurrency/concurrent-ruby/pull/386

pitr-ch:
I would rather not use it. https://github.com/rails/rails/issues/17091#issuecomment-62419427 and https://github.com/headius/thread_safe/issues/15#issuecomment-36153742 mentions that it does not play well with different autoload mechanisms. Since we are only library we cannot control what else is pulled to the application. It may not work well together - needs further investigation.

sschepens:
I don't see much benefit in using it, plus it implies overhead during runtime instead of on load and I almost never prefer that.

thedarkone:
> plus it implies overhead during runtime

What do you mean?

pitr-ch:
more random thoughts: if we fully migrate to autoload, users will only require 'concurrent', rest of the require points will be removed and its manually maintained require dependencies too. So for us it would be simpler.

thedarkone:
Ad @pitr-ch's last comment:

While at #409 (in a now invisible @jdantonio's comment about some broken code) it just occurred to me, that he said he was running Rails and checking for warnings, but the code responsible for warnings was Rubinius code (that should only ever be required and evaled on rbx) and at the same time I was pretty certain that Rbx doesn't run Rails `master` tests right now (or at least it must be quite difficult to do). So it must have been that he was running MRI that was emitting warnings for Rbx specific `.rb` files, so it means `requires` are not 100% correct and while I'm pretty certain that this wouldn't ever occur with `autoload`.

It is unbelievably annoying to keep `require`s at absolute provable minimum (as in: "all `require` lines are necessary for library to load, removing even a single one will break things").

PS: `concurrent-ruby` is a much bigger lib than `thread_safe` ever was, so there might be some problems with `autoload` that never showed in `thread_safe`, but will pose a problem for `concurrent-ruby`.

jdantonio:
@thedarkone You ate correct, I was getting that warning on MRI. The reason we were loading Rbx files on MRI was located in `utils.rb`, which I've updated in 525cd1801e77129866abe84706da0b1b07369e36.

[...]

PR #411, opened 2015-08-27
[DO NOT MERGE] Spike new Delay implementation.

jdantonio:
This is a spike of a possible new Delay implementation. It arose from the conversation in #395 (see my comment [here](https://github.com/ruby-concurrency/concurrent-ruby/issues/395#issuecomment-135403878)).

The issue: There is a circular dependency between the 0.9.x Delay class and the global configuration. This implementation removes the circular dependency. It breaks the API from 0.9.x, but does not break any internal use. Note:

* PR #409 *also* resolved the circular dependency by `autoload`ing `Options` and moving the call to `Options` from the constructor to the execution method. That PR has been merged.
* The new `Future` implementation in the `Edge` module creates a new `Delay` implementation. That code, which defines the v2.0, will not be released unless it has *no* circular dependencies.

Although I like this implementation much better (it's closer to the original version) we've solved the circular reference problem in PR #409. So I'm probably going to keep what we have and close this PR. I just wanted to spike it for reference.

### DIFF
--- a/lib/concurrent/delay.rb
+++ b/lib/concurrent/delay.rb
@@ -1,150 +1,58 @@
-require 'thread'
-require 'concurrent/concern/obligation'
-require 'concurrent/executor/immediate_executor'
-require 'concurrent/synchronization'
+require 'concurrent/atomic/count_down_latch'
+require 'concurrent/synchronization/object'
 
 module Concurrent
 
-  autoload :Options, 'concurrent/options'
-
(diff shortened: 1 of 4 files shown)


pitr-ch:
I would keep old impl. and use this as internal impl. (`#realize_via` may not be needed) where needed. (followup of https://github.com/ruby-concurrency/concurrent-ruby/issues/395#issuecomment-135487502)

PR #412, opened 2015-08-28
New Agent implementation.

jdantonio:
Needs yardoc, but functionally complete.

### DIFF
--- a/lib/concurrent/agent.rb
+++ b/lib/concurrent/agent.rb
@@ -1,260 +1,579 @@
+require 'concurrent/configuration'
+require 'concurrent/atomic/atomic_reference'
+require 'concurrent/atomic/thread_local_var'
 require 'concurrent/collection/copy_on_write_observer_set'
-require 'concurrent/concern/dereferenceable'
 require 'concurrent/concern/observable'
-require 'concurrent/concern/logging'
-require 'concurrent/synchronization'
+require 'concurrent/synchronization/object'
 
 module Concurrent
 
-  autoload :Options, 'concurrent/options'
-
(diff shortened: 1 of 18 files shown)


pitr-ch:
Does it integrate with TVar as well?

jdantonio:
It does not integrate with TVar. That will be documented in the Yardoc. There is also a small bug with the "await" functions where there are nested actions. I will fix that today.

lib/concurrent/agent.rb
```diff
@@ -1,260 +1,579 @@
...
-  #   score << proc{|current| sleep(0.1); current += 30 }
-  #   score << proc{|current| sleep(0.1); current += 30 }
-  #   
-  #   sleep(1)
-  #   #=> Bingo! [score: 120, time: 2013-07-22 21:26:08 -0400]
+  #   **NOTE** Never, *under any circumstances*, call any of the "await" methods
+  #   ({#await}, {#await_for}, {#await_for!}, and {#wait}) from within an action
+  #   block/proc/lambda. The call will block the Agent and will always fail.
+  #   Calling either {#await} or {#wait} (with a timeout of `nil`) will
+  #   hopelessly deadlock the Agent with no possibility of recovery.
```
pitr-ch: Could we raise instead?

jdantonio: Probably. Interestingly, this is exactly the same as Clojure. I was surprised that Clojure took this approach but it is documented behavior.

pitr-ch: Hm, I would not have expected that in clojure. I would raise if it can be already detected (to protect our users) otherwise let's keep it as is.

PR #422, opened 2015-09-24
Add documentation for informal Ruby memory model and synchronisation …

pitr-ch:
…extensions.

### DIFF
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -32,17 +32,16 @@ time-consuming and error-prone.
 # Ruby memory model
 
 The Ruby memory model is a framework allowing to reason about programs in
-concurrent and parallel environment. It allows to identify what is and what
-is not a [race condition](https://en.wikipedia.org/wiki/Race_condition).
-The memory model is also a contract: if a program is written without race
-conditions it behaves in
-[sequential consistent](https://en.wikipedia.org/wiki/Sequential_consistency)
-manner.
+concurrent and parallel environment. It defines what variable writes can be
+observed by a particular variable read, which is essential to be able to
+determine if a program is correct. It is achieved be defining what subset of
+all possible program execution orders is allowed.
 
 A memory model sources:
 
 -   [Java memory model](http://www.cs.umd.edu/~pugh/java/memoryModel/),
     and its [FAQ](http://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html)
+-   [Java Memory Model Pragmatics](http://shipilev.net/blog/2014/jmm-pragmatics/)
 -   [atomic&lt;&gt; Weapons 1](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-1-of-2)
 and
 [2](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-2-of-2)
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -39,14 +39,28 @@ conditions it behaves in
 [sequential consistent](https://en.wikipedia.org/wiki/Sequential_consistency)
 manner.
 
-Sources:
+A memory model sources:
 
 -   [Java memory model](http://www.cs.umd.edu/~pugh/java/memoryModel/),
     and its [FAQ](http://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html)
 -   [atomic&lt;&gt; Weapons 1](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-1-of-2)
 and
 [2](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-2-of-2)
-- [JRuby's wiki page about concurrency](https://github.com/jruby/jruby/wiki/Concurrency-in-jruby)
+
+Concurrent behavior sources of Ruby implementations:
+
+-   Source codes.
+-   [JRuby's wiki page](https://github.com/jruby/jruby/wiki/Concurrency-in-jruby)
+-   [Rubinius's wiki page](http://rubini.us/doc/en/systems/concurrency/)
+
+> A similar document for MRI was not found. Key fact about MRI is GVL (Global
+> VM lock) which ensures that only one thread can interpret a Ruby code at any
+> given time. When the GVL is handed from one thread to another a mutex is
+> released by first and acquired by the second thread implying that everything
+> done by first thread is visible to second thread. See
+> [thread_pthread.c](https://github.com/ruby/ruby/blob/ruby_2_2/thread_pthread.c#L101-L107)
+> and
+> [thread_win32.c](https://github.com/ruby/ruby/blob/ruby_2_2/thread_win32.c#L95-L100).
 
 This memory model was created by: comparing
 [MRI](https://www.ruby-lang.org/en/), [JRuby](http://jruby.org/),
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -71,8 +71,13 @@ C++11). This is not a formal model.
 Key properties are:
 
 -   **volatility (V)** - Any written value is immediately visible to any
-    subsequent volatile reads including all writes leading to this value. (Same
-    meaning as in Java.)
+    subsequent volatile read of the same variable. Any write executed before
+    volatile write based on program-order is visible to the read as well, which
+    allows
+    [piggybacking](http://stackoverflow.com/questions/8769570/volatile-piggyback-is-this-enough-for-visiblity).
+    (Same meaning as in Java, it creates synchronizes-with (JMM term) order
+    between write and read, which participates in creating happens-before
+    order.)
 -   **atomicity (A)** - Operation is either done or not as a whole.
 -   **serialized (S)** - Operations are serialized in some order (they
     cannot disappear). This is a new property not mentioned in other memory
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -100,8 +100,9 @@ Notes:
 -   Variable read reads value from preexisting variable.
 -   Variable definition creates new variable (operation is serialized with
     writes, implies an update cannot be lost).
--   Module/Class definition is actually constant definition. It is defined
-    instantly, however its methods are then processed sequentially.
+-   A Module or a Class definition is actually a constant definition.
+    The definition is atomic, it assigns the Module or the Class to the
+    constant, then its methods are defined atomically one by one.
 -   `||=`, `+=`, etc. are actually two operations read and write which implies
     that it's not an atomic operation. See volatile variables
     with compare-and-set.
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -139,8 +139,10 @@ Notes:
 ### Core classes
 
 `Mutex`, `Monitor`, `Queue` have to work correctly on each implementation. Ruby
-implementations should not crash when e.g. Array is used in parallel environment
-but it may loose updates etc.
+implementation VMs should not crash when for example `Array` or `Hash` is used
+in parallel environment but it may loose updates, or raise Exceptions. (If
+`Array` or `Hash` were synchronized it would have too much overhead when used
+in a single thread.)
 
 > TODO: This section needs more work: e.g. Thread.raise and similar is an open
 > issue, better not to be used.
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -128,7 +128,7 @@ Current Implementation differences from the model:
 | Operation | V | A | S | Notes |
 |:----------|:-:|:-:|:-:|:-----|
 | requiring | x | x | x | File will not be required twice, classes and modules are still defined gradually. |
-| autoload | x | x | x | Only one autoload at a time. |
+| autoload | x | x | - | Only one autoload at a time for a given constant, others will be blocked until first triggered autoload is done. Different constants may be loaded concurrently. |
 
 Notes:
 
--- a/doc/synchronization.md
+++ b/doc/synchronization.md
@@ -1,0 +1,356 @@
+> Quotations are used for notes.
+
+> This document is work-in-progress.
+> Intentions of this effort and document are: to sum up the behavior
+> of Ruby in concurrent and parallel environment, initiate discussion,
+> identify problems in the document, find flaws in the Ruby
+> implementations if any, suggest what has to be enhanced in Ruby itself
+> and cooperate towards the goal in all implementations (using
+> `concurrent-ruby` as compatibility layer).
+
+# Synchronization
+
+This layer provides tools to write concurrent abstractions independent on any
+particular Ruby implementation. It is build on top of Ruby memory model which is
+also described here. `concurrent-ruby` abstractions are build using this layer.
+
+**Why?** Ruby is great expressive language, but it lacks in support for
+concurrent and parallel computation. It's hoped that this document will provide
+ground steps for Ruby to become as good in this area as in others.
+
+Without memory model and this layer it's very hard to write concurrent
+abstractions for Ruby. To write a proper concurrent abstraction it often means
+to reimplement it more than once for different Ruby runtimes, which is very
+time-consuming and error-prone.
+
+# Ruby memory model
+
+Ruby memory model is a framework allowing to reason about programs in concurrent
+and parallel environment. It allows to identify what is and what is not a [race
+condition](https://en.wikipedia.org/wiki/Race_condition). Memory model is also a
+contract: if a program is written without the race conditions it behaves in
+[sequential consistent](https://en.wikipedia.org/wiki/Sequential_consistency)
+manner.
+
+Sources:
+
+-   [Java memory model](http://www.cs.umd.edu/~pugh/java/memoryModel/),
+    and its [FAQ](http://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html)
+-   [atomic&lt;&gt; Weapons 1](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-1-of-2)
+and
+[2](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-2-of-2)
+- [JRuby's wiki page about concurrency](https://github.com/jruby/jruby/wiki/Concurrency-in-jruby)
+
+This memory model was created by: comparing
+[MRI](https://www.ruby-lang.org/en/), [JRuby](http://jruby.org/),
+[JRuby+Truffle](https://github.com/jruby/jruby/wiki/Truffle),
+[Rubinius](http://rubini.us/); taking account limitations of the implementations
+or their platforms; inspiration drawn from other existing memory models (Java,
+C++11). This is not a formal model.
+
+Key properties are:
+
+-   **volatility (V)** - Any written value is immediately visible to any
+    subsequent volatile reads including all writes leading to this value. (Same
+    meaning as in Java.)
+-   **atomicity (A)** - Operation is either done or not as a whole.
+-   **serialized (S)** - Operations are serialized in some order (they
+    cannot disappear). This is a new property not mentioned in other memory
+    models, since Java and C++ do not have dynamically defined fields. All
+    operations on one line in the table are serialized with each other.
+
+### Core behavior:
+
+| Operation | V | A | S | Notes |
+|:----------|:-:|:-:|:-:|:-----|
+| local variable read/write/definition | - | x | x | Local variables are determined during parsing, they are not usually dinamically added (with exception of `local_variable_set`). Therefore definition is quite rare. |
+| instance variable read/write/(un)definition | - | x | x ||
+| class variable read/write/(un)definition | x | x | x ||
+| global variable read/write/definition | x | x | x | un-define us not possible currently. |
+| constant variable read/write/(un)definition | x | x | x ||
+| `Thread` local variable read/write/definition | - | x | x | un-define us not possible currently. |
+| `Fiber` local variable read/write/definition | - | x | x | un-define us not possible currently. |
+| method creation/redefinition/removal | x | x | x ||
+| include/extend | x | x | x | If `AClass` is included `AModule`, `AClass` gets all `AModule`'s methods at once. |
+
+
+Notes:
+
+-   Variable read reads value from preexisting variable.
+-   Variable definition creates new variable (operation is serialized with
+    writes, implies an update cannot be lost).
+-   Module/Class definition is actually constant definition. It is defined
+    instantly, however its methods are then processed sequentially.
+-   `||=`, `+=`, etc. are actually two operations read and write which implies
+    that it's not an atomic operation. See volatile variables
+    with compare-and-set.
+-   Method invocation does not have any special properties that includes
+    object initialization.
+
+Implementation differences from the model:
+
+-   MRI: everything is volatile.
+-   JRuby: `Thread` and `Fiber` local variables are volatile. Instance
+    variables are volatile on x86 and people may un/intentionally depend
+    on the fact.
+-   Class variables require investigation.
+
+### Source loading:
+
+| Operation | V | A | S | Notes |
+|:----------|:-:|:-:|:-:|:-----|
+| requiring | x | x | x | File will not be required twice, classes and modules are still defined gradually. |
+| autoload | x | x | x | Only one autoload at a time. |
+
+Notes:
+
+-   Beware of requiring and autoloading in concurrent programs, it's possible to
+    see partially defined classes. Eager loading or blocking until class is
+    fully loaded has should be used to mitigate.
+
+### Core classes
+
+`Mutex`, `Monitor`, `Queue` have to work correctly on each implementation. Ruby
+implementations should not crash when e.g. Array is used in parallel environment
+but it may loose updates etc.
+
+> TODO: This section needs more work: e.g. Thread.raise and similar is an open
+> issue, better not to be used.
+
+### Standard libraries
+
+Standard libraries were written for MRI so unless they are rewritten in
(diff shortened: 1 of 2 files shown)


pitr-ch:
Hi I am sharing the Ruby informal memory model for first round of comments, then I intend to widen the audience after first round of comments is addressed and silly mistakes removed. Thanks for taking time to look at it. Please invite whoever you would like to, just do not tweet (etc.) yet.

cc @chrisseaton @eregon @jdantonio @headius.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
+> Quotations are used for notes.
+
+> This document is work-in-progress.
+> Intentions of this effort and document are: to sum up the behavior
```
chrisseaton: Could be 'to summarise' rather than 'to sum up'. Sum up sounds more like a conclusion.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+> of Ruby in concurrent and parallel environment, initiate discussion,
+> identify problems in the document, find flaws in the Ruby
+> implementations if any, suggest what has to be enhanced in Ruby itself
+> and cooperate towards the goal in all implementations (using
+> `concurrent-ruby` as compatibility layer).
```
chrisseaton: Maybe as in a JEP, list non-goals as well as goals to limit scope creep. For example it is a non-goal to add high-level new language features.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+
+# Synchronization
+
+This layer provides tools to write concurrent abstractions independent on any
+particular Ruby implementation. It is build on top of Ruby memory model which is
+also described here. `concurrent-ruby` abstractions are build using this layer.
+
+**Why?** Ruby is great expressive language, but it lacks in support for
```
chrisseaton: I think it 'lacks in support for well-defined low-level concurrent and parallel computation', rather than any parallel or concurrent computation, as it can do that with locks etc.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+concurrent and parallel computation. It's hoped that this document will provide
+ground steps for Ruby to become as good in this area as in others.
+
+Without memory model and this layer it's very hard to write concurrent
```
chrisseaton: 'Without *a* memory model'

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+abstractions for Ruby. To write a proper concurrent abstraction it often means
+to reimplement it more than once for different Ruby runtimes, which is very
+time-consuming and error-prone.
+
+# Ruby memory model
+
+Ruby memory model is a framework allowing to reason about programs in concurrent
```
chrisseaton: '*The* Ruby memory model'

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+and parallel environment. It allows to identify what is and what is not a [race
+condition](https://en.wikipedia.org/wiki/Race_condition). Memory model is also a
+contract: if a program is written without the race conditions it behaves in
+[sequential consistent](https://en.wikipedia.org/wiki/Sequential_consistency)
+manner.
```
chrisseaton: In general 'memory model' always needs an article - either 'the' memory model, or 'a' memory model, but never simply 'memory model'.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+abstractions for Ruby. To write a proper concurrent abstraction it often means
+to reimplement it more than once for different Ruby runtimes, which is very
+time-consuming and error-prone.
+
+# Ruby memory model
+
+Ruby memory model is a framework allowing to reason about programs in concurrent
+and parallel environment. It allows to identify what is and what is not a [race
+condition](https://en.wikipedia.org/wiki/Race_condition). Memory model is also a
+contract: if a program is written without the race conditions it behaves in
```
chrisseaton: But 'race conditions' does not need an article as it's plural :) So simply 'without race conditions'

chrisseaton: I'm getting a bit pedantic here, but race conditions are not simply an evil thing. There are several types of race conditions, and here we probably mean data races. Even data races can be part of a well designed application. For example in Go if you have two go-routines and select on them both to get a single value, then that's a race, but that may be what you want.

From the Encylopedia of Parallel Computing (David Padua ed.):

> A race condition occurs in a parallel program execution when two or more threads access a common resource, e.g., a variable in shared memory, and the order of the accesses depends on the timing, i.e., the progress of individual threads.
> The disposition for a race condition is in the parallel program. In different executions of the same program on the same input access events that constitute a race can occur in different order, which may but does not generally result in different program behaviors (non- determinacy).
> Race conditions that can cause unintended non- determinacy are programming errors. Examples of such programming errors are violations of atomicity due to incorrect synchronization.

So a race condition is not an inherently error! It's just a property of the program.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+> Intentions of this effort and document are: to sum up the behavior
+> of Ruby in concurrent and parallel environment, initiate discussion,
+> identify problems in the document, find flaws in the Ruby
+> implementations if any, suggest what has to be enhanced in Ruby itself
+> and cooperate towards the goal in all implementations (using
+> `concurrent-ruby` as compatibility layer).
+
+# Synchronization
+
+This layer provides tools to write concurrent abstractions independent on any
```
eregon: on => of

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+manner.
+
+Sources:
+
+-   [Java memory model](http://www.cs.umd.edu/~pugh/java/memoryModel/),
+    and its [FAQ](http://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html)
+-   [atomic&lt;&gt; Weapons 1](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-1-of-2)
+and
+[2](https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-2-of-2)
+- [JRuby's wiki page about concurrency](https://github.com/jruby/jruby/wiki/Concurrency-in-jruby)
```
chrisseaton: I think we should think about how we talk about JRuby, Java and the JMM. A concern from casual readers could be that we are putting the JMM into Ruby or forcing the JRuby way of doing things onto MRI. How do we handle this question?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+> of Ruby in concurrent and parallel environment, initiate discussion,
+> identify problems in the document, find flaws in the Ruby
+> implementations if any, suggest what has to be enhanced in Ruby itself
+> and cooperate towards the goal in all implementations (using
+> `concurrent-ruby` as compatibility layer).
+
+# Synchronization
+
+This layer provides tools to write concurrent abstractions independent on any
+particular Ruby implementation. It is build on top of Ruby memory model which is
```
eregon: built ... the Ruby MM ...

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+[MRI](https://www.ruby-lang.org/en/), [JRuby](http://jruby.org/),
+[JRuby+Truffle](https://github.com/jruby/jruby/wiki/Truffle),
+[Rubinius](http://rubini.us/); taking account limitations of the implementations
+or their platforms; inspiration drawn from other existing memory models (Java,
+C++11). This is not a formal model.
+
+Key properties are:
+
+-   **volatility (V)** - Any written value is immediately visible to any
+    subsequent volatile reads including all writes leading to this value. (Same
```
chrisseaton: Not sure what 'including all writes leading to this value' means

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+    writes, implies an update cannot be lost).
+-   Module/Class definition is actually constant definition. It is defined
+    instantly, however its methods are then processed sequentially.
+-   `||=`, `+=`, etc. are actually two operations read and write which implies
+    that it's not an atomic operation. See volatile variables
+    with compare-and-set.
+-   Method invocation does not have any special properties that includes
+    object initialization.
+
+Implementation differences from the model:
```
chrisseaton: Differences at the moment, or permanent differences?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+| include/extend | x | x | x | If `AClass` is included `AModule`, `AClass` gets all `AModule`'s methods at once. |
+
+
+Notes:
+
+-   Variable read reads value from preexisting variable.
+-   Variable definition creates new variable (operation is serialized with
+    writes, implies an update cannot be lost).
+-   Module/Class definition is actually constant definition. It is defined
+    instantly, however its methods are then processed sequentially.
```
chrisseaton: I don't understand the difference between 'instantly' and 'sequentially'.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+    with compare-and-set.
+-   Method invocation does not have any special properties that includes
+    object initialization.
+
+Implementation differences from the model:
+
+-   MRI: everything is volatile.
+-   JRuby: `Thread` and `Fiber` local variables are volatile. Instance
+    variables are volatile on x86 and people may un/intentionally depend
+    on the fact.
```
chrisseaton: Why volatile on x86 but not other platforms?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+
+-   Beware of requiring and autoloading in concurrent programs, it's possible to
+    see partially defined classes. Eager loading or blocking until class is
+    fully loaded has should be used to mitigate.
+
+### Core classes
+
+`Mutex`, `Monitor`, `Queue` have to work correctly on each implementation. Ruby
+implementations should not crash when e.g. Array is used in parallel environment
+but it may loose updates etc.
```
chrisseaton: I think this is absolutely the right decision - but I think it needs some narrative to justify. We do not syncohronise Array etc on your behalf. You need to use locks on them when you use them concurrently, but this means you do not.

Also, what does crash mean? JRuby will give you concurrent modification exceptions, which are normal Ruby exceptions. I think Rubinius might as well. Does that count as a crash?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+Key properties are:
+
+-   **volatility (V)** - Any written value is immediately visible to any
+    subsequent volatile reads including all writes leading to this value. (Same
+    meaning as in Java.)
+-   **atomicity (A)** - Operation is either done or not as a whole.
+-   **serialized (S)** - Operations are serialized in some order (they
+    cannot disappear). This is a new property not mentioned in other memory
+    models, since Java and C++ do not have dynamically defined fields. All
+    operations on one line in the table are serialized with each other.
```
eregon: in a row

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+class ImmutableTreeNode < Concurrent::Synchronization::Object
+  # mark this class to publish final instance variables safely
+  safe_initialization!
+
+  def initialize(left, right)
+    # Call super to allow proper initialization.
+    super()
+    # By convention final instance variables have CamelCase names
+    # to distinguish them from ordinary instance variables.
+    @Left  = left
```
chrisseaton: Does this capitalisation mean something?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Volatile instance variable
+
+`Synchronization::Object` children can have volatile instance variables. A Ruby
+library cannot alter meaning of `@a_name` expression therefore when a
+`attr_volatile :a_name` is called, declaring the instance variable named
+`a_name` to be volatile, it creates method accessors.
+
+> However there is Ruby [issue](https://redmine.ruby-lang.org/issues/11539)
+> filled to address this.
```
chrisseaton: '*filed*'

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+-   **serialized (S)** - Operations are serialized in some order (they
+    cannot disappear). This is a new property not mentioned in other memory
+    models, since Java and C++ do not have dynamically defined fields. All
+    operations on one line in the table are serialized with each other.
+
+### Core behavior:
+
+| Operation | V | A | S | Notes |
+|:----------|:-:|:-:|:-:|:-----|
+| local variable read/write/definition | - | x | x | Local variables are determined during parsing, they are not usually dinamically added (with exception of `local_variable_set`). Therefore definition is quite rare. |
```
eregon: dynamically.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+| instance variable read/write/(un)definition | - | x | x ||
+| class variable read/write/(un)definition | x | x | x ||
+| global variable read/write/definition | x | x | x | un-define us not possible currently. |
+| constant variable read/write/(un)definition | x | x | x ||
+| `Thread` local variable read/write/definition | - | x | x | un-define us not possible currently. |
```
eregon: *is* not possible

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+    models, since Java and C++ do not have dynamically defined fields. All
+    operations on one line in the table are serialized with each other.
+
+### Core behavior:
+
+| Operation | V | A | S | Notes |
+|:----------|:-:|:-:|:-:|:-----|
+| local variable read/write/definition | - | x | x | Local variables are determined during parsing, they are not usually dinamically added (with exception of `local_variable_set`). Therefore definition is quite rare. |
+| instance variable read/write/(un)definition | - | x | x ||
+| class variable read/write/(un)definition | x | x | x ||
```
eregon: Why this is magically volatile? class vars are also mostly deprecated AFAIK.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+| `Fiber` local variable read/write/definition | - | x | x | un-define us not possible currently. |
+| method creation/redefinition/removal | x | x | x ||
+| include/extend | x | x | x | If `AClass` is included `AModule`, `AClass` gets all `AModule`'s methods at once. |
+
+
+Notes:
+
+-   Variable read reads value from preexisting variable.
+-   Variable definition creates new variable (operation is serialized with
+    writes, implies an update cannot be lost).
```
eregon: For all variables? even locals? or just ivars?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+    cannot disappear). This is a new property not mentioned in other memory
+    models, since Java and C++ do not have dynamically defined fields. All
+    operations on one line in the table are serialized with each other.
+
+### Core behavior:
+
+| Operation | V | A | S | Notes |
+|:----------|:-:|:-:|:-:|:-----|
+| local variable read/write/definition | - | x | x | Local variables are determined during parsing, they are not usually dinamically added (with exception of `local_variable_set`). Therefore definition is quite rare. |
+| instance variable read/write/(un)definition | - | x | x ||
```
eregon: Maybe this should mention that an ivar which is defined must appear eventually?

doc/synchronization.md
eregon: The fact that you can use `Thread#[]` or `Thread#thread_variable_get` on another thread seems like bad design. Indeed, not guaranteeing that it is volatile makes sense to me.

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+    variables are volatile on x86 and people may un/intentionally depend
+    on the fact.
+-   Class variables require investigation.
+
+### Source loading:
+
+| Operation | V | A | S | Notes |
+|:----------|:-:|:-:|:-:|:-----|
+| requiring | x | x | x | File will not be required twice, classes and modules are still defined gradually. |
+| autoload | x | x | x | Only one autoload at a time. |
```
eregon: Why only one at a time? To ensure partial classes are not seen?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+
+Notes:
+
+-   Beware of requiring and autoloading in concurrent programs, it's possible to
+    see partially defined classes. Eager loading or blocking until class is
+    fully loaded has should be used to mitigate.
```
eregon: -has

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+    self.value = value
+  end
+
+  # Very cheap reader of the Counter's current value, just a volatile read.
+  def count
+    # volatile read
+    value
+  end
+
+  # Safely increment the value without loosing updates
```
eregon: increments

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+  # (as it would happen with just += used).
+  def increment(add)
+    # Wrap he two volatile operations to make them atomic.
```
eregon: Wrap the

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+      # volatile write and read
+      self.value = self.value + add
+    end
+  end   
+end
+```
+
+> This is currently planned to be migrated to a module to be able to add
+> volatile fields any object not just `Synchronization::Object` children. The
+> instance variable itself is named `@"volatile_#(name)"` to distinguish it and
```
eregon: interesting notation, I would go with either the quotes outside and #{} or just `@volatile_<name>`

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+## Concurrent Ruby Notes
+
+### Locks
+
+Concurrent Ruby also has an internal extension of `Object` called
+`LockableObject`, which provides same synchronization primitives as Java's
+Object: `synchronize(&block)`, `wait(timeout = nil)`,
+`wait_until(timeout = nil, &condition)`, `signal`, `broadcast`. This class is
+intended for internal use in `concurrent-ruby` only and it does not support
+subclassing (since it cannot protect its lock from its children, for more
```
eregon: Is this checked? via a inherited/included hook throwing an exception?

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+to allow to call methods in the same object without double locking.
+
+``` ruby
+class Node
+  # ...
+  def left
+    synchronize { ns_left }
+  end  
+
+  def left
```
eregon: right + below

doc/synchronization.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+Three of them were used in the example above.
+
+> Current implementation relies on final instance variables where a instance of
+> `AtomicReference` is held to provide compare-and-set operations. That creates
+> extra indirection which is hoped to be removed over time when better
+> implementation will become available in Ruby implementations.
+
+## Concurrent Ruby Notes
```
eregon: This should be another document

doc/synchronization.md
eregon: Also it is not proved in the current JRuby implementation, is it?

doc/synchronization.md
jdantonio: :+1: I don't think that's pedantic at all. The Ruby community, in general, has a very poor understanding of concurrency. We should be very specific with our verbiage so that we don't create additional confusion.

doc/synchronization.md
jdantonio: :+1: That's why the README and other docs generally use words like "inspire" and "influence."

doc/synchronization.md
jdantonio: I don't think the quotes are necessary at all, the backticks provide enough highlighting.

doc/synchronization.md
pitr-ch: I am pointing to a piggybacking technique, this explains it better: http://stackoverflow.com/questions/8769570/volatile-piggyback-is-this-enough-for-visiblity.

doc/synchronization.md
pitr-ch: Bad wording. Is "Module or Class is defined atomically, then its methods are defined atomically one by one." better?

doc/synchronization.md
pitr-ch: `Thread#[]` or `Thread#thread_variable_get` are synchronised methods on JRuby.

IVars on JRuby are not fully correct at the moment, there is loadFence missing (considering only path using fences), but @headius created an issue to fix it AFAIK. So for this document I've considered them already volatile. The volatile ivar parts was agreed (best to my understanding) to be removed from https://github.com/jruby/jruby/wiki/Concurrency-in-jruby to avoid users getting used to it.

doc/synchronization.md
pitr-ch: Hm right you could be loading concurrently different constants, given constant can auto-loaded only once though, rest is blocking until the particular auto-load finishes. If you agree, I'll update this part.

doc/synchronization.md
pitr-ch: Interesting indeed, I would say totally messed up :)

doc/synchronization.md
pitr-ch: True, splitting.

doc/synchronization.md
pitr-ch: Its a TODO to do that.

pitr-ch:
Thanks for your comments, following commit will include only typo fixes, I'll comment on the more important questions after that.

doc/synchronization.md
pitr-ch: Good idea, added.

doc/synchronization.md
pitr-ch: I'll include explanation why there is a JRuby resource but not a one for MRI and Rubinius. Basically I could not find similar comprehensive page for MRI and Rubinius.

pitr-ch: Updated in d4dc311.

doc/synchronization.md
pitr-ch: Yeah I did not get this part quite right, improved in f0c4cce.

doc/synchronization.md
pitr-ch: Updated in b2ba23a.

doc/synchronization.md
pitr-ch: Updated in 8af0aa6.

doc/synchronization.md
pitr-ch: There were 2 things I wanted to discuss more, but you've already asked about both :) First is class variables. Since they are mostly deprecated I did not spend much time investigation, what the current behaviour is. AFAIK: MRI: volatile, JRuby: volatile, rbx: ?, J+T: non-volatile.

I am considering following:

- they are deprecated, not that important to be performant. -> volatile
- they are shared across child classes making them volatile could be quite expensive -> non-volatile
- if non-volatile, extension creating volatile method accessors can be added. -> non-volatile
- to best my knowledge they are usually used for a shared global flag where volatility is required. -> volatile
- there is no extension to make it volatile. -> volatile

I think the correct solution would be having them non-volatile. On the other hand it may not be worth it since they used rarely and deprecated. I would like to hear your opinions.

doc/synchronization.md
pitr-ch: This line is intended as a general explanation what is meant by definition of a variable and what does it imply when combined with variable write.

pitr-ch: Second thing I wanted to discuss more is: should local variables be serialised?

Local variables are determined during parsing, only problematic thing is `local_variable_set`. I thought I understand it but this example surprised me:
```ruby
[8] pry(main)> b=-> {a}
=> #<Proc:0x007fc68a76c430@(pry):8 (lambda)>
[9] pry(main)> b.binding.local_variable_set :a, 1
=> 1
[10] pry(main)> b.call
NameError: undefined local variable or method `a' for main:Object
```
This works though:
```
[11] pry(main)> a = 0
=> 0
[12] pry(main)> b=-> {a}
=> #<Proc:0x007fc68e922d70@(pry):12 (lambda)>
[13] pry(main)> b.binding.local_variable_set :a, 1
=> 1
[14] pry(main)> b.call
=> 1
```
So maybe a local variable cannot be really defined? More information here http://stackoverflow.com/questions/19321698/in-ruby-is-there-no-way-to-dynamically-define-a-local-variable-in-the-current-c .

chrisseaton: When line 8 is parsed, `a` isn't a local variable, so it's assumed to be a method call. The error message says 'undefined local variable or method `a'', but really I think the decision has already been taken that it is a method. When you add a local variable, it doesn't go back and reconsider what `a` could be.

doc/synchronization.md
```diff
@@ -0,0 +1,340 @@
...
+Key properties are:
+
+-   **volatility (V)** - Any written value is immediately visible to any
+    subsequent volatile read of the same variable. Any write executed before
+    volatile write based on program-order is visible to the read as well, which
+    allows
+    [piggybacking](http://stackoverflow.com/questions/8769570/volatile-piggyback-is-this-enough-for-visiblity).
+    (Same meaning as in Java, it creates synchronizes-with (JMM term) order
+    between write and read, which participates in creating happens-before
+    order.)
```
eregon: This is very complex now, I think piggybacking is also going too far from the definition.
Maybe we could just say that 'subsequent read' is either same thread, later in the program, or there is some synchronization between the write and the subsequent read.

doc/synchronization.md
eregon: One way to understand this is the binding has an additional frame of its own, so new variables can only be read from that binding. If the variable already exists in scope, it is however just set to the new value.

doc/synchronization.md
eregon: How is it implemented in JRuby? Similar to ivars?
It probably make sense to have them volatile as they are some sort of "namespaced" global state.

pitr-ch: Currently the JRuby implementation is volatile but it breaks the serialization, see https://github.com/jruby/jruby/issues/3357

doc/synchronization.md
pitr-ch: updated in 2b6a7bd

doc/synchronization.md
pitr-ch: Updated 661a5b6.

pitr-ch:
Thanks for your comments so far. It's getting quite complicated here to follow and to comment, therefore I'll move further discussion to Google docs. Two main remaining topics to solve is volatile class variables and serialised local variables.

pitr-ch:
The document was migrated to https://docs.google.com/document/d/1pVzU8w_QF44YzUCCab990Q_WZOdhpKolCIHaiXG-sPw/edit?usp=sharing Please use it instead, I hope it'll be more pleasant to comment on.

PR #423, opened 2015-09-25
Fix buggy tests

hfenton:
### Fixed a few buggy tests:

- Concurrent::Agent#restart when #failed? does not clear the action queue when :clear_actions is false
- Concurrent::Future#initialize uses the global io executor by default


jdantonio:
The test that failed on Travis is identical to one of the tests fixed in this PR. The test that failed on AppVeyor is unrelated to this PR. I've added that test to my list on issue #318.

PR #428, opened 2015-09-29
Fix Future#flat when failures happen

### LINKED ISSUE #426: Inconsistent `flat` behavior
Using concurrent-ruby-edge 0.2.0pre2 on jRuby 9.0.1.0.

The FlatPromise only resolves if the contained future is successful. If the inner future fails, or raises, then the flat never resolves and hangs the program if you `wait` or call `value`.

```ruby
# Works
foo = Concurrent.future { Concurrent.future { 7 } }.flat
foo.value! # 7

# Fails
bar = Concurrent.future { Concurrent.future { raise "no" } }.flat
bar.value! # Hangs

# OR

baz = Concurrent.future { Concurrent.failed_future("no") }.flat
baz.value! # Hangs
```

pitr-ch:
fix #426

### DIFF
--- a/spec/concurrent/edge/future_spec.rb
+++ b/spec/concurrent/edge/future_spec.rb
@@ -194,7 +194,7 @@ describe 'Concurrent::Edge futures' do
   describe 'Future' do
     it 'has sync and async callbacks' do
       callbacks_tester = ->(future) do
-        queue  = Queue.new
+        queue = Queue.new
         future.on_completion(:io) { |result| queue.push("async on_completion #{ result.inspect }") }
         future.on_completion! { |result| queue.push("sync on_completion #{ result.inspect }") }
         future.on_success(:io) { |value| queue.push("async on_success #{ value.inspect }") }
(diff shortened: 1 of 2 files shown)


spec/concurrent/edge/future_spec.rb
```diff
@@ -312,6 +312,16 @@
     it 'has flat map' do
       f = Concurrent.future { Concurrent.future { 1 } }.flat.then(&:succ)
       expect(f.value!).to eq 2
+
+      err = StandardError.new('boo')
+      f   = Concurrent.future { Concurrent.failed_future(err) }.flat
+      expect(f.reason).to eq err
+
+      f   = Concurrent.future { raise 'boo' }.flat
```
hobodave: I think this should be `Concurrent.future { Concurrent.future { raise 'boo' } }.flat` since `#flat` is expecting to be returned a future. (Or add this case as an additional test)

hobodave:
Thanks for this @pitr-ch! I'm testing it now in our real-world use case.

pitr-ch:
@hobodave great! Btw since this is an edge feature we are looking for a feedback about the API before it finalises and goes to concurrent core. If you have any suggestions, insights, please share.

spec/concurrent/edge/future_spec.rb
pitr-ch: yeah it should be a separate example, I was lazy.

hobodave:
What would be the best mechanism to share API feedback? Creating issues?

pitr-ch:
@hobodave yeah please create an issue.

jdantonio:
@pitr-ch Will you be merging this today? I'd like to publish the next pre-release today.

pitr-ch:
@jdantonio just waiting for @hobodave to confirm that it fixes the issues for him, after that it can be merged.

hobodave:
Confirmed :+1:

PR #430, opened 2015-09-29
Concurrent::Edge::AbstractPromise now rescues Exception

hobodave:
This fixes a bug which causes Futures to hang and never complete if the code within raises a subclass of `Exception`.

Example:

```ruby
module WebMock
  class NetConnectNotAllowedError < Exception; end
end

f = Concurrent.future { raise WebMock::NetConnectNotAllowedError, 'fail' }
f.value! # Hangs
```

See https://github.com/bblimke/webmock/pull/301 for WebMock's (valid) justification for subclassing `Exception` over `StandardError`.

### DIFF
--- a/spec/concurrent/edge/future_spec.rb
+++ b/spec/concurrent/edge/future_spec.rb
@@ -313,6 +313,14 @@ describe 'Concurrent::Edge futures' do
       f = Concurrent.future { Concurrent.future { 1 } }.flat.then(&:succ)
       expect(f.value!).to eq 2
     end
+
+    it 'completes future when Exception raised' do
+      f = Concurrent.future { raise Exception, 'fail' }
+      sleep 0.2
+      expect(f).to be_completed
+      expect(f).to be_failed
+      expect{ f.value! }.to raise_error(Exception, 'fail')
+    end
   end
 
   describe 'interoperability' do
--- a/lib/concurrent/edge/future.rb
+++ b/lib/concurrent/edge/future.rb
@@ -925,8 +925,11 @@ module Concurrent
       # @return [Future]
       def evaluate_to(*args, block)
         complete_with Future::Success.new(block.call(*args))
-      rescue => error
+      rescue StandardError => error
         complete_with Future::Failed.new(error)
+      rescue Exception => error
+        complete_with Future::Failed.new(error)
+        raise error
       end
     end
 


jdantonio:
@hobodave In the example you show, you could pass a timeout value to the `#value!` method and it would timeout as a result of the exception. Blocking a thread without timeout is generally not a great idea. Is there a specific reason you can't pass a timeout in this case?

@pitr-ch Thoughts? In general we capture `StandardError` and not `Exception`, though not always. My assumption has always been that `Exception` should be allowed to crash the application, as is normally the case. Many consider capturing `Exception` to be an anti-pattern. This is something I struggled with in the earliest days of this gem. This is why some classes take a `:rescue_exception` option (such as `SafeTaskExecutor`). I'm not necessarily opposed to capturing `Exception` but I think we should be consistent throughout the library and document this behavior.

hobodave:
@jdantonio Thanks for the tip on using a timeout on `#value!, but it doesn't help.

I think concurrent still needs to mark the future as completed when an Exception occurs. Take this example:

```ruby
f = Concurrent.future { raise Exception, 'fail' }
f.value!(5) # nil -- No exception thrown, no indication of why it failed, or anything
f.state # :pending
```

I understand the dangers of catching Exception, which is why my solution re-raises any Exception caught, after marking the future done. Also, raising an Exception in a Thread does not ever crash the application until you `Thread#join` that Thread, which can never happen if it's still pending.

Example:

```ruby
t = Thread.new { raise Exception, 'boom' } # Nothing happens until ...
t.join # Exception: boom
```

jdantonio:
@hobodave You make great points. You've convinced me. I'm going to leave the decision merge to @pitr-ch since he is leading the effort on the new future components, but I'm comfortable with this direction.

pitr-ch:
Thanks @hobodave! The Future must not hang because of Exception, that's for sure, but it cannot be re-raised either. I've cherrypicked your commit and I'll submit updated PR.

spec/concurrent/edge/future_spec.rb
```diff
@@ -313,6 +313,14 @@
       f = Concurrent.future { Concurrent.future { 1 } }.flat.then(&:succ)
       expect(f.value!).to eq 2
     end
+
+    it 'completes future when Exception raised' do
+      f = Concurrent.future { raise Exception, 'fail' }
+      sleep 0.2
```
pitr-ch: Please avoid sleep in tests, they cause timing issues and slow down tests. It can be replaced with `f.wait(1)`. (I'll fix it in follow up PR)

pitr-ch:
replaced by #431

pitr-ch:
Thanks again @hobodave, this is a good catch.

spec/concurrent/edge/future_spec.rb
hobodave: Thanks for the tip!

PR #431, opened 2015-09-30
rescue Exception in Futures

pitr-ch:
follow up of #430

PR #434, opened 2015-10-01
Improved java extension loading.

jdantonio:
Addresses the issue first noticed in #431.

The original implementation assumed that the Java extensions would always be compiled and loaded when running on JRuby. This is not always the case. During development and testing if the gem in included directly from source, as when the Gemfile references git directly, none of the native extensions will exist. In production if the pure-Ruby gem is loaded under JRuby the extensions won't exist, either.

This commit updates Java extension loading to follow the existing pattern for C extensions. Relevant parts of the code check to see if the extensions have been correctly loaded first, rather than assuming they are. The test suite still assumes that Java and C extensions are loaded, however. This is necessary to catch bugs which prevent extensions from properly compiling and loading.

### DIFF
--- a/lib/concurrent/atomic/atomic_boolean.rb
+++ b/lib/concurrent/atomic/atomic_boolean.rb
@@ -48,7 +48,7 @@ module Concurrent
   #   Explicitly sets the value to false.
   #
   #   @return [Boolean] true is value has changed, otherwise false
-  
+
   ###################################################################
 
   # @!macro [new] atomic_boolean_public_api
@@ -79,7 +79,7 @@ module Concurrent
   # @!visibility private
   # @!macro internal_implementation_note
   AtomicBooleanImplementation = case
-                                when Concurrent.on_jruby?
+                                when defined?(JavaAtomicBoolean)
                                   JavaAtomicBoolean
                                 when defined?(CAtomicBoolean)
                                   CAtomicBoolean
--- a/lib/concurrent/atomic/atomic_fixnum.rb
+++ b/lib/concurrent/atomic/atomic_fixnum.rb
@@ -96,7 +96,7 @@ module Concurrent
   # @!visibility private
   # @!macro internal_implementation_note
   AtomicFixnumImplementation = case
-                               when Concurrent.on_jruby?
+                               when defined?(JavaAtomicFixnum)
                                  JavaAtomicFixnum
                                when defined?(CAtomicFixnum)
                                  CAtomicFixnum
--- a/lib/concurrent/atomic/semaphore.rb
+++ b/lib/concurrent/atomic/semaphore.rb
@@ -91,7 +91,7 @@ module Concurrent
   # @!visibility private
   # @!macro internal_implementation_note
   SemaphoreImplementation = case
-                            when Concurrent.on_jruby?
+                            when defined?(JavaSemaphore)
                               JavaSemaphore
                             else
                               MutexSemaphore
--- a/lib/concurrent/map.rb
+++ b/lib/concurrent/map.rb
@@ -1,30 +1,31 @@
 require 'thread'
 require 'concurrent/constants'
+require 'concurrent/utility/native_extension_loader'
 
 module Concurrent
   # @!visibility private
   module Collection
 
     # @!visibility private
-    MapImplementation = if defined?(RUBY_ENGINE)
-                   case RUBY_ENGINE
-                   when 'jruby'
-                     # noinspection RubyResolve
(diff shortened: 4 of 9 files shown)


jdantonio:
@hobodave If you have a few moments, would you please point your development project to this branch and let me know if it resolves the error you experienced in #431? Thank you!

@pitr-ch Please see me comment [here](https://github.com/ruby-concurrency/concurrent-ruby/pull/431#issuecomment-144693575) then take a look at this PR. I think this is the direction we'll need to go if we want to enable users to reference the gem source within their Gemfile.

jdantonio:
Another option is to check the compiled Java code into git. This is only a partial solution, however. It solves the problem of referencing the repo directly from the Gemfile, but it doesn't solve the problem of installing the pure-Ruby gem on JRuby.

The `ref` gem, which was given to us by the original creator, checks the Java files into git. It also builds only one ".gem" file for all platforms and packages the compiled Java files in it. So users who are not on JRuby still have the compiled Java code even though it isn't loaded. This approach has brought a few complaints from DevOps people who view compiled code that isn't used to be a security risk.

At my last company produced packaged, enterprise software that was installed on-premise. We had to package all gem dependencies in the installer (our product has C++, C#, Erlang, Ruby, and JavaScript--it was a beast). We could only package pure-Ruby gems, even if gems with extensions were available. Many of our customers ran security scans against our software before installation. We ran into a problem with Rake once. The Rake gem includes a couple of uncompiled C files (a "hello world" program) in the docs folder as documentation for using Rake as a build tool. Those files triggered a security scanner once. I had to make a private fork of Rake and create a custom gem file that excluded those C files. It was a huge PITA. Anyway, because of experiences like that I would prefer that our "pure Ruby" gem truly be pure Ruby and that we don't package the compiled Java files into it.

mikeknep:
@jdantonio I've been working with @hobodave, who will be out the next few days—not sure if he'll see this until Tuesday. But FYI, I had been having the same issue on my machine, and pointing at this branch (technically at `ref: '7a6cb23fad620ccf74079c51780c62a6ee4a8372'`) resolves the error for me. I get the following messages before the specs start executing (no big deal, and the first one is from JRuby).

```
$ bundle exec rspec
uri:classloader:/jruby/kernel/kernel.rb:17: warning: unsupported exec option: close_others
On JRuby but Java extensions failed to load.
Possibly unsupported Ruby implementation
Possibly unsupported Ruby implementation

.........(etc.)
```

Thanks for being so responsive—it's really helping us out!

jdantonio:
@mikeknep Thanks for the info! I don't know what that "unsupported exec option" message is (I'll look into it) but the other warnings are expected. They are the checks I added in this PR. I'll update the verbiage in the last two to include the class name. I'm not going to merge this until @pitr-ch takes a look since he's the implementer of the synchronization layer, but we should be this in master fairly quickly.

mikeknep:
Very cool. The unsupported exec option is a [known issue](https://github.com/jruby/jruby/issues/1913) with JRuby and Bundler—no worries on your end!

jdantonio:
@hobodave @mikeknep This is part of the 1.0.0.pre4 build which I just released.

pitr-ch:
Sorry for my very delayed reply (moving sucked up all my free time).

Problem is that we cannot go pure Ruby on JRuby. We need the extensions in synchronisation layer to guarantee safety on JVM platform. I think we should fail if the compiled code is not available. If compilation is not possible (when referring github) we could create another companion gem (with same versioning) which would carry precompiled code to solve this.

jdantonio:
I am perfectly comfortable raising an exception in the pure-Ruby gem is installed on JRuby since we provide a gem specifically for JRuby. So I see this as a development issue only. If we check the compiled Java code into git it should solve the issue of referencing the repo from the Gemfile. So long as the pure-Ruby build excludes the compiled code we're fine.

This is a small update. I'll work on it tomorrow.

pitr-ch:
@jdantonio Thanks.

jdantonio:
@pitr-ch It looks like you updated the extension loader in #421 but did not, on that PR, check the Java extensions into git. Is that correct?

pitr-ch:
yeah that part is still missing. I needed the error in place to be sure that I run correct code, since we've discussed it I've committed it already. Would you add the Java extension?

jdantonio:
That can wait until after the 1.0 release, I think. It's an edge case that only occurs during development and only when pulling directly from the repo. We can hold off and revisit if it ever becomes a problem again.

pitr-ch:
:+1:

PR #436, opened 2015-10-02
clean up a few warnings

tenderlove:
The bare `*` is ambiguous, so I added parens to make it unambiguous.
Also I removed some unused assignments.

I don't know how to turn on warnings for RSpec, so I just ran the tests
like this:

```
RUBYOPT=-w be rake spec
```

### DIFF
--- a/lib/concurrent/channel/buffer/timer.rb
+++ b/lib/concurrent/channel/buffer/timer.rb
@@ -62,7 +62,6 @@ module Concurrent
 
             if Concurrent.monotonic_time > @tick
               # only one listener gets notified
-              closed = empty = true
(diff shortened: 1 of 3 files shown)


jdantonio:
Thank you! I couldn't figure out another way, either, so I added `RUBYOPT=-w` to our Travis config, I just keep forgetting to check (or run it that way at home).

While we're chatting, the other day we started seeing a segfault from `ruby 2.3.0dev (2015-10-02 trunk 52005)` on Travis. It consistently happens in the same test, same line of code. Example is [here](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/83354700). Should I submit a bug report? Or just let it be?

PR #441, opened 2015-10-09
Adds a JRuby optimized version of Concurrent::Atom

lucasallan:
Non-optimized version
```
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   4.720000   0.030000   4.750000 (  4.103692)
After JVM warm-up...
   3.920000   0.020000   3.940000 (  3.925786)
```
Optimized version
```
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   2.980000   0.020000   3.000000 (  2.742403)
After JVM warm-up...
   2.440000   0.010000   2.450000 (  2.479192)
```

Benchmark script:
```ruby
#!/usr/bin/env ruby

require 'concurrent'
require 'benchmark'

NUM      = 50_000_000
WARM_UPS = NUM

def jruby?
  defined? JRUBY_VERSION
end

def version
  gem_dir = Gem::Specification.find_by_name('concurrent-ruby').gem_dir
  File.open(File.join(gem_dir, 'VERSION'), 'r').read
rescue
  'unknown'
end

puts "~~~ Ruby version: #{RUBY_VERSION}"
if jruby?
  puts "~~~ JRuby version: #{JRUBY_VERSION}"
else
  puts "~~~ Ruby engine: #{RUBY_ENGINE}"
end
puts "~~~ Gem version: #{version}"

Benchmark.bm do |stats|

  puts "Benchmarking #{Concurrent::Atom}..."

  ref   = Concurrent::Atom.new('foo')
  value = nil

  puts 'Before JVM warm-up...' if jruby?
  stats.report do
    NUM.times { value = ref.value }
  end

  if jruby?
    WARM_UPS.times { value = ref.value }

    puts 'After JVM warm-up...'
    stats.report do
      NUM.times { value = ref.value }
    end
  end
end
```

### DIFF
--- a/lib/concurrent/atom.rb
+++ b/lib/concurrent/atom.rb
@@ -1,7 +1,6 @@
 require 'concurrent/atomic/atomic_reference'
 require 'concurrent/collection/copy_on_notify_observer_set'
 require 'concurrent/concern/observable'
-require 'concurrent/synchronization'
 
 # @!macro [new] thread_safe_variable_comparison
 #
@@ -91,12 +90,8 @@ module Concurrent
   #
   # @see http://clojure.org/atoms Clojure Atoms
   # @see http://clojure.org/state Values and Change - Clojure's approach to Identity and State
-  class Atom < Synchronization::Object
+  class Atom
     include Concern::Observable
-
-    safe_initialization!
-    private(*attr_volatile_with_cas(:value))
-    public :value
 
     # Create a new atom with the given initial value.
     #
@@ -111,15 +106,18 @@ module Concurrent
     #
     # @raise [ArgumentError] if the validator is not a `Proc` (when given)
     def initialize(value, opts = {})
-      @Validator     = opts.fetch(:validator, -> v { true })
+      @validator     = opts.fetch(:validator, -> v { true })
       self.observers = Collection::CopyOnNotifyObserverSet.new
-      super(value)
+      @state = AtomicReference.new(value)
     end
 
     # @!method value
     #   The current value of the atom.
     #
     #   @return [Object] The current value.
+    def value
+      @state.get
+    end
 
     alias_method :deref, :value
 
@@ -160,7 +158,7 @@ module Concurrent
         begin
           new_value = yield(old_value, *args)
           break old_value unless valid?(new_value)
-          break new_value if compare_and_set(old_value, new_value)
+          break new_value if @state.compare_and_set(old_value, new_value)
         rescue
           break old_value
         end
@@ -177,7 +175,7 @@ module Concurrent
     #
     # @return [Boolean] True if the value is changed else false.
     def compare_and_set(old_value, new_value)
-      if valid?(new_value) && compare_and_set_value(old_value, new_value)
+      if valid?(new_value) && @state.compare_and_set(old_value, new_value)
         observers.notify_observers(Time.now, old_value, new_value)
         true
       else
@@ -196,7 +194,7 @@ module Concurrent
     def reset(new_value)
       old_value = value
       if valid?(new_value)
-        self.value = new_value
+        @state.set(new_value)
         observers.notify_observers(Time.now, old_value, new_value)
         new_value
       else
@@ -212,7 +210,7 @@ module Concurrent
     # @return [Boolean] false if the validator function returns false or raises
     #   an exception else true
     def valid?(new_value)
-      @Validator.call(new_value)
+      @validator.call(new_value)
     rescue
       false
     end


lib/concurrent/jruby_atom.rb
```diff
@@ -0,0 +1,223 @@
+if Concurrent.on_jruby?
+
+  require 'concurrent/collection/copy_on_notify_observer_set'
+  require 'concurrent/concern/observable'
+  require 'java'
+
+  java_import 'java.util.concurrent.atomic.AtomicReference'
```
ianks: Out of curiosity, how big is the perf hit of using Concurrent::Reference versus the Java one?

jdantonio:
@lucasallan Thank you very much for working on this. I'm having difficulty following the differences between the implementations. AFAICT the main difference is that the newer version simply uses `java.util.concurrent.atomic.AtomicReference` directly rather than `Concurrent::AtomicReference`. Is that correct?

lucasallan:
@jdantonio This version is directly based on Clojure code[1], but I guess we should be using `Concurrent::AtomicReference` rather than `java.util.concurrent.atomic.AtomicReference`.
But before it wasn't using `AtomicReference` anyway, right?

[1] - https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Atom.java

jdantonio:
I didn't look at the Clojure code before implementing this or Agent. I worked entirely from the documented behavior. I probably should have looked at the Clojure source. :-) If the implementation can be improved following that example I'm all for it, but can we improve it for the base implementation? Do we need a JRuby-specific version?

lucasallan:
If we build an implementation around `AtomicReference` then we don't need a JRuby optimized version, since `AtomicReference` is already jruby-optimized.

jdantonio:
@lucasallan Thank you!

lucasallan:
@jdantonio Results - I executed it 4 times: 

### Using `Concurrent::Atomic` 
```
➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   4.190000   0.040000   4.230000 (  3.116446)
After JVM warm-up...
   2.840000   0.010000   2.850000 (  2.848073)


➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   2.860000   0.020000   2.880000 (  2.546445)
After JVM warm-up...
   2.400000   0.010000   2.410000 (  2.395978)



➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   2.620000   0.020000   2.640000 (  2.438905)
After JVM warm-up...
   2.500000   0.010000   2.510000 (  2.504156)



➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   2.740000   0.010000   2.750000 (  2.386496)
After JVM warm-up...
   2.390000   0.010000   2.400000 (  2.376449)

```

### Using concurrent-ruby-0.9.1(Current implementation):
```
➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   4.690000   0.040000   4.730000 (  4.067498)
After JVM warm-up...
   3.690000   0.010000   3.700000 (  3.705187)


➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   5.000000   0.030000   5.030000 (  4.547322)
After JVM warm-up...
   4.200000   0.010000   4.210000 (  4.217589)


➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   4.710000   0.020000   4.730000 (  4.314722)
After JVM warm-up...
   4.290000   0.010000   4.300000 (  4.258959)


➜  concurrent-ruby git:(master) ✗ ruby ../atom_bench.rb
~~~ Ruby version: 2.2.2
~~~ JRuby version: 9.0.1.0
~~~ Gem version: unknown
Benchmarking Concurrent::Atom...
Before JVM warm-up...
   4.320000   0.020000   4.340000 (  3.944181)
After JVM warm-up...
   3.820000   0.020000   3.840000 (  3.842717)
```

jdantonio:
Thank you!

pitr-ch:
@lucasallan I very appreciate your contributions but this particular change is unsafe. The @state instance variable is not properly initialized. Generally speaking I think all of our abstractions should use synchronisation layer to avoid this type of problems. It also helps greatly to solve issues across the gem if one is found. I know the benchmark shows improvement but when the number of iterations is taken into-account it's not big gain. It's also best to avoid optimisations of a method inlining kind, that's a job for the Ruby interpreter. We should prefer code clarity and algorithmic optimisations. I would suggest to revert this change.

lucasallan:
@pitr-ch Could you clarify what you meant by `is not properly initialized`? Also, we're using `Concurrent:: AtomicReference` which already give us some benefits + performance, I don't understand why it would be a bad idea.

pitr-ch:
Compiler or processor can reorder writes, in particular it may choose to reorder write to `@state` variable with write of the newly created `Atom` object to a variable accessible by other threads. If it does than another thread may observe `Atom` object with @state variable still uninitialized which would lead to `#value` method failing on nil error.

There is nothing bad about `Concurrent::AtomicReference`, in fact `attr_volatile_with_cas` uses it to implement CAS operations over the state field. The problematic part is that this is already abstracted pattern and it's expected to be different in future based on Ruby implementation. Some Ruby implementations may implement this directly (CAS operations over instance variables) which will allow to get rid of the indirection caused by `AtomicReference`. When this happens this implementation would not get the benefit of the abstraction provided by the layer, but it would have to be modified again to use the layer.

lucasallan:
Fair enough. Thanks for the awesome explanation @pitr-ch - I'll revert it later today.

pitr-ch:
Thanks @lucasallan.

PR #449, opened 2015-10-15
MVar#take accepts a block as parameter

lionelperrin:
When using MVar, one should very often do the following:
 1. acquire the value
 1. perform an action on it
 1. release it.

It seems to me that a block syntax is suitable in this case:
```ruby
f = MVar.new(IO.open('file', 'w'))
new_lines = f.take { |s| s.read }
```
instead of
```ruby
f = MVar.new(IO.open('file', 'w'))
new_lines = nil
begin
  s = f.take
  new_lines = s.read
ensure
  f.put s
end
```
or
```ruby
f = MVar.new(IO.open('file', 'w'))
new_lines = nil
f.modify do |s|
  new_lines = s.read
  s
end
```

### DIFF
--- a/lib/concurrent/mvar.rb
+++ b/lib/concurrent/mvar.rb
@@ -68,10 +68,14 @@ module Concurrent
 
         # If we timed out we'll still be empty
         if unlocked_full?
-          value = @value
-          @value = EMPTY
-          @empty_condition.signal
-          apply_deref_options(value)
+          if block_given?
+            yield @value
+          else
+            value = @value
+            @value = EMPTY
+            @empty_condition.signal
+            apply_deref_options(value)
+          end
         else
           TIMEOUT
         end
--- a/lib/concurrent/mvar.rb
+++ b/lib/concurrent/mvar.rb
@@ -68,14 +68,27 @@ module Concurrent
 
         # If we timed out we'll still be empty
         if unlocked_full?
-          if block_given?
-            yield @value
-          else
-            value = @value
-            @value = EMPTY
-            @empty_condition.signal
-            apply_deref_options(value)
-          end
+          value = @value
+          @value = EMPTY
+          @empty_condition.signal
+          apply_deref_options(value)
+        else
+          TIMEOUT
+        end
+      end
+    end
+
+    # acquires lock on the from an `MVAR`, yields the value to provided block,
+    # and release lock. A timeout can be set to limit the time spent blocked,
+    # in which case it returns `TIMEOUT` if the time is exceeded.
+    # @return [Object] the value returned by the block, or `TIMEOUT`
+    def borrow(timeout = nil)
+      @mutex.synchronize do
+        wait_for_full(timeout)
+
+        # if we timeoud out we'll still be empty
+        if unlocked_full?
+          yield @value
         else
           TIMEOUT
         end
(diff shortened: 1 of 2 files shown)


spec/concurrent/mvar_spec.rb
```diff
@@ -68,6 +68,28 @@ def dereferenceable_subject(value, opts = {})
         expect(m.take(0.1)).to eq MVar::TIMEOUT
       end
 
+      context 'when a block is given' do
+        it 'yields current value to the block and puts back value' do
```
ianks: Nit: follow same space styling as the tests in the file.

lib/concurrent/mvar.rb
```diff
@@ -68,10 +68,14 @@ def take(timeout = nil)
 
         # If we timed out we'll still be empty
         if unlocked_full?
-          value = @value
-          @value = EMPTY
-          @empty_condition.signal
-          apply_deref_options(value)
+          if block_given?
+            yield @value
```
ianks: This doesn't seems to make sense to me. Why would take remove the value in this case?

lionelperrin: Sorry for the space styling. I'll change this.

I'm not sure if I understand correctly your question. 

I've tried to reproduce the behavior of IO#open for instance:
 * When no block is given, it returns a file.
 * When a block is given, the file is given to this block. 

In the MVar#take case : 
 * when no block is given: it returns the value and remove it from the MVar (I haven't changed this behavior).
 * When a block is given, the value is given to the block but not removed from the MVar.

I'm wondering if apply_deref_options should be called when passing the value to the block. I've chosen not to call it, like it is done in MVar#modify.

ianks: Why would the item not be removed from the MVar when a block is given? That seems semantically inconsistent.

lionelperrin: My objective is to be able to write code like this:
```ruby
f = MVar.new(IO.open('file', 'w'))
new_lines = f.take { |s| s.read }
new_lines = f.take { |s| s.read }
```
When using the block syntax, the user should not have to worry about putting back the value in the mvar.

jdantonio:
@lionelperrin I really like the idea behind this PR. I 100% agree with you that Ruby's block syntax is a perfect fit for the operation you want to perform. I wonder if `#take` is the correct place for it, however. Semantically, a "take" operation implies that the value should be removed. I agree with you that in your code example the user shouldn't be required to put the value back. The user isn't *taking* the value, the user is performing an operation *with* the value. What if we reinforce that in the method name?

```ruby
f = MVar.new(IO.open('file', 'w'))
new_lines = f.with { |s| s.read } # no timeout
new_lines = f.with(1) { |s| s.read } # 1 second timeout
```

Thoughts?

ianks:
@jdantonio What about something like this:

```ruby
f = MVar.new(IO.open('file', 'w'))
new_lines = f.borrow { |s| s.read } # no timeout
new_lines = f.borrow(timeout: 1) { |s| s.read } # 1 second timeout
```

lionelperrin:
Thanks for your feedback,

It seems to me that the pro for #take are:
 * it is consistent with other standard methods such as #open which has two different behaviors when a block is given or not.
 * it is simpler: only one method in the documentation. It makes easier to understand the goal of the MVar class. I have the feeling that it is important that a user can easily identify the main methods: #take and #remove. Providing syntactic sugar with blocks should not make this identification harder.

cons are:
 * the name #take is not well chosen since the value is put back/released when the block ends.
 * #borrows looks a good candidate.

I have updated the pull request with the #borrow version. I let you decide which of the commit (before or after renaming #take to #borrow) should be integrated.

Regards,

jdantonio:
Thank you!

PR #451, opened 2015-10-18
Add ready-length attr_reader

brandonhilkert:
The ready queue is available via a local instance variable. Exposing
the length of the array of ready workers will allow the health of the
pool to be measured/monitored as tasks are moving through the pool.

Important values are total workers (`length`), idle workers
(`ready_length`), and busy workers (difference between the 2).

Note: I wasn't sure how to handle the java side. It appears there are some [instance methods](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/executor/java_thread_pool_executor.rb#L52) pertaining to accessing this data, but I couldn't find more. I'm happy to make suggested changes if interested.

### DIFF
--- a/lib/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent/executor/ruby_thread_pool_executor.rb
@@ -70,6 +70,11 @@ module Concurrent
       synchronize { @queue.length }
     end
 
+    # @!macro thread_pool_executor_attr_reader_ready_length
+    def ready_length
+      synchronize { @ready.length }
+    end
+
     # @!macro thread_pool_executor_attr_reader_remaining_capacity
     def remaining_capacity
       synchronize do
--- a/spec/concurrent/executor/thread_pool_executor_shared.rb
+++ b/spec/concurrent/executor/thread_pool_executor_shared.rb
@@ -162,6 +162,37 @@ shared_examples :thread_pool_executor do
     end
   end
 
+  context '#ready_length' do
+    subject do
+      described_class.new(
+        min_threads: 2,
+        max_threads: 5,
+        max_queue: 0,
+        fallback_policy: :discard
+      )
+    end
+
+    it 'returns 0 prior to enqueueing any tasks' do
+      expect(subject.ready_length).to eq 0
+    end
+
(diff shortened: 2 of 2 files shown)


jdantonio:
I appreciate what this PR intends but I'm not sure it will work. We need parity with JRuby and Java's TPE doesn't expose this information. It's an internal implementation detail.

In a perfect world, size of the ready pool should always be near zero. If the size of the pool is capped then new jobs should go to the queue. If the size of the pool is unbound, then new jobs should be immediately post to a thread (new or existing). Idle threads should be pruned. Generally speaking, the only time we should have waiting threads is when the number of threads in the pool is fixed, and the better metric in that case is the queue size.

I'll take a look at Java's TPE more closely, but I don't think it provides anything similar to this.

brandonhilkert:
Makes sense that it should generally be nothing. I'm trying to get up-to-date stats on the system for the web UI. This allows a user to see exactly what's going on. I'd like to be able to provide visibility to the workers, in addition to the number of jobs processed. Is there any other way to get that info?

fe0a0932-75dc-11e5-8260-788b2a98e4c0

jdantonio:
Not really. Workers are an internal detail, encapsulated within the pool. The core abstractions are the pool and the task/job. It would be very easy to provide more insight into running jobs, however. It would be fairly easy to increment a counter at the start of a task and decrement it after the tasks is complete, and then know how many jobs are running at any given time, for example. Something like that could be done at the application level, too, by putting that code in the block that's passed to the thread pool. I haven't had a chance to look at your sucker_punch working branch yet, but I'm thinking something like this:

```ruby
@running_task_count = Concurrent::AtomicFixnum.new # defaults to zero

def do_job
  @running_task_count.increment
  SuckerPunchJobClass.new.perform
rescue => ex
  # call error handler that triggers AirBrake, etc.
ensure
  @running_task_count.decrement
end

@thread_pool.post { do_job }
```

Would something like this be valuable?

brandonhilkert:
That's fair. I think that approach could work. I haven't decorated the Job classes at all, but it seems like it might be inevitable given what I'd like to accomplish.

PR #452, opened 2015-10-19
Move completed_task_count incrementer after task is successful

brandonhilkert:
Completed task count was previously incremented no matter the success of
the task. In hopes to introduce a failed_task_count counter, this moves
the increment of the completed_task_count to the block where the thread
execution takes place

### DIFF
--- a/lib/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent/executor/ruby_thread_pool_executor.rb
@@ -103,7 +103,7 @@ module Concurrent
 
     # @!visibility private
     def worker_task_completed
-      @completed_task_count += 1
+      synchronize { @completed_task_count += 1 }
     end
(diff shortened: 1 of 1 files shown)


lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -101,6 +101,11 @@ def worker_died(worker)
       synchronize { ns_worker_died worker }
     end
 
+    # @!visibility private
+    def worker_task_completed
+      @completed_task_count += 1
```
jdantonio: This callback needs synchronized. Can you please look to `#worker_died` as an example?

brandonhilkert:
@jdantonio Updated.

jdantonio:
Thank you!!!

PR #453, opened 2015-10-19
Keep track of failed task count

brandonhilkert:
In hopes to provide a more metrics around Sucker Punch jobs, I'm hoping
to expose the number of failed jobs in the pool.

Failed jobs would be visible in the chart on the left and potentially in the chart if I can find a way to not make it too busy:

screen shot 2015-10-18 at 9 11 46 pm

### DIFF
--- a/lib/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent/executor/ruby_thread_pool_executor.rb
@@ -108,7 +108,7 @@ module Concurrent
 
     # @!visibility private
     def worker_task_failed
-      @failed_task_count += 1
+      synchronize { @failed_task_count += 1 }
     end
 
     private
--- a/lib/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent/executor/ruby_thread_pool_executor.rb
@@ -55,6 +55,11 @@ module Concurrent
       synchronize { @completed_task_count }
     end
 
+    # @!macro thread_pool_executor_attr_reader_failed_task_count
+    def failed_task_count
+      synchronize { @failed_task_count }
+    end
+
     # @!macro executor_service_method_can_overflow_question
     def can_overflow?
       synchronize { ns_limited_queue? }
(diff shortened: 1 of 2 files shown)


lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -101,6 +106,11 @@ def worker_died(worker)
       synchronize { ns_worker_died worker }
     end
 
+    # @!visibility private
+    def worker_task_failed
+      @failed_task_count += 1
```
jdantonio: This needs synchronized as well.

jdantonio:
I really like the intent of this PR, but we need to look at the JRuby side. We have a strict rule that we don't ever build platform-specific features, only platform-specific optimizations. Since Java's [ThreadPoolExecutor](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html) doesn't expose this information we will need to track it in Ruby. I think that's workable.

`JavaExecutorService` defines an internal, private class called `Job` that calls the task block. If we added a `rescue` in there we could increment a counter. The (small) complexity is that we want the counter in `JavaThreadPoolExecutor` only, not `JavaExecutorService` (for compatibility with the Ruby implementation). So would could catch the error in `Job`, make a callback to the parent class, then re-raise the error. The default callback in `JavaExecutorService` would be a noop, which we could override in `JavaThreadPoolExecutor`. The latter would use an `AtomicFixnum` for the counter to make it thread safe.

Thoughts?

brandonhilkert:
@jdantonio I'm not opposed to exploring, but based on your previous suggestion for `@ready_length`, I [implemented something quick](https://github.com/brandonhilkert/sucker_punch/blob/concurrent-web/lib/sucker_punch/job.rb#L37) and it's working well. I'm not opposed to doing the same for failed count in a `rescue` block. You tell if this has any value to the project because it would be just as easy for me to keep it in SP.

jdantonio:
Cool. Since this feature would be a little messy on JRuby I'd prefer to hold off for now and leave it to the app to track. Thread pools are intended to be pretty low level and fast. Thanks for the input. I'm really excited to see how Sucker Punch evolves!

PR #457, opened 2015-10-30
Put debug log statements inside block

soupmatt:
This prevents potentially expense calls to #inspect and avoids creating a lot of useless string for applications that send a lot of messages to actors.

When I did some memory profiling, this change halved the number of objects that were getting allocated by my application. (The call to inspect was very expensive for me)

Also of note. I tried to add the same behavior to 2 log debug statements in the behavior classes, but they both caused test failures. I only tried this under MRI 2.2.3.

### DIFF
--- a/lib/concurrent/actor/core.rb
+++ b/lib/concurrent/actor/core.rb
@@ -91,7 +91,7 @@ module Concurrent
       # @param [Envelope] envelope
       def on_envelope(envelope)
         schedule_execution do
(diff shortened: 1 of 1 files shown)


soupmatt:
Also, if you guys would be willing to backport this to the 0.9 branch for 0.9.2, I would be extremely grateful.

jdantonio:
If @pitr-ch is comfortable with the change I'll backport it to 0.9.x

pitr-ch:
@soupmatt Thanks! this is a great catch, it should have been in a block from the beginning. Superseded by https://github.com/ruby-concurrency/concurrent-ruby/pull/458.

PR #461, opened 2015-11-03
Move AtomicMarkableReference out of Edge

ianks:
It would be nice to get this in at some point .Originally, it was planned for 1.0 but that is fine if it is not possible. Let me know your thoughts.

### DIFF
--- a/lib/concurrent/atomic/atomic_markable_reference.rb
+++ b/lib/concurrent/atomic/atomic_markable_reference.rb
@@ -1,6 +1,5 @@
 module Concurrent
-  module Edge
-
+  module Atomic
     # @!macro [attach] atomic_markable_reference
     #
     #   An atomic reference which maintains an object reference along with a mark bit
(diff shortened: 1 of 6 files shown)


lib/concurrent/atomic/atomic_markable_reference.rb
```diff
@@ -1,6 +1,5 @@
 module Concurrent
-  module Edge
-
+  module Atomic
```
jdantonio: There is no `Atomic` module. We are non-idiomatic in the way we handle some subdirectories and namespaces. We are going to fix this in 2.0. For now please leave `atomic_markable_reference.rb` in the `lib/concurrent/atomic` directory but make the namespace `Concurrent:: AtomicMarkableReference`.

jdantonio:
After 1.0 is released we'll make a concerted effort to move more things from edge to core. For the 1.0 release (before the end of the month) this will need to stay in edge. We can keep this PR open for now but not merge it.

pitr-ch:
replaced by #737

Issue #459, opened 2015-11-02
Migrate all abstractions to Synchronisation layer or to edge


pitr-ch:
Classes using mutex:
- [ ] MVar
- [ ] TVar
- [ ] Dereferencable (when is it suppose to be removed?)

Classes with final ivars not using layer: 
- [ ] Maybe
- [ ] Tick
- [ ] possibly others

jdantonio:
Removing Dereferencable is a breaking API change. I think 2.0 is the appropriate time.

I've tried to update MVar to use the synchronization layer but the tests fail every time. I haven't explored it deeply.

Anything related to channels (like Tick) can wait. Those are Edge and and need a lot more work before I would recommend using them.

PR #462, opened 2015-11-03
Sync updates

jdantonio:
See #459 

* Dereferenceable
* Maybe
* Tick
* TVar (`ensure_ivar_visibility!` only)
* MVar (`ensure_ivar_visibility!` only)

### DIFF
--- a/examples/go-by-example-channels/non-blocking-channel-operations.rb
+++ b/examples/go-by-example-channels/non-blocking-channel-operations.rb
@@ -15,9 +15,9 @@ Channel.select do |s|
   s.default { print "no message received\n" }
 end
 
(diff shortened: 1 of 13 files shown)


pitr-ch:
Ah sorry I should have communicated it better, I had to remove `#ensure_ivar_visibility!` in favor of of `.safe_initialization!`. Subclassing with `ensure_ivar_visibility!` was leading to multiple barriers which is undesirable. See https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/synchronization/object.rb#L48 I've opened #464

pitr-ch:
:+1:

PR #463, opened 2015-11-03
Fixed bug with requires for Map.

jdantonio:
PR #421 changed the way we load native extensions. Specifically, requiring the synchronization layer first. `lib/concurrent/map.rb` loaded things in the wrong order. This PR fixes that.


jdantonio:
@pitr-ch I discovered this while working on a [load time improvement](https://github.com/jdantonio/rails/commit/218aa0f56f16ca7f17451c3978e115619022591d) for Rails.

pitr-ch:
Thanks.

PR #469, opened 2015-11-22
add timeoutable method for run futures with timeout

fntz:
Hi,
I add a new method for work with `future` and timeout.

### DIFF
--- a/concurrent-ruby-edge.gemspec
+++ b/concurrent-ruby-edge.gemspec
@@ -29,4 +29,5 @@ Please see http://concurrent-ruby.com for more information.
   s.required_ruby_version = '>= 1.9.3'
 
   s.add_runtime_dependency 'concurrent-ruby', "~> #{Concurrent::VERSION}"
+  s.add_dependency 'timeout', "~> 0.0.1"
 end
--- a/examples/edge_futures.in.rb
+++ b/examples/edge_futures.in.rb
@@ -41,6 +41,12 @@ failing_zip.then { |v| 'never happens' }.result
 failing_zip.rescue { |a, b| (a || b).message }.value
 failing_zip.chain { |success, values, reasons| [success, values.compact, reasons.compactß] }.value
 
+### Timeout
+
+future = Concurrent.timeoutable(1) { sleep(2) } # Raise Timeout::Error after 1 second
+future = Concurrent.timeoutable(1) { :success } # Ok, return => :success
+
+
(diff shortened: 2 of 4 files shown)


jdantonio:
@fntz Thank you very much for taking the time to work on this. I greatly appreciate your effort. Unfortunately I cannot accept this PR. It violates a couple strict rules for this gem. I apologize that these rules aren't documented well. The next time I work on documentation I will make these rules clear. The two issues are:

1. **No downstream dependencies.** This gem is intended to be a core library for a wide range of frameworks and applications. Downstream dependencies limit our ability to control the effect we have on the projects that use us. One of the reasons that projects like Rails and Sidekiq have chosen to use concurrent-ruby is my commitment to this ideal.
2. **Never monkey-patch Ruby.** Basically for the reasons stated above. The `timeout` gem is pretty abusive to both `Thread` and `Kernel`. Monkey-patching like that wouldn't just affect the code within this gem, it would transitively affect the application which uses c-r and *every* one of its downstream dependencies. This would pose significant risk for our users.

Those issues aside, I am fairly opposed to adding a `timeout` method to c-r. Timeout is very complex, problematic, and is best handled by the application itself. This has been [discussed here before](https://github.com/ruby-concurrency/concurrent-ruby/pull/285#commitcomment-11217248). In fact, I *removed* an earlier `timeout` implementation for these reasons (I should blog about this sometime). I recommend reading Mike Perham's ["Timeout: Ruby's Most Dangerous API"](http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/) and Charlie Nutter's ["Ruby's Thread#raise, Thread#kill, timeout.rb, and net/protocol.rb libraries are broken"](http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html) for more context.

fntz:
ok, thanks for reply.

pitr-ch:
@fntz I am curious, what lead you to add this API, why was not `Future#value(timeout)` not sufficient ?

fntz:
@pitr-ch i think `Future#value` non async and block current thread (from my view of course and java experience :)).

pitr-ch:
So you wanted to be able to build a job which runs at most `x` seconds in background. For that the job has to be safely cancel-able which we plan to tackle in next major release.

fntz:
ok, this fine for me. many thanks.

PR #475, opened 2015-11-27
Fix typo in spec_helper.rb

eregon:
Alternatively, maybe the whole condition can be removed since 9.0.0.0 is not so recent anymore?


jdantonio:
Thank you!

PR #477, opened 2015-12-03
Raise better error when input is not an IVar

schneems:
This will show the values received such as:

```
rgumentError: Not all dependencies are IVars.
Dependencies: [false]
/Users/richardschneeman/.gem/ruby/2.3.0/gems/concurrent-ruby-1.0.0/lib/concurrent/dataflow.rb:61:in `call_dataflow'
/Users/richardschneeman/.gem/ruby/2.3.0/gems/concurrent-ruby-1.0.0/lib/concurrent/dataflow.rb:40:in `dataflow_with'
/Users/richardschneeman/.gem/ruby/2.3.0/gems/concurrent-ruby-1.0.0/lib/concurrent/dataflow.rb:35:in `dataflow'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/manifest.rb:203:in `block in compile'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/manifest.rb:131:in `block (2 levels) in find'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/base.rb:76:in `find_all_linked_assets'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/manifest.rb:130:in `block in find'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/manifest.rb:129:in `each'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/manifest.rb:129:in `find'
/Users/richardschneeman/Documents/projects/sprockets/lib/sprockets/manifest.rb:164:in `compile'
/Users/richardschneeman/.gem/ruby/2.3.0/bundler/gems/sprockets-rails-ad4a43bd1bb1/lib/sprockets/rails/task.rb:68:in `block (3 levels) in define'
/Users/richardschneeman/Documents/projects/sprockets/lib/rake/sprocketstask.rb:147:in `with_logger'
/Users/richardschneeman/.gem/ruby/2.3.0/bundler/gems/sprockets-rails-ad4a43bd1bb1/lib/sprockets/rails/task.rb:67:in `block (2 levels) in define'
```

### DIFF
--- a/lib/concurrent/dataflow.rb
+++ b/lib/concurrent/dataflow.rb
@@ -39,7 +39,7 @@ module Concurrent
     call_dataflow(:value, executor, *inputs, &block)
   end
   module_function :dataflow_with
-  
+
   def dataflow!(*inputs, &block)
     dataflow_with!(Concurrent.global_io_executor, *inputs, &block)
   end
@@ -50,12 +50,14 @@ module Concurrent
   end
   module_function :dataflow_with!
 
-  private 
+  private
 
   def call_dataflow(method, executor, *inputs, &block)
     raise ArgumentError.new('an executor must be provided') if executor.nil?
     raise ArgumentError.new('no block given') unless block_given?
-    raise ArgumentError.new('not all dependencies are IVars') unless inputs.all? { |input| input.is_a? IVar }
+    unless inputs.all? { |input| input.is_a? IVar }
+      raise ArgumentError.new("Not all dependencies are IVars.\nDependencies: #{ inputs.inspect }")
+    end
 
     result = Future.new(executor: executor) do
       values = inputs.map { |input| input.send(method) }


lib/concurrent/dataflow.rb
```diff
@@ -50,12 +50,15 @@ def dataflow_with!(executor, *inputs, &block)
...
 
-  private 
+  private
 
   def call_dataflow(method, executor, *inputs, &block)
     raise ArgumentError.new('an executor must be provided') if executor.nil?
     raise ArgumentError.new('no block given') unless block_given?
-    raise ArgumentError.new('not all dependencies are IVars') unless inputs.all? { |input| input.is_a? IVar }
+    not_ivars = inputs.map { |input| !input.is_a? IVar }
+    unless not_ivars.empty?
```
jdantonio: @schneems Thank you for taking the time to work on this update. There are several failing tests we need to look into. I haven't had an opportunity to pull the code and run the tests locally, but I believe this line is the culprit. Won't the `#map` function create an array with a number of elements equal to the size of the `inputs` array? I think `#empty?` will always return `false` in this case. Can you please take a look?

schneems: Yes, that should have been a `select`, it was right locally but not committed. I'm updating to not allocate an array on every run, will preserve the original logic.

schneems:
Updated to use the original logic, instead of only showing the non conforming inputs, we can easily show all of the inputs. By not using a `select` we can avoid allocating an array so speed wont be impacted.

schneems:
Tests are back to passing

lucasallan:
:+1:

PR #479, opened 2015-12-03
Fix typos in async.rb

eregon:
There is also a sentence which made me wonder its meaning:
```
Assignment is not thread-safe in Ruby
```
What kind of assignment? Surely, `var = value` is atomic, right? Is it about visibility?

### DIFF
--- a/lib/concurrent/async.rb
+++ b/lib/concurrent/async.rb
@@ -58,7 +58,7 @@ module Concurrent
   # end
   # ```
   #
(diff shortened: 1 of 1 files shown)


ianks:
Assignment is thread-safe in MRI currently, but not other platforms (jruby/rubinius). And you are right, it is about visibility.

chrisseaton:
> Surely, var = value is atomic, right? Is it about visibility?

Who knows with bindings!

Issue #480, opened 2015-12-07
NameError: uninitialized constant Concurrent::Future::Options


tarcieri:
Involved gem versions can be inferred from the backtrace. JRuby 9.0.4.0.

```
NameError: uninitialized constant Concurrent::Future::Options
/project/vendor/bundle/jruby/2.2.0/gems/concurrent-ruby-1.0.0-java/lib/concurrent/future.rb:134:in `ns_initialize'
/project/vendor/bundle/jruby/2.2.0/gems/concurrent-ruby-1.0.0-java/lib/concurrent/ivar.rb:66:in `block in initialize'
/project/vendor/bundle/jruby/2.2.0/gems/concurrent-ruby-1.0.0-java/lib/concurrent/ivar.rb:66:in `synchronize'
/project/vendor/bundle/jruby/2.2.0/gems/concurrent-ruby-1.0.0-java/lib/concurrent/ivar.rb:66:in `initialize'
/project/vendor/bundle/jruby/2.2.0/gems/concurrent-ruby-1.0.0-java/lib/concurrent/future.rb:32:in `initialize'
/project/vendor/bundle/jruby/2.2.0/gems/concurrent-ruby-1.0.0-java/lib/concurrent/future.rb:75:in `execute'
/project/vendor/bundle/jruby/2.2.0/gems/sprockets-3.5.1/lib/sprockets/manifest.rb:192:in `block in compile'
```

tarcieri:
This occurs when sprockets is precompiling assets as part of a standard Rails workflow

atambo:
Hitting this on jruby 1.7.23 as well :(

jdantonio:
This is a problem with an autoload call behaving differently on JRuby than MRI. The `manifest.rb` file in sprockets calls `require 'concurrent/future'` rather than the recommended `require 'concurrent'`. I've opened an [issue with sprockets](https://github.com/rails/sprockets/issues/199). I'll keep working on it from my end and see if I can find a solution. In the interim, please try adding `require 'concurrent'` in your Rakefile and see if that clears the error for now.

// @schneems @headius

jdantonio:
Possible fix in PR #481

cavpollo:
Adding <code>require 'concurrent'</code> in the Rakefile fixed temporarly the issue, thanks

jdantonio:
@cavpollo Thank you for testing this.

[...]

PR #481, opened 2015-12-08
Resolved NameError on JRuby by using fewer autoloads.

jdantonio:
See issue #480.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ rvm:
   - 1.9.3
   - ruby-head
   - jruby-1.7.19
+  - jruby-9.0.1.0
   - jruby-9.0.3.0
   - jruby-9.0.4.0
   - jruby-head
--- a/lib/concurrent/delay.rb
+++ b/lib/concurrent/delay.rb
@@ -5,6 +5,7 @@ require 'concurrent/synchronization'
 
 module Concurrent
 
+  # This file has circular require issues. It must be autoloaded here.
   autoload :Options, 'concurrent/options'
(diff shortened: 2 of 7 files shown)


jdantonio:
@pitr-ch Can you do a quick code review run some local tests against this, please? I want to make sure there aren't any unintended bugs created by this.

I'll pull the latest Rails code and run its tests agains this branch to verify I haven't broken anything there and that we aren't generating warnings.

Here are the shell commands I ran to spot check this:

```
be ruby -w -e "require 'concurrent/future'; Concurrent::Future.execute { puts 'Boom' }"
be ruby -w -e "require 'concurrent/promise'; Concurrent::Promise.execute { puts 'Boom' }"
be ruby -w -e "require 'concurrent/scheduled_task'; Concurrent::ScheduledTask.execute(0.1) { puts 'Boom' }"
```

schneems:
Could we introduce some integration style tests for something like this? Small scripts that use different parts of this gem that we can call via Process.spawn and assert success or output perhaps?

jdantonio:
That's a very good idea. I have a set of scripts in the `build-tests` directory that I run whenever a create a new release. I have to manually watch them run and verify the results. I really need to automate those, too. I'll work on that as soon as I can.

schneems:
I do it all the time. Here's an example in Puma Worker Killer https://github.com/schneems/puma_worker_killer/blob/9be33d056d86f8bd312b815f30395934dee3d500/test/puma_worker_killer_test.rb#L9

PR #496, opened 2016-01-16
Better API for future zipping

### LINKED ISSUE #438: Inconsistent `zip` return types
The inconsistent return types of `Concurrent#zip` (Edge) has led to some annoying boilerplate code in our codebase. To demonstrate the inconsistency:

```ruby
foo = []
bar = [Concurrent.succeeded_future(1)]
baz = [Concurrent.succeeded_future(1), Concurrent.succeeded_future(2)]

zip_foo = Concurrent.zip(*foo) # => Concurrent::Edge::Event
zip_bar = Concurrent.zip(*bar) # => Concurrent::Edge::Future
zip_baz = Concurrent.zip(*baz) # => Concurrent::Edge::Future

zip_foo.value! # => raises undefined method `value!` for Concurrent::Edge::Event
zip_bar.value! # => 1
zip_baz.value! # => [1, 2]
```

`zip_baz` feels correct to me, while I think `zip_bar.value!` should return a collection, i.e. `[1]`. And I don't understand if there's a valid reason for `zip_foo.value!` to return a different type (an `Event` instead of a `Future`).

As it stands now, I find myself writing a lot of code that looks like this:

```ruby
futures = widget.parts.map do |part|
  Concurrent.future { do_something_with(part) }
end

x = futures.present? ? Concurrent.zip(*futures) : Concurrent.succeeded_future([])

x.then do |*futures|
  Array(futures).map(&:some_other_task)
end
```

If `zip` always returned a collection, I could just write:

```ruby
futures = widget.parts.map do |part|
  Concurrent.future { do_something_with_part(part) }
end

Concurrent.zip(*futures).then do |*futures|
  futures.map(&:some_other_task)
end
```

pitr-ch:
fix #438

### DIFF
--- a/lib/concurrent/edge/future.rb
+++ b/lib/concurrent/edge/future.rb
@@ -106,6 +106,7 @@ module Concurrent
       # @return [Future]
       def select(*channels)
         future do
+          # noinspection RubyArgCount
           Channel.select do |s|
             channels.each do |ch|
               s.take(ch) { |value| [value, ch] }
@@ -514,9 +515,9 @@ module Concurrent
       # @!visibility private
       class PartiallyFailed < CompletedWithResult
         def initialize(value, reason)
+          super()
           @Value  = value
           @Reason = reason
-          super()
         end
 
         def success?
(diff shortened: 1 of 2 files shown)


hobodave:
:+1: Looking forward to this.

PR #498, opened 2016-01-19
Alter backends used by Truffle

chrisseaton:


lib/concurrent/atomic_reference/jruby+truffle.rb
```diff
@@ -1 +1 @@
-require 'concurrent/atomic_reference/rbx'
+require 'concurrent/atomic_reference/mutex_atomic'
```
pitr-ch: Why this switch?

lib/concurrent/synchronization/lockable_object.rb
```diff
@@ -13,7 +13,7 @@ module Synchronization
                                    when Concurrent.on_rbx?
                                      RbxLockableObject
                                    when Concurrent.on_truffle?
-                                     TruffleLockableObject
+                                     MriMutexLockableObject
```
pitr-ch: object.rb should also have MRI implementation for j+t

lib/concurrent/atomic_reference/jruby+truffle.rb
chrisseaton: We seem to spin very badly when using CAS in JRuby+Truffle at the moment.

lib/concurrent/synchronization/lockable_object.rb
chrisseaton: I don't get any spec failures (or I do but they're intermittent) so I assumed that whatever `object.rb` is doing must be ok at the moment.

pitr-ch: The Synchronisation::Object also has different implementations, it was never designed to be cross used, it may work but it's probably just accidental and may break any time.

PR #499, opened 2016-01-22
Test with JRuby+Truffle on Travis

chrisseaton:
In order to make JRuby+Truffle work here I've had to do a few things:

* bypass normal `bundler`, as we don't support `openssl`, and `rake`, as we can't run that yet (or nobody has tried - it probably works somewhat)
* disable `simplecov` and `coveralls` as we don't run them yet
* tag a few failing specs
* run the specs in batches as I think we leak threads and it all seizes up if you run them in one go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,14 +25,21 @@ branches:
     - master
 
 matrix:
+  include:
+    - rvm: jruby-head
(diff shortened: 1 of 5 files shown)


pitr-ch:
Nice! Thanks.

Issue #500, opened 2016-01-22
ThreadExecutor doesn't survive fork


matthewd:
```ruby
require 'concurrent'
20.times { Concurrent::Future.execute { 1 } }
fork { Concurrent::Future.execute { 7 }.wait! }
```

This causes a deadlock in the child, because its executor is full of dead threads.

It feels like the executor should be able to detect that condition, and flush the dead workers out of the pool.

pitr-ch:
@matthewd Thanks for contacting us and letting us know. Unfortunately this will require larger discussion. It's not just about how to add a callback to restart the threads after forking. Concurrent-ruby is designed to run on any Ruby implementation. Forking is not supported on all of them, so we need to come up with some solution/API which will work on all Ruby implementations. @matthewd could you elaborate on your use-case? It'll give us a better idea what needs to be added.

jdantonio:
My initial reaction to this request is that this isn't a bug. Passing a thread pool to a forked process isn't something we should be reasonably expected to support. The workers we would be flushing aren't "dead" threads, they are references to threads that *don't exist* in the forked process. Forking and threading are both concurrency techniques. They have distinctly different advantages and disadvantages. Mixing the two feels like an anti-pattern to me.

That being said, I'm certainly open to discussing it further. Please describe the use case in more detail. That context will help us better understand why you are requesting this feature.

matthewd:
concurrent-ruby has global executors (like the one used for futures). If those globals don't survive a fork, then concurrent-ruby is incompatible with forking, and that's gonna be a problem.

> Forking is not supported on all of them, so we need to come up with some solution/API which will work on all Ruby implementations.

This part is easy: if the implementation doesn't support forking, then there's no problem to worry about.

> The workers we would be flushing aren't "dead" threads, they are references to threads that don't exist in the forked process.

Indeed... I was talking about the ruby-level Thread's state.

---

My argument is that if the executor is idle at the time of the fork, the fact the threads technically still exist is, as far as the caller is concerned, merely a transparent performance optimisation.

If it's not supported to create (and then wait for) a future within a method, because somewhere much later the application may have reason to fork (and would then have an at best incomplete, at worst useless, executor pool), then I don't see how any library can use concurrent-ruby internally.

---

My proposed solution doesn't require noticing the fork at the time it occurs: you just need to know it's happened next time you go to enqueue work. In the simple case, the executor knows it was idle, and it can throw away the "dead" threads with impunity.

[...]

PR #501, opened 2016-01-24
Reset ThreadPoolExecutor on fork.

jdantonio:
When a `ThreadPoolExecutor` in the *forked* process detects that a fork has occurred it immediately takes the following actions:

* Clears all pending jobs from its queue (assuming they will be handled by the *forking* process).
* Deletes all worker threads (they will have died during the fork).
* Resets all job counters (these counts will be reflected in the *forking* process).
* Begins posting new jobs as normal.


jdantonio:
This latest update actually clears the pool if it detects a fork. I've updated the test script to perform a fork using the example given [here](https://github.com/ruby-concurrency/concurrent-ruby/issues/500#issuecomment-174224308). The new implementation survives the fork and posts the new job. I ran the test script on my work computer (2.8 GHz Intel Core i7, 16 GB 1600 MHz DDR3) and the difference in performance is still minimal.

matthewd:
> We cannot use Thread.main to determine if our process has forked because Ruby copies all attributes of Thread.main on the new fork.

FWIW, my imagined `Thread.main`-using solution would look something like this: https://gist.github.com/matthewd/aeb876b9fad8e189bcff (modulo a method on Worker to avoid the ivar get, of course).

But yeah.. my results with that script match yours: they're so close I can't even get the original, clearly-less-code, version to consistently bench as the fastest... and when they do, $$ seems to be faster than the thread check anyway.

examples/benchmark_thread_pool_executor.rb
```diff
@@ -0,0 +1,142 @@
...
+    ns_reset_if_forked
+    super
+  end
+
+  def ns_reset_if_forked
+    if $$ != @ruby_pid
+      @queue.clear
+      @ready.clear
+      @pool.clear
+      @ruby_pid == $$
```
matthewd: `=`

jdantonio: Ty. Fixed.

jdantonio:
@matthewd @pitr-ch @chrisseaton Please review.

doc/thread_pools.md
```diff
@@ -159,3 +159,20 @@ future = Future.new(:executor => pool).execute do
...
+
+*Concurrent Ruby guarantees that any thread pools created on by the forking process before the fork remain functional on the forked process after the fork. It also guarantees that jobs post to a thread pool before a fork remain on the thread pool in the forking process after the fork.*
```
matthewd: "created on by"

---

Maybe "remain on the thread pool in (only) the forking process"?

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -273,6 +276,18 @@ def ns_prune_pool
       @next_gc_time = Concurrent.monotonic_time + @gc_interval
     end
 
+    def ns_reset_if_forked
```
matthewd: `ns_shutdown_execution` should also call this, to ensure the pool is "valid" before it waits on it

jdantonio: Good catch. Done.

doc/thread_pools.md
```diff
@@ -159,3 +159,20 @@ future = Future.new(:executor => pool).execute do
...
+
+*Concurrent Ruby guarantees that all threads created by thread pools in the forking process before the fork remain only in the forking process after the fork. It also guarantees that jobs post to a thread pool before a fork remain on the thread pool in the forking process after the fork. Finally, it guarantees that thread pools copied to a forked process will continue to function normally on the forked process.*
```
eregon: Looks good, but the terminology here confused me a couple times.
Maybe `parent`/`child` would already help to figure which process we are talking about.
Also, the first guarantee is not a Concurrent Ruby one but one of the semantics of `fork(2)` at the OS level, right?

Maybe something simpler along the lines of
"Current jobs will be processed by the parent process. The child process does not inherit any job." would help.

schneems:
Another angle of attack is to ask the the executor if it has a thread that is alive with `Thread#alive?` I benchmarked speed and it's in the same ballpark as `$$` (though slightly slower), we could either loop through workers via `Enumerable#detect` or set up a "watchdog" thread that we would check every time to see if it is alive. I don't know if there's other occasions when threads might die unexpectedly other than via forking, or if it makes sense to turn this special failure case into a general failure case.

doc/thread_pools.md
```diff
@@ -159,3 +159,20 @@ future = Future.new(:executor => pool).execute do
    #work
 end
 ~~~
+
+## Forking
+
+Some Ruby versions allow the Ruby process to be [forked](http://ruby-doc.org/core-2.3.0/Process.html#method-c-fork). Generally, mixing threading and forking is an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern). Threading and forking are both concurrency techniques and mixing the two is rarely beneficial. Moreover, Ruby does not copy any threads execept the main thread when forking. Thus threads created before the fork become unusable ("dead") in the forked process. This aspect of forking is a significant issue for any application or library which spawns threads. It is strongly advised that applications using `ThreadPoolExecutor`, either directly or indirectly (via high-level concurrency abstractions like `Future` and `Actor`), do **not** also fork. Since Concurrent Ruby is a foundational library often used by gems which are in turn used by other applications, it is impossible to predict or prevent upstream forking. Concurrent Ruby therefore makes a few guaratnees about the behavior of `ThreadPoolExecutor` after forking.
```
eregon: "Moreover, Ruby does not copy any threads execept the main thread when forking."
s/Ruby/the Operating System/ I believe.
Also, it's not necessarily the main thread:
`ruby -e 'p Thread.main; Thread.new { fork {p Thread.main; p Thread.current}}.join'`

Maybe use the example of locks taken by threads (other than the one forking) as why it's an anti-pattern (the lock would remained locked forever and unlock-able in the child process).

jdantonio:
Another update to the docs.

schneems:
Can we get a release cut with this fix? Any major issues to date?

jdantonio:
I'll cut a release this week. I was just waiting to see if other bugs were reported.

schneems:
Great, thanks!

jdantonio:
@schneems v1.0.1 has been released. With this update you should now also be able to safely use `require 'concurrent/future'` and not experience problems on JRuby. Emphasis on "should." :-)

schneems:
Awesome! Thanks :heart: we need to get sprockets-rails back in the green. The threads dying on fork was causing failures

Issue #144, opened 2014-08-07
IO example


elcuervo:
I know there's some ides in #142 but I'd like to see some examples like `celluloid-io` but using `concurrent-ruby`. Anyone have some or should I just do some?

asppsa:
I've created a simple gem that does IO using the Actors framework [here][io_actors].  Included in the "examples" folder is an example that does "ping pong" over a socket pair.  This isn't really "battle tested" yet -- any feedback on the approach taken is appreciated.

[io_actors]: https://github.com/asppsa/io_actors

pitr-ch:
@asppsa that's great! I did not have time look yet though, I'll do so when I' am back from vacation in a week. Would you consider working with us to create utils for IO cooperation in `concurrent-ruby` gem itself? I would very much like to have some support in `concurrent-ruby` directly.

jdantonio:
@asppsa +1 to what @pitr-ch said. I've take a quick look at your gem and it looks great. We've had a number of requests for better IO-based operations and I'd love to add something of that nature. If you would be interested in possibly implementing some of your work directly in `concurrent-ruby` we'd love to talk more.

asppsa:
Hi guys,

Firstly, I'm very flattered.  In principle I'd be all for getting this
into concurrent-ruby!  I extracted this code out of a closed source
project I am working on with the hope that others could help to improve
the quality of the implementation.  Making it part of your project would
definitely help to advance that goal from what I can see!

The main issue I can see with it at the moment is the dependency on
[nio4r][1], which in MRI/RBX land is essentially a wrapper around libev.
I'm guessing that this hard dependency would have to be removed, as
otherwise people installing CR would have to install libev too, even if
they did not intend to do any IO.  This is easily solved however -- I
can rework the code to probe for nio4r and fall back to `IO#select` when
it is not available (perhaps displaying a warning).

A secondary problem is that there seem to be some garbage collection
issues with the code as-is.  The ping-pong example shows the total
number of actors in memory (and those "alive") growing over time,
whereas it should remain constant.  Because of these problems, I've not
been able to integrate the gem back into the project that it originally
came out of as yet.  From my perspective though this is actually good
reason to get this code included in CR, as it will hopefully mean more
people testing it and submitting patches.  Still, it's worth noting that
the code as-is is a bit wonky.

[1]: https://github.com/celluloid/nio4r

asppsa:
Hi there,

[...]

PR #502, opened 2016-01-28
add an udp server example

schmurfy:
This might not be mergeable as is but so far this is the best I got for creating an udp server, what I was going for here is that if anything bad happens the server should restart and still be able to receive data.

The design started from a discussion in #144 with the modified code @pitr-ch and after playing around a bit I came to this, I am not entirely happy with this but it works, if you send "crash" to it it will crash and get restarted to continue process the inputs, the udp socket itself is never stopped.

If anyone has any idea to improve this you are welcome :)

### DIFF
--- a/examples/udp_server.rb
+++ b/examples/udp_server.rb
@@ -1,0 +1,60 @@
+require 'bundler/setup'
+require 'concurrent/actor'
+
+
+class MyUDPServer < Concurrent::Actor::RestartingContext
+  @@thread = {}
+
+  def initialize
+    super
+    puts "started [#{name}]"
+    if @@thread[name] == nil
+      tell(:start)
+    end
+  end
+
+  def on_message(msg)
+    command, *args = msg
+    case command
+    when :start then  start()
+    when :stop  then  stop()
+    when :got
+      p args
+      if args[0].start_with?('crash')
+        raise "arrrrrr"
+      end
+      
+    else
+      pass
+    end
+  end
+
+  def start
+    server = UDPSocket.new
+    server.setsockopt(:SOCKET, :REUSEADDR, true)
+    server.bind('127.0.0.1', 5555)
+
+    @@thread[name] = Thread.new(server) do |server|
+      begin
+        while true
+          p [:loop]
+          data, src = server.recvfrom(4096)
+          p [:data, data]
+          tell( [:got, data, src] )
+        end
+      rescue Exception => err
+        p [:err, err]
+      end
+    end
+  end
+
+  def default_executor
+    # this does not do IO no need for io executor
+    Concurrent.global_fast_executor
+  end
+
+end
+
+a = MyUDPServer.spawn!('udp1')
+
+sleep


pitr-ch:
Thank you for opening the PR!

examples/udp_server.rb
```diff
@@ -0,0 +1,60 @@
+require 'bundler/setup'
+require 'concurrent/actor'
+
+
+class MyUDPServer < Concurrent::Actor::RestartingContext
+  @@thread = {}
```
pitr-ch: I am curious why is a class variable used?

schmurfy: I couldn't figure out a way to properly exit the thread if the actor dies so I tried another road which was to keep the receiving thread linked to the socket alive and only restart the actor itself, the thread variable is used so that the thread is only spawned once.

As I said in my pr message I not that entirely happy with this solution but I tested it and at least it works, at least from what I remember when I tried :)

pitr-ch:
Thank you @schmurfy I have look later and I'll add it to our documentation.

schmurfy:
it should be refined before adding to the documentation but I wanted to start the conversation, if you have ideas on improvements I can give them a try, the main thing I was looking for was to have a reliable, fault resilient actor receiving the udp packets, it might miss a few but no matter what happens it should resume listening.

pitr-ch:
@schmurfy could you have look if this https://gist.github.com/pitr-ch/119e48c75ab5ccbbdac9 works for you? If yes, please update the PR.

schmurfy:
Sure but after a quick look I am pretty sure it has the same issue I tried to dodge: if the server crash it will not be able to rebind on the same port because it will be already used (the old thread won't be dead), so if your actor crash for any reason you won't ever receive another message.

pitr-ch:
Isn't the tread terminated on actor crash? https://gist.github.com/pitr-ch/119e48c75ab5ccbbdac9#file-udp_server-rb-L38

schmurfy:
I just did another test with your version, I just added a crash condition as I did in mine when receiving "crash" and here what I get hust after the crash:

```
[2016-05-04 11:43:28.363] DEBUG -- /udp1: event: [:terminated, nil] (public)
[2016-05-04 11:43:28.363] DEBUG -- /: was told [:terminated, nil] by #<Concurrent::Actor::Reference:0x7fe99516a9a0 /udp1 (MyUDPServer)>
[2016-05-04 11:43:28.363] DEBUG -- /: was told :remove_child by #<Concurrent::Actor::Reference:0x7fe99516a9a0 /udp1 (MyUDPServer)>
[2016-05-04 11:43:34.945] DEBUG -- /udp1: was told [:read, #<UDPSocket:fd 10>] by #<Thread:0x007fe995160270>
[2016-05-04 11:43:34.946] DEBUG -- /udp1: rejected [:read, #<UDPSocket:fd 10>] from #<Thread:0x007fe995160270>
[2016-05-04 11:43:34.946] DEBUG -- /default_dead_letter_handler: was told #<Concurrent::Actor::Envelope:0x007fe995141ff0 @message=[:read, #<UDPSocket:fd 10>], @future=nil, @sender=#<Thread:0x007fe995160270@examples/udp_server.rb:52 sleep>, @address=#<Concurrent::Actor::Reference:0x7fe99516a9a0 /udp1 (MyUDPServer)>> by #<Concurrent::Actor::Reference:0x7fe99516a9a0 /udp1 (MyUDPServer)>
[2016-05-04 11:43:34.947]  INFO -- /default_dead_letter_handler: got dead letter #<Concurrent::Actor::Envelope:0x007fe995141ff0 @message=[:read, #<UDPSocket:fd 10>], @future=nil, @sender=#<Thread:0x007fe995160270@examples/udp_server.rb:52 sleep>, @address=#<Concurrent::Actor::Reference:0x7fe99516a9a0 /udp1 (MyUDPServer)>>
```

The messages appear to be received so I imagined that the socket is either never closed or the reopening works but the read handler is never called after the crash

pitr-ch:
Sorry for letting this PR without attention for so long. I'll get back to in next release.

PR #514, opened 2016-02-19
Fix Future + Executor example

tak1n:
When using Future.new you have to pass the block of work to it.


tak1n:
Oops, forgot ci skip sry

jdantonio:
Thank you for pointing this out and updating the doc. Can you please change the example to use `Future.execute`? That's the preferred approach. Also, I'm not a big fan of making method calls off of an `end` and in `end.execute`.

tak1n:
I'm also not a fan of `end.execute` :+1: 
Updated it to pass in the executor when calling `Future.execute`

PR #515, opened 2016-02-25
Improve AtomicFixnum type-checking error message

soulcutter:
`rubyFixnumToLong` is used in increments/updates in addition to initialization
so the error message here should be more generic.


soulcutter:
This came up in a real application where I had something like the following:

```ruby
@metrics = Concurrent::Map.new { |h, k| h[k] = Concurrent::AtomicFixnum.new(0) }
[…]
metrics[metric_name].increment (clock.now - start_time).to_f # ERROR LINE IN APPLICATION CODE

ERROR -- : initial value must be a Fixnum
com/concurrent_ruby/ext/JavaAtomicFixnumLibrary.java:72:in `increment']
```

jdantonio:
Thank you!

PR #519, opened 2016-03-12
[WIP] Integrate CSP channels from Normandy

### LINKED ISSUE #510: CSP channels implementation
I've had some chat with @coding-horror over Twitter and concurrent-ruby was brought again to my mind, so I went on to look for the channel implementation, which I found to be lacking. Turns out I've had some implementation collecting dust, so maybe someone over here could make use of this.

I strived to keep the implementation simple, keeping the Go spirit while still being idiomatic Ruby: take a look at the [examples](https://github.com/lloeki/normandy/tree/master/examples) and open the corresponding Go by Example page.

The implementation is almost feature complete save for non-blocking select, for which I'm still looking for a straightforward and idiomatic syntax.

Anyway, feel free to ignore this if you don't feel like it's fit for this project.

lloeki:
Started off issue #510, this PR aims to track work on integration of [normandy](https://github.com/lloeki/normandy/tree/master/examples) into concurrent-ruby, centralising questions and comments about the code as I get along.

`WIP` marker file and title tag will be removed once the code crosses a sufficient threshold.

fix #510


pitr-ch:
Thank you for working on this! Since this is rather big, please comment if you'd like somebody to review, or if you have any questions about c-r internals.

lloeki:
This is not ready for a detailed review yet, but general feedback is appreciated.

I'm currently looking into [select statements](https://golang.org/ref/spec#Select_statements). The current implementation is nice to read in Ruby, simple to implement, and reasonably fast, but it currently doesn't support non-blocking via default clause nor writes. Having our select implementation behave reasonably closely to Go's *"Execution of a "select" statement proceeds in several steps"* while still being readable and efficient is quite a challenge but I've been prototyping some ideas successfully.

Another step is to leverage thread pooling instead of using Ruby's `Thread.new` directly.

`WaitGroup` is outside of `Channel` because it's completely independent and the construct is definitely usable and useful without channels, the only requirement being that if both need to be used at the same time they have to do so within the same concurrency model (see `WaitGroup#wait`).

Within the `normandy` gem I encapsulated the concurrency model within `Channel::Runtime`, which I ported as is here (my dream is that someday a Ruby implementation could support non-cooperative lightweight concurrency similar to goroutines ;-). I get the feeling this isn't required here though, as if I'm not mistaken the concurrency model prevalent here is to use thread pools. Yet using `Channel.go` just sounds wrong to me, as "goroutines" (or whatever concurrency model is in use) is something orthogonal to channels, as it can be seen with `WaitGroup`, so `Runtime` should theoretically be expected to live outside `Channel`. Well, for now I'll consider this a bikeshed issue and leave it be but this is something I'd like feedback on later on.

lloeki:
Oh by the bays, is there any style guide I should be following? Do you contemplate using [Rubocop](https://github.com/bbatsov/rubocop) via a `.rubocop.yml` file? Should you be interested, this is also something I may be open to contribute to (just as I did [for Homebrew](https://github.com/Homebrew/legacy-homebrew/pull/33219)).

jdantonio:
> my dream is that someday a Ruby implementation could support non-cooperative lightweight concurrency similar to goroutines

You and me both, my friend. You and me both! :-)

> if I'm not mistaken the concurrency model prevalent here is to use thread pools

Correct. We have a global thread pool to which we post tasks. Most of our abstractions also support custom thread pools via dependency injection (using an option called `:executor`). We generally discourage users from creating their own thread pools unless they are experts, however.

> Yet using Channel.go just sounds wrong to me, as "goroutines" (or whatever concurrency model is in use) is something orthogonal to channels

I agree, but we also don't want users to create their own thread pools. `Channel.go` is a convenience that allows the user to access the global thread pool while keeping the underlying details hidden. I'm fine getting rid of `Channel.go` but we'll want to provide a similar convenience. We can figure that part out later once channels are fully implemented.

> is there any style guide I should be following?

No strict style guide, just a few [guidelines](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/CONTRIBUTING.md#follow-the-guidelines)

> Do you contemplate using Rubocop via a .rubocop.yml file?

I hadn't planned on it. I appreciate that people are willing to give their free time to the project so I've tried to make participation as easy as possible (writing concurrency tools is hard enough!). I'm not opposed to the idea (we use static analysis at work). I just don't want to create a situation where we're spending a ton of time fixing linter errors and not implementing new features.

pitr-ch:
@lloeki Would you be interested to pick up the work on this?

lloeki:
@pitr-ch Definitely interested. It's been quite some time though, I'll need some time to dive in again.

pitr-ch:
@lloeki Great! Let me know if you need anything. I had a brief look at the old changes, it might be better to add the new implementation next to the old one. Even though it's in edge, I suspect some people are using it so we need to deprecate it gracefully alongside the new implementation.

zhulik:
Any updates on this? Current channels implementation is completely unusable due to #563

lloeki:
@zhulik no updates, but I still want to work on this at some point. Glad to see there's still some interest!

pitr-ch:
@lloeki great. Before you do please have a look at http://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Promises/Channel.html and we should have a chat. I would like any new abstractions to be more integrated together rather than standing on their own in the gem.

chrisseaton:
Thank you very much for this work.

I'm going to close this issue as we going into a phase of attempting to get Concurrent Ruby under control and better maintained, rather than adding new features. Please don't throw away the work, and please do re-suggest the feature in the future, but I'm not likely to merge at the moment.

PR #529, opened 2016-04-21
Conc::Map - MRI backend fixes

### LINKED ISSUE #528: Hidden pitfalls in Conc::Map's MRI strategy?
There is a report by a Rails user (rails/rails#24627) that on MRI `Concurrent::Map`'s `@backend` (a pure Ruby `Hash`) is throwing "insertion during iteration" errors.

As pointed out by @jeremy, there are couple of candidates:

#### `NonConcurrentMapBackend#clear`

The method calls `@backend.clear`, the C implementation (https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/hash.c#L1488) of `Hash#clear` might invoke `rb_hash_foreach`, that in turn [bumps iter level](https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/hash.c#L384) and after some preparation dispatches into `st_foreach_check` (https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/st.c#L890).

While in `st_foreach_check` I don't see anyway for MRI to yield to Ruby land (release GVL), except for the provided iterator fun, but in the case of `Hash#clear` https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/hash.c#L1465 `clear_i` also doesn't yield to Ruby.

If the code never releases GVL, it is "atomic", the elevated iter level is also atomic and is never exposed to other Ruby code.

##### Conclusion

`NonConcurrentMapBackend#clear` is safe: `Hash#clear` never yields GVL.

---------

#### `NonConcurrentMapBackend#value?`

The method calls `@backend.value?`, the C implementation (https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/hash.c#L2050) of `Hash#value?` uses `rb_hash_foreach` with `rb_equal` (https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/hash.c#L2029) callback, `rb_equal` might invoke custom Ruby-level methods.

##### Conclusion

`NonConcurrentMapBackend#value?` is not safe: `Hash#value?` might yield GVL.

##### Solution

Remove custom `value?` impl in `NonConcurrentMapBackend`, let the default method (https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/map.rb#L147) do the work.

---------

#### `NonConcurrentMapBackend#initialize_copy`

The method doesn't call `Hash#initialize_copy`, the copying work is performed higher up in `Map#populate_from` (https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/map.rb#L216), this in turn relies on `Map#each_pair`.

##### Conclusion

`NonConcurrentMapBackend#initialize_copy` is safe as long as `Map#each_pair` is safe.

---------

#### `NonConcurrentMapBackend#each_pair`

The method calls `@backend.dup` (in `dupped_backend`). `Hash` duping is then done via `Hash#initialize_copy`, the C implementation (https://github.com/ruby/ruby/blob/2f6821a7c4e7a52d/hash.c#L1581) uses `st_copy` to copy the underlying hash table structure, while doing so, it doesn't yield to Ruby. After the copying is done, the new copy is rehashed by `rb_hash_rehash`, however we don't care about what happens in `rb_hash_rehash`: it operates on a thread local copy.

##### Conclusion

`NonConcurrentMapBackend#each_pair` is safe: `Hash#dup` might yield GVL only while working on thread local data.

---------

So we need to fix `value?` otherwise everything looks fine?

thedarkone:
This should hopefully cover all "insertion during iteration" errors for MRI backend.

Fixes #528.
Possibly related: rails/rails#24627.

### DIFF
--- a/lib/concurrent/collection/map/non_concurrent_map_backend.rb
+++ b/lib/concurrent/collection/map/non_concurrent_map_backend.rb
@@ -76,10 +76,6 @@ module Concurrent
         @backend.key?(key)
       end
 
-      def value?(value)
-        @backend.value?(value)
-      end
-
       def delete(key)
         @backend.delete(key)
       end
--- a/lib/concurrent/collection/map/synchronized_map_backend.rb
+++ b/lib/concurrent/collection/map/synchronized_map_backend.rb
@@ -53,10 +53,6 @@ module Concurrent
         synchronize { super }
       end
 
-      def value?(value)
-        synchronize { super }
-      end
-
       def delete(key)
         synchronize { super }
       end
--- a/lib/concurrent/map.rb
+++ b/lib/concurrent/map.rb
@@ -149,7 +149,7 @@ module Concurrent
         return true if value.equal?(v)
       end
       false
-    end unless method_defined?(:value?)
+    end
 
     def keys
       arr = []
--- a/lib/concurrent/map.rb
+++ b/lib/concurrent/map.rb
@@ -202,6 +202,15 @@ module Concurrent
 
     undef :freeze
 
+    # @!visibility private
+    DEFAULT_OBJ_ID_STR_WIDTH = (2**50).class == Fixnum ? 14 : 7 # we want to look "native", 7 for 32-bit, 14 for 64-bit
+    # override default #inspect() method: firstly, we don't want to be spilling our guts (i-vars), secondly, MRI backend's
+    # #inspect() call on its @backend i-var will bump @backend's iter level while possibly yielding GVL
+    def inspect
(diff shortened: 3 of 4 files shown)


jeremy:
I'm not seeing how the Action View resolver cache could be hitting either of these paths, but: fingers crossed!

thedarkone:
@jdantonio please check if I'm doc-hiding `DEFAULT_OBJ_ID_STR_WIDTH` correctly. Other than that we should merge and release something for Rails (so they can bump their dep).

@jeremy I'm also skeptical that this is going to fix that Rails issue. Btw thanks for the help, especially pointing out `#inspect` as a possible culprit, I would never have thought of that :smile:.

PR #530, opened 2016-04-29
Fix cached_thread_pool to allow to overwite default values for min_th…

sbrossie:
…reads, max_threads, max_queue


jdantonio:
@sbrossie Thank you for the PR, but the original behavior is correct. A `CachedThreadPool` is defined as a thread pool with no maximum queue size, no minimum number of threads, and no maximum. That's why those are overrides. Users can set *other* options, but they cannot change those. Users who need more control can use `ThreadPoolExecutor` directly. `CachedThreadPool` is simply a convenience class that guarantees that certain configuration values are always set.

sbrossie:
Ok got it.

PR #554, opened 2016-07-19
Fix JRuby Semaphore not allowing #new and #reduce_permits with zero

ivoanjo:
These behaviors were different from the non-JRuby semaphore implementation, where this was already allowed.

I also improved the test coverage to cover the behavior of the various other methods when getting `0` as argument.


jdantonio:
Nice catch @ivoanjo! Thank you very much for fixing this.

PR #557, opened 2016-07-23
Allow ThreadLocalVar to be initialized with a block

ivoanjo:
`ThreadLocalVar`s can now receive a block in the `#initialize` method, which will be called to initialize the value on each thread.

This allows default initialization to work with non-value types, where
each thread gets its own copy:

```ruby
v = ThreadLocalVar.new { {} }
h1 = Thread.new { v.value }.value
h2 = Thread.new { v.value }.value

h1.equal? h2 # -> false
```

### DIFF
--- a/lib/concurrent/atomic/abstract_thread_local_var.rb
+++ b/lib/concurrent/atomic/abstract_thread_local_var.rb
@@ -8,8 +8,17 @@ module Concurrent
   class AbstractThreadLocalVar
 
     # @!macro thread_local_var_method_initialize
-    def initialize(default = nil)
-      @default = default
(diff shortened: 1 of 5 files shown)


jdantonio:
Thank you so much @ivoanjo! This is fantastic work. I want to pull the code locally and experiment with it before I merge it in, but I expect I will merge it without changes.

ivoanjo:
Thanks! I was inspired a bit by `Hash` which has a similar API (although without saving the default value after getting it from the block), and it really feels much more natural and high-level to use.

ivoanjo:
Awesome!

PR #569, opened 2016-09-02
Reject promise on any exception

davishmcclurg:
Currently, promises will hang if an exception
is raised that does not extend from StandardError.

```
Concurrent::Promise.execute { raise Exception }.value
...
```

This behavior is unexpected and seems likely to be
wrong for most use cases.

Futures already use `safe_execute` which includes the
`rescue_exception` option.

### DIFF
--- a/lib/concurrent/promise.rb
+++ b/lib/concurrent/promise.rb
@@ -525,7 +525,7 @@ module Concurrent
     # @!visibility private
     def realize(task)
       @executor.post do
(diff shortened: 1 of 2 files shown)


jdantonio:
Nice catch! You are right, this was inconsistent with Future. Thank you for fixing this.

davishmcclurg:
@jdantonio No problem. Thanks for all the hard work.

PR #573, opened 2016-09-13
TimerSet now survives a fork.

nickelser:
This is very similar in spirit to #501: when a fork is detected, immediately:

- Clear all pending jobs from the queue.
- Rest the wait condition.

I found this issue while debugging [a Gem I am working on](https://github.com/nickelser/caddy) which uses a `Concurrent::TimerTask` behind the scenes. If I started any tasks before a fork, they didn't survive, but the bug was that no more `TimerTask`s could be started after that.

This allows new `TimerTask`s to be created after a fork.


jdantonio:
God catch! thank you very much for doing this.

nickelser:
@jdantonio thanks for merging. My pleasure, thanks for maintaining an excellent library.

PR #576, opened 2016-09-22
Use block form of IO.popen to prevent zombies

matthewd:
```
107965 ?        Z      0:00      \_ [nproc] <defunct>
```

😟 

---

The fact we do the fork on any linux with an nproc, and not only the one with an untrustworthy cpuinfo, seems separately unfortunate... but we can at least avoid leaving a mess.

### DIFF
--- a/lib/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent/utility/processor_counter.rb
@@ -86,27 +86,25 @@ module Concurrent
               "select NumberOfLogicalProcessors from Win32_Processor")
             result.to_enum.collect(&:NumberOfLogicalProcessors).reduce(:+)
           elsif File.executable?("/usr/bin/nproc")
-            IO.popen("/usr/bin/nproc --all").read.to_i
+            IO.popen("/usr/bin/nproc --all", &:read).to_i
           elsif File.readable?("/proc/cpuinfo")
             IO.read("/proc/cpuinfo").scan(/^processor/).size
           elsif File.executable?("/usr/bin/hwprefs")
-            IO.popen("/usr/bin/hwprefs thread_count").read.to_i
+            IO.popen("/usr/bin/hwprefs thread_count", &:read).to_i
           elsif File.executable?("/usr/sbin/psrinfo")
-            IO.popen("/usr/sbin/psrinfo").read.scan(/^.*on-*line/).size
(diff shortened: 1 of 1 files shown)


pitr-ch:
Nice catch! Would you know how to avoid the fork?

matthewd:
Basically, do one of the things proposed in the later paragraphs of https://github.com/ruby-concurrency/concurrent-ruby/pull/549 :smile:

Something like the following sounds plausible:

```ruby
          elsif File.readable?("/proc/cpuinfo") and
              (cpuinfo_count = IO.read("/proc/cpuinfo").scan(/^processor/).size) > 0
            cpuinfo_count
          elsif File.executable?("/usr/bin/nproc")
            IO.popen("/usr/bin/nproc --all", &:read).to_i
```

Locally ugly, but does avoid a more substantial rearrangement.

Without an Alpha machine to check it on, though... :confused: (cc @graaff)

graaff:
I don't have access to Alpha machines either, I'm the middle man for our Gentoo Linux Alpha team here, but I can ask them to test patches.

pitr-ch:
@graaff I've updated and rebased the PR, could you try it on Alpha?

graaff:
This change does not address the issues outlined in #549. This basically reverts to the way things were initially, which does not work on Alpha.

pitr-ch:
Ah sorry mi mistake, I had looked only at #549's diff :/ I'll include the explanation to the code as well. Do you know what `RbConfig::CONFIG["arch"]` will return on Alpha?

matthewd:
I think [being more discerning about whether the contents of cpuinfo is usable](https://github.com/ruby-concurrency/concurrent-ruby/pull/576#issuecomment-249226079) will be more general than a straight platform conditional, despite being a little uglier.

If we know it's possible for cpuinfo to exist but not contain any processors, I'm not sure I'd want to sign up for maintaining a blacklist of platforms on which that's true.

graaff:
I've just added #601 based on the suggestion from matthewd (just for the ordering and correct processor_count, not the block form part). Tested on amd64 and alpha (I have an account on our dev box now so I can test things myself if needed).

pitr-ch:
@graaff Thank you! I've incorporated your PR here, when the travis passes I'll merge.

pitr-ch:
Merging, thanks again @matthewd and @graaff.

PR #577, opened 2016-10-04
Make all?, any? return something more meaningful

sheerun:
Currently all of them return `nil`, what can be confusing when someone comes from JS world.

Also, I'm in need of creating `Promise.all?` function what returns *unscheduled* promise of results of all passed promises promises.. The same as `Promise.all` in JS.

Could we maybe add `Promise.all` in concurrent ruby as well?

### DIFF
--- a/lib/concurrent/promise.rb
+++ b/lib/concurrent/promise.rb
@@ -479,6 +479,7 @@ module Concurrent
         unless completed.empty? || completed.send(method){|promise| promise.fulfilled? }
           raise PromiseExecutionError
         end
+        completed
       end
       composite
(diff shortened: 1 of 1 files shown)


pitr-ch:
Hi @sheerun, thank you for your feedback. Could you elaborate on the change in this PR, what does the aggregate return now?

The problem you are describing is solved in new implementation of premisses which will deprecate this one. The new implementation is available in edge. The new promises are scheduled for major API renaming update (no internal changes) in few months in edge, then it will be moved to core.

sheerun:
ok then

PR #579, opened 2016-10-07
Minimal changes to avoid Fixnum deprecation warnings

matthewd:
This does not revise the documentation to match -- I figured that might warrant closer attention to which values are happy to mean "all integers", and which feel obliged to now specify "integers in the range X".

As for the changes: simple counts like `parties` will now accept Bignums -- that doesn't seem particularly useful, but also seems no more problematic than `2**60`. The only one that felt potentially notable was `MutexAtomicFixnum` -- but while more values pass the class check, they're caught by the range check immediately following.

### DIFF
--- a/lib/concurrent/atomic/mutex_count_down_latch.rb
+++ b/lib/concurrent/atomic/mutex_count_down_latch.rb
@@ -9,9 +9,9 @@ module Concurrent
 
     # @!macro count_down_latch_method_initialize
     def initialize(count = 1)
-      unless count.is_a?(Integer) && count >= 0
-        raise ArgumentError.new('count must be in integer greater than or equal zero')
-      end
+      Utility::NativeInteger.ensure_integer_and_bounds count
+      Utility::NativeInteger.ensure_positive count
+
       super()
       synchronize { ns_initialize count }
     end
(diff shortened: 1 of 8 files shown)


jdantonio:
@pitr-ch Can you take a look at this, please?

pitr-ch:
@jdantonio sure.
Thanks @matthewd, it looks good. We have to consider enforcing the small integer check on the other classes though, I plan to optimise some of them and it would help not to have to support Bignum. I'll get back to the PR and I'll merge it for `1.0.(3 or 4)`.

lib/concurrent/atomic/mutex_count_down_latch.rb
```diff
@@ -9,9 +9,9 @@ class MutexCountDownLatch < Synchronization::LockableObject
 
     # @!macro count_down_latch_method_initialize
     def initialize(count = 1)
-      unless count.is_a?(Integer) && count >= 0
-        raise ArgumentError.new('count must be in integer greater than or equal zero')
-      end
+      Utility::NativeInteger.ensure_integer_and_bounds count
```
matthewd: @pitr-ch this is missing a `require` (context: https://github.com/rack/rack/pull/1198)

pitr-ch:
@matthewd thanks for letting me know.

lib/concurrent/atomic/mutex_count_down_latch.rb
pitr-ch: @matthewd thanks for letting me know.

PR #584, opened 2016-10-11
Fix some typos

shotat:
Fixed some typos in ...

- lib/concurrent/actor.rb
- lib/concurrent/async.rb
- lib/concurrent/atom.rb


pitr-ch:
Great thank you!

PR #598, opened 2016-11-05
mark buggy tests; cleanup gitignore

pitr-ch:


spec/concurrent/mvar_spec.rb
```diff
@@ -387,7 +387,8 @@ def m.simulate_spurious_wake_up
           expect(m.put(14)).to eq 14
         end
 
-        it 'returns TIMEOUT on timeout on a full MVar' do
+        it 'returns TIMEOUT on timeout on a full MVar', buggy: true do
```
ianks: As a heads up, in the future this can just be:

```ruby
it 'returns TIMEOUT on timeout on a full MVar', :buggy
```

PR #601, opened 2016-11-07
Avoid forking for processor_count if possible

graaff:
This re-orders the checks for the processor count so that forking is
only done when needed. alpha-linux does not support the processor entry
in /proc/cpuinfo so we must catch that case and fall back to using nproc
which does work on alpha.

This is a rework of #549 based on suggestions made in #576


pitr-ch:
solved in #576

PR #607, opened 2016-11-27
Add Cyclic Barrier example.

anildigital:
Add a simple Cyclic Barrier example in documentation.

### DIFF
--- a/lib/concurrent/atomic/cyclic_barrier.rb
+++ b/lib/concurrent/atomic/cyclic_barrier.rb
@@ -5,6 +5,25 @@ module Concurrent
 
   # A synchronization aid that allows a set of threads to all wait for each
   # other to reach a common barrier point.
+  # @example
+  #   barrier = Concurrent::CyclicBarrier.new(3)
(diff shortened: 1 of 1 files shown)


pitr-ch:
Thanks for the example. I've tried pushed changes to this branch, but maybe you did not allow it. Could you update the example to:
```
  # A synchronisation aid that allows a set of threads to all wait for each
  # other to reach a common barrier point.
  # @example
  #   barrier = Concurrent::CyclicBarrier.new(3)
  #   jobs    = Array.new(3) { |i| -> { sleep i; p done: i } }
  #   process = -> (i) do
  #     # waiting to start at the same time
  #     barrier.wait
  #     # execute job
  #     jobs[i].call
  #     # wait for others to finish
  #     barrier.wait
  #   end
  #   threads = 2.times.map do |i|
  #     Thread.new(i, &process)
  #   end
  #
  #   # use main as well
  #   process.call 2
  #
  #   # here we can be sure that all jobs are processed
```

anildigital:
@pitr-ch Updated PR with your example.

pitr-ch:
@anildigital Thank you!

PR #628, opened 2017-02-03
Output Ruby warnings when running rspec

meineerde:
This implements #588 in that it outputs any warning along with the rspec output. It does not force the test run to fail though.


meineerde:
The AppVeyor failure seems to be unrelated to this change.

PR #629, opened 2017-02-07
Always initialize @default and @default_block for AbstractThreadLocalVar

meineerde:
This avoids Ruby warnings like this when accessing the default value:

> lib/concurrent/atomic/abstract_thread_local_var.rb:57: warning: instance variable @default_block not initialized

This pull request fixes one of the warnings made visible in #628.


pitr-ch:
Thanks a lot! Merging

PR #633, opened 2017-02-26
Documentation/use the correct url in the comments

parasquid:
Some of the comments in the `examples/go-by-example-channels` files have the wrong url.


pitr-ch:
Thanks for the fix!

PR #644, opened 2017-03-29
Fix Map#each and #each_pair not returning enumerator outside of MRI

### LINKED ISSUE #643: Concurrent::Map#each behaves differently in MRI and JRuby
-   Operating system: linux
-   `concurrent-ruby` version: 1.0.5 and master (tried both)
-   `concurrent-ruby-ext` installed: no

### Description:

`Concurrent::Map#each` when called without a block returns an enumerator on MRI but `self` on JRuby.

### Testcase:

```ruby
require 'concurrent'
puts "#{RUBY_ENGINE}: #{Concurrent::Map.new.each.inspect}"
```

Output in MRI 2.4.1:
```
ruby: #<Enumerator: #<Concurrent::Map:0x0055dd74e2dcc0 entries=0 default_proc=nil>:each_pair>
```

Output in JRuby 9.1.8.0:
```
jruby: #<Concurrent::Map:0x00000000000fa0 entries=0 default_proc=nil>
```

ivoanjo:
When no block was passed to Map#each_pair or its alias Map#each, the NonConcurrentMapBackend would return an enumerator, which would allow enumerable methods to be used on a Map.

But the alternate Map backends, JRubyMapBackend and AtomicReferenceMapBackend did not implement the same logic.

As a fix, let's move the logic down to the shared Map class, so all implementations can use it.

Fixes #643

### DIFF
--- a/lib/concurrent/collection/map/non_concurrent_map_backend.rb
+++ b/lib/concurrent/collection/map/non_concurrent_map_backend.rb
@@ -95,7 +95,6 @@ module Concurrent
       end
 
       def each_pair
-        return enum_for :each_pair unless block_given?
         dupped_backend.each_pair do |k, v|
           yield k, v
         end
--- a/lib/concurrent/map.rb
+++ b/lib/concurrent/map.rb
@@ -171,6 +171,11 @@ module Concurrent
       each_pair {|k, v| yield v}
     end unless method_defined?(:each_value)
 
+    def each_pair
+      return enum_for :each_pair unless block_given?
+      super
+    end
+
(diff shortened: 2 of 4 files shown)


ivoanjo:
Hey @pitr-ch any chance this can get released? I just got bitten by it again (and it's been so long I didn't even remember I had already fixed it).

pitr-ch:
Sorry, I am catching up on c-r after busy summer, I'll get a release out.

ivoanjo:
Great, thanks! :+1: :+1: :+1:

PR #649, opened 2017-04-21
Fix promise example in comments

bmulvihill:
Rejection example does not currently work as written.  This fixes the example to have the expected output.

Thanks!


pitr-ch:
Sorry you did not get any response. I am not sure how your PR got missed. The doc you were fixing was indeed faulty, I've fixed it in https://github.com/ruby-concurrency/concurrent-ruby/commit/967f98a3086e077bcf2f6f34cd6dc0ba629b7c9c. Thanks for bringing it to my attention.

PR #651, opened 2017-04-27
#inspect should not output negative object IDs.

voxik:
It might happen that object ID is 'negative', since Ruby tries to avoid Bignums:

https://bugs.ruby-lang.org/issues/13397

However, I am not sure if this does not break any precondition for the fec11e79839ded538f15159933ed2275a1aa28e9 commit which introduced one of the #inspect methods. But I am using Kernel#to_s (as suggested by @nobu) instead of #inspect, so hopefully this is correct ...


pitr-ch:
Thanks @voxik, rebasing and merging when CI passes

PR #654, opened 2017-05-25
Promise.zip execution changes

davishmcclurg:
I added details for each change in the commit messages.
Let me know if you'd like these in separate PRs.

### DIFF
--- a/lib/concurrent/promise.rb
+++ b/lib/concurrent/promise.rb
@@ -387,7 +387,9 @@ module Concurrent
     #
     # @return [Promise<Array>]
     def self.zip(*promises)
-      zero = Promise.new(executor: ImmediateExecutor.new) { [] }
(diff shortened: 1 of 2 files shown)


pitr-ch:
Thanks. The change makes sense but as you said it's a breaking change. Could you introduce the non-executing behavior with an option passed to zip?

davishmcclurg:
@pitr-ch I added an `execute` option and switched the default back to returning a fulfilled promise.

pitr-ch:
@davishmcclurg Thanks for the update, looks good. One last think though, could you update the documentation to mention the new option?

davishmcclurg:
@pitr-ch I finally got around to updating the docs. I believe `@overload` was necessary here, but let me know if there's something cleaner.

pitr-ch:
Thanks looks great! I've rebased the PR to pass the CI. When green I'll merge it.

PR #660, opened 2017-06-06
Add specs for Promise#zip/Promise.zip ordering

ianks:
I noticed there were not any specs on the ordering of `Promise.zip`. Since it is part of the spec (as it applies to `Promise.all`), I figured I could add some tests for that.

For more info: https://stackoverflow.com/a/28066851

### DIFF
--- a/spec/concurrent/promise_spec.rb
+++ b/spec/concurrent/promise_spec.rb
@@ -371,36 +371,19 @@ module Concurrent
 
       it 'preserves ordering of the executed promises' do
         10.times do
-          running = Mutex.new
-          cond = ConditionVariable.new
-          cond2 = ConditionVariable.new
+          latch1 = CountDownLatch.new
+          latch2 = CountDownLatch.new
           executor = SimpleExecutorService.new
 
-          p1 = Concurrent::Promise.execute(executor: executor) do
-            running.synchronize do
-              cond.wait(running)
-              'one'
-            end
-          end
+          p1 = Concurrent::Promise.execute(executor: executor) { latch1.wait; 'one' }
+          p2 = Concurrent::Promise.execute(executor: executor) { latch2.wait; 'two' }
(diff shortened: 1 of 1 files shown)


spec/concurrent/promise_spec.rb
```diff
@@ -368,6 +368,27 @@ def get_ivar_from_args(opts)
 
         expect(composite).to be_rejected
       end
+
+      it 'preserves ordering of the executed promises' do
+        promise1 = Promise.execute do
+          # resolves after the second promise
+          sleep 0.2
```
hobodave: You shouldn't rely on `sleep` for thread synchronization/ordering. Instead use concurrency primites:

```ruby
running = Mutex.new
cond = ConditionVariable.new
cond2 = ConditionVariable.new

p1 = Concurrent::Promise.execute do
  running.synchronize {
    cond.wait(running)
    'one'
  }
end

p2 = Concurrent::Promise.execute do
  running.synchronize {
    cond2.wait(running)
    'two'
  }
end

p3 = Concurrent::Promise.execute do
  running.synchronize {
    'three'
  }
end

cond2.signal
cond.signal

result = Concurrent::Promise.zip(p1, p2, p3).value
expect(result) .to eq(['one', 'two', 'three'])
```

pitr-ch: @hobodave is right. @ianks could you update the spec? I would like to include it since this is indeed the expected behavior.

ianks:
@hobodave @pitr-ch strangely, i am getting deadlocks using the cv implementation in the spec.

pitr-ch:
@ianks could you use `Countdouwnlatch.new(1)` instead? It's what we usually use and it could resolve the deadlock without digging in.

pitr-ch:
I've updated the PR to use count-down-latch and rebased.

PR #665, opened 2017-07-03
Initialize the monitor for new subarrays on Rubinius

### LINKED ISSUE #559: Array#slice! broken on Rubinius
### Expected behavior:

Array#slice! returns a fully-functional Array (intuitively a Concurrent::Array) containing the data removed from the primary array.
### Actual behavior:

Array#slice! returns a near-complete version of Array, but with no valid `@_monitor`, thus causing any methods that require synchronization on the object to fail.
### Example:

``` irb
~ ❯❯❯ uname -a
Linux 4.6.4-1-ARCH #1 SMP PREEMPT Mon Jul 11 19:12:32 CEST 2016 x86_64 GNU/Linux
~ ❯❯❯ ruby --version
rubinius 3.47 (2.3.1 43e411b2 2016-07-15 3.8.0) [x86_64-linux-gnu]
~ ❯❯❯ pry
[1] pry(main)> require 'concurrent/array'
=> true
[2] pry(main)> a = Concurrent::Array.new [1,2,3,4,5,6]
=> [1, 2, 3, 4, 5, 6]
[3] pry(main)> a.instance_variable_get :@_monitor
=> #<Monitor:0x98b30 @mon_count=0, @mon_mutex=#<Mutex:0x98c28 @owner=nil>, @mon_owner=nil>
[4] pry(main)> b = a.slice!(0..2)
=> #<Concurrent::Array:0x4c7d4>
[5] pry(main)> puts b
NoMethodError: undefined method `synchronize' on nil:NilClass.
from core/zed.rb:1425:in `synchronize (method_missing)'
[6] pry(main)> b.instance_variable_get :@_monitor
=> nil
```

---

I honestly have no idea what's going on with this. I tested a few other things, and it looks like `#slice!` and indexing with a range (e.g. `a[0..2]`) both return invalid `Concurrent::Array`s. My first thought from looking in `thread_safe/util/array_hash_rbx.rb` is that `_mon_initialize` isn't being called when it should be, but I don't know.

meineerde:
This pull request should fix #559.

Rubinius uses the public methods `new_range` and `new_reserved` to
create new sub-arrays in methods like `Array#slice`, `Array#[](range)`
or similar.

Here, Rubinius does not call `Array.allocate` to create the new object
resulting in the monitor not being initialized which then causes errors
like

    NoMethodError: undefined method `synchronize' on nil:NilClass.

when calling any methods of the new sub-array. By overwriting the two
methods and explicitly initializing the monitor there, we can ensure
that the monitor is always initialized correctly.

### DIFF
--- a/spec/concurrent/array_spec.rb
+++ b/spec/concurrent/array_spec.rb
@@ -14,5 +14,15 @@ module Concurrent
         end
       end.map(&:join)
     end
+
+    describe '#slice' do
+      # This is mostly relevant on Rubinius and Truffle
+      it 'correctly initializes the monitor' do
+        ary.concat([0, 1, 2, 3, 4, 5, 6, 7, 8])
+
+        sliced = ary.slice!(0..2)
+        expect { sliced[0] }.not_to raise_error
+      end
+    end
   end
 end
--- a/lib/concurrent/thread_safe/util/array_hash_rbx.rb
+++ b/lib/concurrent/thread_safe/util/array_hash_rbx.rb
@@ -18,11 +18,22 @@ module Concurrent
         end
 
         klass.superclass.instance_methods(false).each do |method|
-          klass.class_eval <<-RUBY, __FILE__, __LINE__ + 1
-            def #{method}(*args)
-              @_monitor.synchronize { super }
-            end
-          RUBY
+          case method
+          when :new_range, :new_reserved
+            klass.class_eval <<-RUBY, __FILE__, __LINE__ + 1
+              def #{method}(*args)
+                obj = super
+                obj.send(:_mon_initialize)
+                obj
+              end
+            RUBY
+          else
+            klass.class_eval <<-RUBY, __FILE__, __LINE__ + 1
+              def #{method}(*args)
+                @_monitor.synchronize { super }
+              end
+            RUBY
+          end
         end
       end
     end


meineerde:
For reference, `Array#new_range` is defined in [core/array.rb#L15-L18](https://github.com/rubinius/rubinius/blob/e0790fab81048b2428067967b166c75439ceea65/core/array.rb#L15-L18) resp. [machine/class/array.cpp#L66-L103](https://github.com/rubinius/rubinius/blob/03271d7b97af5ae1a3d64f05c23e0785cf750b28/machine/class/array.cpp#L66-L103).

It is used in e.g. `Array#[]` to construct a new sub-array in [core/array.rb#L307](https://github.com/rubinius/rubinius/blob/e0790fab81048b2428067967b166c75439ceea65/core/array.rb#L307).

spec/concurrent/array_spec.rb
```diff
@@ -14,5 +14,15 @@ module Concurrent
         end
       end.map(&:join)
     end
+
+    describe '#slice' do
+      # This is mostly relevant on Rubinius and Truffle
```
eregon: FWIW TruffleRuby tries to avoid extra public methods, so new_range/new_reserved are not defined on Array in TruffleRuby.
Arguably, this is a Rubinius bug to expose internal methods as public.

meineerde: The problem is not so much that this method is public rather than that they are "special" to Rubinius in that they create objects without calling `allocate`. They would still have to be overwritten by concurrent-ruby if they were private, although it would have to be done slightly different...

As for the comment itself, I have not tested this on Truffle nor have I specifically checked their implementation to be honest. The comment might thus be not 100% exact in this regard. However, since the monitor is only used on Rubinius and Truffle at all, the test still applies to this locking variant used here.

Issue #564, opened 2016-08-02
Thread Safe Set implementation


thedrow:
Currently there is no Thread Safe set implementation in this library and I found a need for it.
What would it take to add one?

pitr-ch:
As a quick dirty solution you could use the Concurrent::Hash's keys as set, inserting always true for values. If you look at stdlib's Set it does the same. We look into adding Set for convenience later.

thedrow:
Is there a reason not to save the memory for the values?

pitr-ch:
Not really, it's just that nobody wrote the Set implementation yet.

jdantonio:
@thedrow Thank you very much for the suggestion!

> What would it take to add one?

I'd be happy to look at a PR which adds this feature to Edge. Unfortunately none of the core maintainers are available to work on this right now. Our top priorities are bugs and the upcoming 2.0 effort. Once 2.0 is complete we can revisit the idea. In the mean time I'm going to pause this issue and leave it open.

PR #666, opened 2017-07-12
thread safe set implementation

haridutt12:


haridutt12:
This is thread safe set implementation along the lines of thread safe Array
fixes #564

haridutt12:
@pitr-ch ping

pitr-ch:
Thank you for your contribution! At first glance it looks good, however I'll have to return to it and give it more thought. Maybe @thedarkone could help. @thedarkone would you have time to help with the review?

spec/concurrent/set_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'set' 
+module Concurrent
+  describe Set do 
+    let!(:set) { described_class.new }
+
+    it 'concurrency' do
+      (1..THREADS).map do |i|
+        Thread.new do
+            set << i
```
thedarkone: Change this whole spec to look like `Hash` spec: https://github.com/ruby-concurrency/concurrent-ruby/blob/df482db36caf1b0c1d69a8ff97a2407469e1e315/spec/concurrent/hash_spec.rb#L8 (by adding another inner `1000.times do |j|` loop, just like for `Hash`).

haridutt12:
@thedarkone can u please review the changes.

lib/concurrent/set.rb
```diff
@@ -0,0 +1,44 @@
...
+    #
+    #   A thread-safe subclass of Set. This version locks against the object
+    #   itself for every method call, ensuring only one thread can be reading
+    #   or writing at a time. This includes iteration methods like `#each`.
+    #
+    #   @note `a += b` is **not** a **thread-safe** operation on
+    #   `Concurrent::Set`. It reads Set `a`, then it creates new `Concurrent::Set`
+    #   which is concatenation of `a` and `b`, then it writes the concatenation to `a`.
+    #   The read and write are independent operations they do not form a single atomic
+    #   operation therefore when two `+=` operations are executed concurrently updates
```
eregon: Unfinished sentence: `updates` ... `may be lost` from https://github.com/ruby-concurrency/concurrent-ruby/blob/df482db36caf1b0c1d69a8ff97a2407469e1e315/lib/concurrent/array.rb#L21

spec/concurrent/set_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+module Concurrent
+  describe Set do 
+    let!(:set) { described_class.new }
+
+    it 'concurrency' do
+      (1..THREADS).map do |i|
+        Thread.new do
+           1000.times do |j|
+            set << i 
+            set.empty?
```
ianks: This spec does not make set any expectations. Can you clarify what you are testing and make expectations about the behavior in the spec?

haridutt12: @ianks actually this is my first ever contribution so i dont have much knowledge about minitests so can u plese help me out in this?

pitr-ch:
@thedarkone @ianks @eregon Thanks you for helping with the review!

spec/concurrent/set_spec.rb
pitr-ch: Actually the tests follow what we have for Array and Hash. I'll improve all of them at least a bit.

pitr-ch:
I've improved the tests a bit, updated documentation, squashed and rebased. After CI passes it can be merged.

PR #679, opened 2017-11-18
add missing require

pitr-ch:


jdantonio:
All of the failing tests seem to be a deadlock with `Actor`. All are Java builds and the last successful test is `forbids Immediate executor` after which the tests timeout.

PR #685, opened 2017-12-08
Avoid RSpec warnings about raise_error

olleolleolle:
This PR **avoids an RSpec warning** about `raise_error` without any arguments.

While fixing this output warning, I found that one test which had been using a bare expectation to `raise_error` actually masked a NameError, just like the RSpec warning said it could.


olleolleolle:
This fix does not change the status of the JRuby issues - those builds stall after 

```
Concurrent::Actor

  forbids Immediate executor
```

pitr-ch:
Rebased on master to rerun CI.

PR #686, opened 2017-12-08
Avoid RSpec warnings about unused variables

olleolleolle:
This PR avoids a few RSpec warnings about unused variables.


pitr-ch:
Thank you! I've rebased and when it passes I'll merge.

PR #700, opened 2018-02-25
Fixing CI

### LINKED ISSUE #681: Travis tests are failing on JRuby

pitr-ch:
fix #681

### DIFF
--- a/lib/concurrent/atomic/count_down_latch.rb
+++ b/lib/concurrent/atomic/count_down_latch.rb
@@ -55,7 +55,7 @@ module Concurrent
   # @!macro internal_implementation_note
   CountDownLatchImplementation = case
                                  when Concurrent.on_jruby?
-                                   JavaCountDownLatch
+                                   MutexCountDownLatch
                                  else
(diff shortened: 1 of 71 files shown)


lib/concurrent/atomic/count_down_latch.rb
```diff
@@ -55,7 +55,7 @@ module Concurrent
   # @!macro internal_implementation_note
   CountDownLatchImplementation = case
                                  when Concurrent.on_jruby?
-                                   JavaCountDownLatch
```
kares: hey, could we get some details on why the 'native' JRuby stuff, seems to be going away?
understand the limited time for maintenance but still seems weird these had to go ...

pitr-ch: Sure, please have a look at #717. I am updating the changelog before release, and when the 1.0.6.pre goes out it will mention this as well.

pitr-ch: Please ask any follow up questions there.

PR #706, opened 2018-05-03
Update ruby versions in CI

ianks:
This PR will get travis up to the latest Ruby versions as noted from: https://www.ruby-lang.org/en/downloads/releases/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,14 @@ language: ruby
 
 rvm:
   # start with the latest
-  - 2.5.0
+  - 2.5.1
   - jruby-9.1.17.0
 
   # older versions
-  - 2.4.3
-  - 2.3.4
-  - 2.2.7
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,14 +2,14 @@ language: ruby
...
 
   # older versions
-  - 2.4.3
-  - 2.3.4
-  - 2.2.7
-  - 2.1.9
+  - 2.4.4
+  - 2.3.7
+  - 2.2.10
+  - 2.1.10
```
gfx: FYI: you can omit the patchlevels, i.e. `2.4` refers the latest version of 2.4.x.

pitr-ch: I know it's more work but I rather keep doing it manually so I or others don't get failure on random PR but rather on the on updating the versions. Thanks for the hint though.

PR #708, opened 2018-05-03
Prefer platform specific memory barriers

ianks:
Following the wisdom of postgres, we opt to use platform specific memory barriers when available. These are generally more performant. In this PR, we add specific cases for i386, x86_64.

In the future, we could look at using pg's atomics library directly: https://github.com/postgres/postgres/tree/9d4649ca49416111aee2c84b7e4441a0b7aa2fac/src/include/port/atomics

### DIFF
--- a/ext/concurrent/atomic_reference.c
+++ b/ext/concurrent/atomic_reference.c
@@ -9,6 +9,26 @@
 
 #include "atomic_reference.h"
 
+#if (defined(__i386__) || defined(__i386)) && (defined(__GNUC__) || defined(__INTEL_COMPILER))
+#define memory_barrier() \
+  __asm__ __volatile__ ("lock; addl $0,0(%%esp)" : : : "memory", "cc")
+#elif defined(__x86_64__) && (defined(__GNUC__) || defined(__INTEL_COMPILER))
+#define memory_barrier() \
+  __asm__ __volatile__ ("lock; addl $0,0(%%rsp)" : : : "memory", "cc")
+#elif defined(HAVE_GCC__ATOMIC_INT32_CAS)
+#define memory_barrier() \
+  __atomic_thread_fence(__ATOMIC_SEQ_CST)
+#elif (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 1))
+#define memory_barrier() \
+  __sync_synchronize();
+#elif defined _MSC_VER
+#define memory_barrier() \
+  MemoryBarrier();
+#elif __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1050
+#define memory_barrier() \
+  OSMemoryBarrier();
+#endif
+
 void ir_mark(void *value) {
   rb_gc_mark_maybe((VALUE) value);
 }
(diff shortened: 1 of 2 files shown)


thedarkone:
GCC emits `mfence` for `__sync_synchronize()` (so does Clang), I don't know what MS compiler does (presumably the same). Per Shipilev's [findings](https://bugs.openjdk.java.net/browse/JDK-8050147#comment-13521925) using the proposed asm should result in about 15 % speed up for the synch. instruction in question. I don't think the improvement is measurable with the general overhead that usually comes with MRI. @ianks, have you done any benching?

If this is to be merged the PR should also clean up `extconf.rb`: https://github.com/ruby-concurrency/concurrent-ruby/blob/b461ffcb0876110402c7c6dd54bd982b5e1096a4/ext/concurrent/extconf.rb#L46 since `HAVE_GCC_SYNC` doesn't seem to be used anywhere else.

I would consider merging this if and only if there was a measurable difference on MRI.

ianks:
@thedarkone Just did some benchmarking on x86 (do not have another other machines at the moment). 

### AtomicBoolean#true?

Pessimistic results (taking slowest result from this branch, fastest from master), show 7.89314% improvement. The optimistic (this branch best, master worst) was 14.8174% improvement

### AtomicBoolean#make_true/make_false

Pessimistic results (taking slowest result from this branch, fastest from master), show 7.19176% improvement. The optimistic (this branch best, master worst) was 11.5249%% improvement

```ruby
require 'benchmark/ips'
require 'concurrent'

bool = Concurrent::AtomicBoolean.new(false)
`bundle exec rake compile`

# master
# ==========================================================================
#               #true?      9.922M (± 2.7%) i/s -     49.768M in   5.020095s
#               #true?      9.675M (± 1.2%) i/s -     48.484M in   5.011882s
#               #true?      9.550M (± 4.4%) i/s -     47.841M in   5.021202s
# #make_true/#make_false    7.469M (± 2.7%) i/s -     37.386M in   5.009572s
# #make_true/#make_false    7.386M (± 2.5%) i/s -     37.123M in   5.029803s
# #make_true/#make_false    7.398M (± 2.2%) i/s -     37.123M in   5.021120s

# update-memory-barriers
# ==========================================================================
#               #true?     11.098M (± 1.9%) i/s -     55.497M in   5.002488s
#               #true?     10.891M (± 1.4%) i/s -     54.673M in   5.021049s
#               #true?     10.766M (± 5.6%) i/s -     53.849M in   5.021705s
# #make_true/#make_false   8.289M (± 2.2%) i/s -     41.663M in   5.029083s
# #make_true/#make_false   8.208M (± 2.2%) i/s -     41.167M in   5.018527s
# #make_true/#make_false   8.047M (± 4.7%) i/s -     40.175M in   5.006055s
Benchmark.ips do |x|
  x.config(iterations: 3)

  x.report('#true?') do
    bool.true?
  end

  x.report('#make_true/#make_false') do
    bool.make_true
    bool.make_false
  end
end
```

ianks:
@pitr-ch Are there any other changes you would like to see here?

pitr-ch:
Thanks for reminding me about this. Looks good. Could you put a comment in the code explaining when it comes from, as you did int the PR's description. Then we can merge.

pitr-ch:
@thedarkone thank you for your input on the PR!

PR #729, opened 2018-06-08
Fix `Promise#then`.

reitermarkus:
The docs say it should work like `then(executor: executor)`.

### DIFF
--- a/lib/concurrent/promise.rb
+++ b/lib/concurrent/promise.rb
@@ -307,7 +307,16 @@ module Concurrent
     # @yield The block operation to be performed asynchronously.
     #
     # @return [Promise] the new promise
-    def then(rescuer = nil, executor = @executor, &block)
+    def then(*args, &block)
+      if args.last.is_a?(::Hash)
+        executor = args.pop[:executor]
+        rescuer = args.first
+      else
+        rescuer, executor = args
+      end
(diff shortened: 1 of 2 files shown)


pitr-ch:
Hi, thanks for the fix. However the docs need to be fixed instead at the moment. The change as is would be backward incompatible. (Even though I would also like to use named arguments.)

reitermarkus:
Maybe we could support both in order to not break compatibility?

pitr-ch:
yeah, that's a good plan, however concurrent-ruby has to drop support for 1.9 officially first to be able to do it. So I won't be able to merge this PR until before next minor release.

reitermarkus:
> concurrent-ruby has to drop support for 1.9 officially first

Why? There are other methods using named parameters.

reitermarkus:
@pitr-ch, ping.

pitr-ch:
Sorry I was away. I think they are using last argument as an options Hash, named arguments should not be used. If you saw named arguments please point me to it.

reitermarkus:
> I think they are using last argument as an options Hash

Yes, I just meant that they can be used with named parameters, not that they are implemented with them.

Anyways, I implemented it to support both the current and the version with an option Hash.

PR #732, opened 2018-06-22
Fixnum dep. warning in CountDownLatch

theturtle32:
Resolves this warning:

gems/jruby-9.2.0.0/gems/concurrent-ruby-1.0.5-java/lib/concurrent/atomic/java_count_down_latch.rb:12: warning: constant ::Fixnum is deprecated


pitr-ch:
Hi @theturtle32, thanks for the fix! I am sorry I cannot merge though, I already had a fix in https://github.com/ruby-concurrency/concurrent-ruby/commit/ec66f05ce926eea7393e4a0a8c4e0b1cc15efeba which is more appropriate.

Issue #720, opened 2018-05-23
Add TruffleRuby to CI


pitr-ch:

eregon:
@pitr-ch Please coordinate with me ;)

https://github.com/ruby/mspec/blob/297309c05b89e1a964595d6b8d67a01fd62966e0/.travis.yml#L23-L29
shows one way to run TruffleRuby in CI.
Starting from the upcoming 1.0.0-rc3 release, running the post-install should not be needed on Travis CI.

https://github.com/travis-ci/travis-ci/issues/9803 tracks support for `rvm: truffleruby` to just work, but I don't know when that will be supported.

eregon:
https://github.com/oracle/truffleruby/blob/master/doc/user/standalone-distribution.md now documents how to test TruffleRuby in TravisCI

PR #733, opened 2018-06-23
Update travis CI adding latest Ruby and JRuby versions; fix failure on JRuby 9.2

ivoanjo:
Hello there, here's a small addition!

I also took a stab at adding truffleruby which is now available in rvm master but was not able to force an rvm upgrade before Travis attempts to install the target Ruby, so I guess we'll need to wait until the next rvm release.

Note: Tests seem to be broken with JRuby 9.2.0.0, I'll submit a separate issue for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ rvm:
   - jruby-1.7.27
 
   - ruby-head
+  - ruby-2.6.0-preview2
   - jruby-head
 
   - rbx-3
@@ -42,6 +43,7 @@ matrix:
     - rvm: jruby-head
     - rvm: 1.9.3
     - rvm: rbx-3
+    - rvm: ruby-2.6.0-preview2
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: ruby
 rvm:
   # start with the latest
   - 2.5.1
-  - jruby-9.1.17.0
+  - jruby-9.2.0.0
 
   # older versions
   - 2.4.4
@@ -13,6 +13,7 @@ rvm:
   - 2.0.0
   - 1.9.3
 
+  - jruby-9.1.17.0
   - jruby-9.0.5.0
   - jruby-1.7.27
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,10 +13,13 @@ rvm:
   - 2.0.0
   - 1.9.3
 
+  - jruby-9.2.0.0
```
pitr-ch: `jruby-9.2.0.0` should be on line 6, could you move it there please.

.travis.yml
```diff
@@ -13,10 +13,13 @@ rvm:
...
+  - jruby-9.1.17.0
   - jruby-9.0.5.0
   - jruby-1.7.27
 
   - ruby-head
+  - ruby-2.6.0-preview2
```
pitr-ch: please add it to allowed failure

pitr-ch:
Thanks for trying truffleruby. I have an #720 for it. You might want to leave it up to me, since I can more easily solve any issues there. I hope that it will be straight forward after I'll make sure it works for c-r.

ivoanjo:
@pitr-ch fixed, thanks for the feedback!

pitr-ch:
@ivoanjo would you be interested to look into why is Travis failing for JRuby 9.2?

ivoanjo:
@pitr-ch I already did! ;)

It's caused by https://github.com/jruby/jruby/issues/5229, which has already been fixed upstream in https://github.com/jruby/jruby/pull/5231.

I was not sure if it was worth adding it to `allow_failures` or to submit a CR to change `concurrent-ruby` to avoid the mutation; let me know if you'd like any of those and I'll gladly submit a PR.

pitr-ch:
@ianks cool and thanks! Since people will use JRuby 9.2 anyway could you fix it in concurrent-ruby so it does not fail for the users even though it's JRuby's bug.

ivoanjo:
I've added created the workaround and added it to this PR (rather than creating a separate one), so that we can merge with a clean travis.

PR #734, opened 2018-06-24
Fix Array/Hash/Set construction broken on TruffleRuby

ivoanjo:
While trying to get the [Persistent-💎](https://gitlab.com/ivoanjo/persistent-dmnd) gem specs to work on TruffleRuby, I discovered that Array/Hash/Set construction was broken in several cases.

I've added test cases for the expected behaviors of Array/Hash/Set `.[]` and `.new`, and fixed a number of broken test cases on TruffleRuby (all specs are green on MRI).

There is one remaining spec in `array_spec.rb` that was already broken and that is non-trivial to fix, so I'll create an issue for it.

### DIFF
--- a/lib/concurrent/array.rb
+++ b/lib/concurrent/array.rb
@@ -2,7 +2,8 @@ require 'concurrent/utility/engine'
 require 'concurrent/thread_safe/util'
 
 module Concurrent
-  if Concurrent.on_cruby?
+  case
+  when Concurrent.on_cruby?
 
     # Because MRI never runs code in parallel, the existing
     # non-thread-safe structures should usually work fine.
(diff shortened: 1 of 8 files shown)


eregon:
@pitr-ch Please take a look :)

pitr-ch:
Thanks @ivoanjo for the fix and the tests! I'll have a in depth look soon.

pitr-ch:
I've updated the PR with a different solution which I've already had locally. It should be simpler since it is not needed to look up all the Array and Hash methods which create new instance with a special behavior without initialize, allocate, etc.

ivoanjo:
Nice! I also think the previous approach was rather brittle and could easily break in newer versions.

Let me know if I can help with anything else! TruffleRuby without concurrent-ruby is like going to Disneyland on a rainy day ;)

PR #735, opened 2018-06-24
Fix wrong expected exception in spec assertion

ivoanjo:
I've fixed the spec and re-enabled running on travis as otherwise it seems harmless (famous last words?).


ivoanjo:
@pitr-ch updated! :rocket:

PR #738, opened 2018-07-06
Support keyword arguments on all Concurrent Struct classes

### LINKED ISSUE #694: [Feature Request] Keyword parameters for MutableStruct in Ruby 2.5.0
Ruby 2.5.0 introduced Struct keyword parameters by adding "keyword_init: true" optional parameter. 
Mutable / ImmutableStruct / SettableStruct don't have support for those.

numeraltwo:
📓 Original issue: https://github.com/ruby-concurrency/concurrent-ruby/issues/694

⚠️ Putting this up for initial comments, there's still one tiny bit to address which I will complete shortly...

(fix #694)

### DIFF
--- a/lib/concurrent/immutable_struct.rb
+++ b/lib/concurrent/immutable_struct.rb
@@ -71,20 +71,20 @@ module Concurrent
     end
 
     # @!macro struct_new
-    def self.new(*args, &block)
+    def self.new(*args, **kw_args, &block)
       clazz_name = nil
       if args.length == 0
         raise ArgumentError.new('wrong number of arguments (0 for 1+)')
       elsif args.length > 0 && args.first.is_a?(String)
         clazz_name = args.shift
       end
-      FACTORY.define_struct(clazz_name, args, &block)
+      FACTORY.define_struct(clazz_name, args, kw_args, &block)
     end
 
     FACTORY = Class.new(Synchronization::LockableObject) do
-      def define_struct(name, members, &block)
+      def define_struct(name, members, kw_args, &block)
         synchronize do
-          Synchronization::AbstractStruct.define_struct_class(ImmutableStruct, Synchronization::Object, name, members, &block)
+          Synchronization::AbstractStruct.define_struct_class(ImmutableStruct, Synchronization::Object, name, members, kw_args, &block)
         end
       end
     end.new
--- a/lib/concurrent/mutable_struct.rb
+++ b/lib/concurrent/mutable_struct.rb
@@ -197,20 +197,20 @@ module Concurrent
     end
 
     # @!macro struct_new
-    def self.new(*args, &block)
+    def self.new(*args, **kw_args, &block)
       clazz_name = nil
       if args.length == 0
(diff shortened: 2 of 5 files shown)


pitr-ch:
Awesome! Let me know when I can have a look.

numeraltwo:
@pitr-ch 👌 – this would definitely make MRI > 2 and other analogues a requirement – otherwise `SyntaxError` due to `unexpected tPOW`.

Also if all looks good I'll squash the commit history 👍

pitr-ch:
I've just merged https://github.com/ruby-concurrency/concurrent-ruby/pull/740, could you rebase on master? Squashing is not necessary.

numeraltwo:
@pitr-ch all set.

lib/concurrent/immutable_struct.rb
```diff
@@ -71,20 +71,20 @@ def select(&block)
     end
 
     # @!macro struct_new
-    def self.new(*args, &block)
+    def self.new(*args, **kw_args, &block)
```
pitr-ch: It would be better to have the signature more specific `def self.new(*args, keyword_init: false, &block)`. Afaik no other keyword argument has any effect.

lib/concurrent/mutable_struct.rb
```diff
@@ -197,20 +197,20 @@ def []=(member, value)
     end
 
     # @!macro struct_new
```
pitr-ch: Please update the [`struct_new`](https://github.com/ruby-concurrency/concurrent-ruby/blob/e14ccb3718a256338e55adc6a71970d73c639bfe/lib/concurrent/mutable_struct.rb#L13) definition to include documentation about `keyword_init` option.

lib/concurrent/synchronization/abstract_struct.rb
```diff
@@ -6,9 +6,18 @@ module Synchronization
...
 
       # @!visibility private
-      def initialize(*values)
+      def initialize(*values, **kw_values)
         super()
-        ns_initialize(*values)
+        ns_initialize(*values, **kw_values)
+      end
+
+      # @!macro [attach] struct_keyword_init
```
pitr-ch: The macro definition at the beginning of the can doc can be omitted, if the doc is not shared with other methods.

lib/concurrent/synchronization/abstract_struct.rb
```diff
@@ -6,9 +6,18 @@ module Synchronization
...
+      #
+      #   Returns `true` if the struct uses keyword arguments.
+      #
+      #   @return [Boolean] true if struct uses keyword arguments
+      def keyword_init?
+        self.class::KEYWORD_INIT
       end
```
pitr-ch: The `keyword_init?` on struct instance probably has not usage other then the initialization, I think it can be omitted or moved to the struct class replacing the KEYWORD_INIT constant.

pitr-ch:
@numeraltwo friendly reminder :)

chrisseaton:
Closing both due to inactivity, and also because there's a proper way to delegate keyword arguments now. Please do re-engage if you still want to work on it.

PR #749, opened 2018-07-13
TimerTask: kill the execution when it exceeds the task timeout 

### LINKED ISSUE #639: Thread leaks in TimerTask for long-executing jobs
For long-executing jobs and small `execution_interval` values, `TimerTask` will leak threads at the rate of `execution_interval`.

`timeout_interval`, whilst documented, doesn't seem to be referenced anywhere. There seems to be an old (but unmerged) PR (https://github.com/ruby-concurrency/concurrent-ruby/pull/526) that at least solves the problem of timeouts running at the rate of `execution_interval` instead of `timeout_interval`.

But that still wouldn't solve the problem of a long-running task still being rescheduled every `timeout_interval`, even if it couldn't be explicitly killed off. Nor can `TimerTask` be run without a timeout interval, if you were happy to just let tasks run until they're complete and reschedule them for another period afterwards.

rubemz:
Before this commit when a TimerTask task exceeded the timeout the job kept running. That could lead to thread leaking. Related to #639

This PR changes the TimerTask executor to be a SingleThreadExecutor. So whenever the task doesn't finish in time, the execution thread is killed (in this case the pool is killed).

The spec ensures that this new behavior is working correctly. It fails on master. If the main job isn't killed after the timeout, the `latch.wait(2)` would raise, since the timer task sleeps for 5 seconds.

If you want to see the leak by yourself take a look on this https://github.com/rubemz/concurrent-ruby-leak, which is pretty much a copy from https://github.com/ruby-concurrency/concurrent-ruby/issues/639#issuecomment-309792568. @shun0309 thanks for that code.

Closes #639

### DIFF
--- a/lib/concurrent/timer_task.rb
+++ b/lib/concurrent/timer_task.rb
@@ -280,7 +280,8 @@ module Concurrent
       self.execution_interval = opts[:execution] || opts[:execution_interval] || EXECUTION_INTERVAL
       self.timeout_interval = opts[:timeout] || opts[:timeout_interval] || TIMEOUT_INTERVAL
       @run_now = opts[:now] || opts[:run_now]
-      @executor = Concurrent::SafeTaskExecutor.new(task)
+      @task = task
+      @executor = Concurrent::RubySingleThreadExecutor.new()
       @running = Concurrent::AtomicBoolean.new(false)
       @value = nil
 
@@ -309,13 +310,26 @@ module Concurrent
     def execute_task(completion)
       return nil unless @running.true?
       ScheduledTask.execute(timeout_interval, args: [completion], &method(:timeout_task))
-      _success, value, reason = @executor.execute(self)
+      @thread_completed = Concurrent::Event.new
+      @value = @reason = nil
+
+      @executor.post do
+        begin
+          @value = @task.call(self)
+        rescue Exception => ex
+          @reason = ex
+        ensure
+          @thread_completed.set
+        end
+      end
+
+      @thread_completed.wait
+
       if completion.try?
-        self.value = value
         schedule_next_task
         time = Time.now
         observers.notify_observers do
-          [time, self.value, reason]
+          [time, self.value, @reason]
         end
       end
       nil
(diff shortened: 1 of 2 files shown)


lib/concurrent/timer_task.rb
```diff
@@ -325,6 +339,10 @@ def execute_task(completion)
     def timeout_task(completion)
       return unless @running.true?
       if completion.try?
+        @executor.kill
+        @executor.wait_for_termination
+        @executor = Concurrent::RubySingleThreadExecutor.new()
```
fabiokr: Maybe just `new`?

boone: (also line 284)

pitr-ch:
Thanks for the fix, I'll have a look in couple of days.

lib/concurrent/timer_task.rb
rubemz: 61e0074

pitr-ch:
Hi, sorry I did not forgot about your PR. I needed to get out 1.1.0.pre1 for testing. I'll review the PR soon.

lib/concurrent/timer_task.rb
```diff
@@ -280,7 +280,8 @@ def ns_initialize(opts, &task)
       self.execution_interval = opts[:execution] || opts[:execution_interval] || EXECUTION_INTERVAL
       self.timeout_interval = opts[:timeout] || opts[:timeout_interval] || TIMEOUT_INTERVAL
       @run_now = opts[:now] || opts[:run_now]
-      @executor = Concurrent::SafeTaskExecutor.new(task)
+      @task = task
+      @executor = Concurrent::RubySingleThreadExecutor.new
```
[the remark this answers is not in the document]
rubemz: A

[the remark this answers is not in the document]
rubemz: @wjordan thanks. fixed in 9e2fd1c

pitr-ch:
Sorry still did not find time to do a proper review. There is a problem with killing threads, if it's killed while running e.g. ensure block it may corrupt data and other unpleasant things (more info e.g. https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/). I need to give this a more careful thought.

PR #752, opened 2018-07-20
CI: ruby-head with --jit enabled

olleolleolle:
This adds another CI target: Ruby trunk with the `--jit` option enabled.

This PR was inspired by https://github.com/puma/puma/pull/1559/


pitr-ch:
Thanks for the idea, however, I am afraid that nobody would check the results and we would just be wasting resources.

Issue #736, opened 2018-07-04
Feature Request: AtomicInteger


herwinw:
Ruby has deprecated the Fixnum class in favour of a more generic Integer class ([example blog post](https://blog.bigbinary.com/2016/11/18/ruby-2-4-unifies-fixnum-and-bignum-into-integer.html)). The class `Concurrent::AtomicFixnum` still uses the Fixnum limits. This means the maximum value of this class depends on the architecture of the system.

At least when using MRI this doesn't make much sense anymore (not sure about C extensions, JRuby or Rubinius). It looks pretty trivial to create a `AtomicInteger` object, just remove the call to `NativeInteger.ensure_integer_and_bounds`. It could get `min` and `max` keyword arguments to specify new limits (which would make it possible to create a n`AtomicFixnum` based on `AtomicInteger` with the current boundaries and remove the code duplication)

I think it's pretty simple trivial to implement, so if the team agrees to this I'm willing to create the pull request for it

pitr-ch:
yeah sounds like a good idea, and I appreciate your offer to help!

Few thoughts:
* Could you add constants to `AtomicFixnum` clearly stating what its max and min value, just taking numbers from `NativeInteger`
* Has to be backward compatible (which you already have in mind).
* I would not add max and min option. Let's keep the API simple.

eregon:
To be efficient, an "atomic integer" has to be able to use compare-and-swap, and that's only available as an assembly instruction for the platform's bitwidth (sometimes for the double of it too, but that's not portable).
Also integers which don't fit in Fixnum are basically slow for any operation in Ruby, no matter what.

So Fixnum is actually the correct requirements for an efficient implementation of an "atomic integer".

If one needs an "atomic integer" with larger values than Fixnum allows, then they can use AtomicReference.

The name may be weird, but aliasing `AtomicFixnum` to `AtomicInteger` seems misleading about the limits. So I'm thinking it's good as it is.

PR #756, opened 2018-08-06
Atomic integer

herwinw:
The first version of my proposed patch for #736 I'd rather have a review on this before I continue.

Known issues:
* Only pure Ruby implementations, C and Java implementations should be done
* Very much a copy of the current `AtomicFixnum`, should share code.


herwinw:
Note to self: don't offer to do something right before leaving on vacation ;)

pitr-ch:
@herwinw thanks for the PR, looks good. Would you address the known issues? Then I'll have a look again. Sorry for late answer.

herwinw:
Both of the issues are addressed in the branch. However, I do have a slight problem with the current solution to reduce the duplicated code. It uses inheritance, with `AtomicFixnum < AtomicInteger`. This means code like this may break:

```ruby
if val.is_a?(Concurrent::AtomicInteger)
  val.value = Concurrent::Utility::NativeInteger::MAX_VALUE + 1
end
```

This code will work if `val` is an `AtomicInteger`, but raise an exception on the second line if it is an `AtomicFixnum`. Besides, the C and Java implementations need a lot of restructuring before we can remove any duplicated code. I think this attempt does more harm than good, and should better be removed.

Currently there is some comparable code between the different Atomic things (e.g. `MutexAtomicBoolean` and `MutexAtomicFixnum`), any attempt at removing code duplication should probably work at a more generic level.

chrisseaton:
Thanks for this code. We've been going back and triaging old PRs and issues. We're happy with the general idea of an atomic non-fixnum. This PR is basically a wrapper around atomic reference, which makes sense, as non-fixnum integers are implemented as references. I think we'd encourage people to use atomic reference therefore, like they would with any other reference value. Could this wrapper be implemented in Ruby? We definitely want less C and Java code, but this PR currently adds more.

We're open to doing something more in the future, but since this is an old PR, with conflicts, without a clear pressing user requesting now, we're going to close for the moment as we try to get down to zero PRs.

If you want to come back to us with a pressing need, or a willingness to update the branch and talk about the relation to atomic reference, we're very open to doing that. The underlying issue #736 remains open.

PR #757, opened 2018-08-08
use alpha vantage to extract stock information

GustavoCaso:


pitr-ch:
I think it would be better for users to have an example which can be executed immediately without any additional setup. Not sure where would we put the API_KEY, it should not be public probably. The examples can also use any API, not just about stock. If there is something similar we could dig from GitHub, it might be simpler.

pitr-ch:
The scala parts can be dropped.

pitr-ch:
no worries! Thanks.

PR #761, opened 2018-08-31
make the pruning of thread pools a public API

### LINKED ISSUE #652: expose the prune operation from thread pools
I'm specifically interested in the `CachedThreadPool` on MRI, but I assume that other thread pools/implementations probably have the same issue.

`RubyCachedThreadPool` prunes itself only on execute. I'm having an execution flow that creates a lot of threads in a short period of time and then stops using the thread pool altogether for a long period. This leads to having a few 100s of threads alive that are unused, and breaks the pruning functionality altogether.

I'm currently periodically calling `ns_prune_pool` from an event loop. However, this is private API, and would prefer having it exposed as public API. Would you consider a PR that does that ? I guess that implementation which take care of the pruning (e.g. Java) would simply expose a no-op.

******
-   Operating system: linux
-   `concurrent-ruby-ext` installed: yes

doudou:
This makes the pruning of thread pools a public API, which may
or may not be a no-op, depending on the underlying thread pool
implementation.

It also reworks the Ruby implementation to be more aggressive in
the pruning - it will kill all to-be-pruned threads in one call
instead of doing them one by one. Moreover, workers do not get
woken up to check for their idleness anymore, the thread of the
caller does it instead.

Closes #652

### DIFF
--- a/lib/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent/executor/ruby_thread_pool_executor.rb
@@ -87,13 +87,8 @@ module Concurrent
     end
 
     # @!visibility private
-    def ready_worker(worker)
-      synchronize { ns_ready_worker worker }
-    end
-
-    # @!visibility private
-    def worker_not_old_enough(worker)
-      synchronize { ns_worker_not_old_enough worker }
+    def ready_worker(worker, last_message)
+      synchronize { ns_ready_worker worker, last_message }
     end
 
     # @!visibility private
@@ -104,6 +99,11 @@ module Concurrent
     # @!visibility private
     def worker_task_completed
       synchronize { @completed_task_count += 1 }
+    end
+
+    # @!macro thread_pool_executor_method_prune_pool
+    def prune_pool
+      synchronize { ns_prune_pool }
     end
 
     private
@@ -185,7 +185,7 @@ module Concurrent
     # @!visibility private
     def ns_assign_worker(*args, &task)
       # keep growing if the pool is not at the minimum yet
-      worker = (@ready.pop if @pool.size >= @min_length) || ns_add_busy_worker
+      worker, _ = (@ready.pop if @pool.size >= @min_length) || ns_add_busy_worker
       if worker
         worker << [task, args]
         true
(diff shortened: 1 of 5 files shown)


pitr-ch:
Thanks for the PR! I'll have a careful look. I'll also need to look into the java part, I'd like the behavior to match.

doudou:
Ping ?

OunmanNgampeerapong:
O

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -152,7 +149,7 @@ def ns_execute(*args, &task)
         handle_fallback(*args, &task)
       end
 
-      ns_prune_pool if @next_gc_time < Concurrent.monotonic_time
+      ns_prune_pool
```
pitr-ch: Unfortunately, this means that it will do even more work on every task posting. I'd rather see it removed completely from the hot path (task posting). Would having pruning scheduled externally twice per idle-time be enough? Is there a reason why do you need this to be more aggressive. The external scheduling would get rid of the problem where it does not prune after a spike when no more tasks are posted.

lib/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -268,12 +257,17 @@ def ns_remove_busy_worker(worker)
...
+      stopped_workers = 0
+      while !@ready.empty? && (@pool.size - stopped_workers > @min_length)
+        worker, last_message = @ready.first
+        if now - last_message > self.idletime
+          stopped_workers += 1
+          @ready.shift
+          worker << :stop
+        else break
+        end
+      end
```
pitr-ch: 👍

lib/concurrent/executor/ruby_thread_pool_executor.rb
doudou: I didn't think it would be a problem.

I guess external scheduling would be fine, but doesn't that mean running a separate thread per thread pool to use as timer ? This seems to be quite a bit of additional complexity ...

In fine, I would be personally fine with keeping the line as it was.

doudou: I've re-established pruning at @next_gc_time intervals as it was.

doudou:
@pitr-ch updated. Sorry for taking so long.

pitr-ch:
@doudou thank you. And no worries, I appreciate you've found time to contribute.

doudou:
Gents ... any chance this could be included ?

chrisseaton:
We'll see if we can merge this.

PR #765, opened 2018-10-09
Use supported TruffleRuby APIs

chrisseaton:


pitr-ch:
I've updated TR version to rc7.

chrisseaton:
Did you need to update to RC 7? I didn't understand the failures. Are any of them an issue?

I added shims both ways - TruffleRuby still provides the old methods, and Concurrent Ruby defines the new methods using the old ones if the new ones are missing. Should avoid anyone being inconvenienced.

pitr-ch:
Sorry for confusion. I just needed updating. The changes are fine. The failures are rather result of badly written c-r tests, it's too sensitive to timing.

PR #769, opened 2018-11-01
Put 1.9.3 support back

### LINKED ISSUE #768: 1.1.0 is a breaking build but wasn't marked as such
1.1.0 was released, but it requires ruby `>= 2.0.0`. Previously (in 1.0.5), it was marked as `>= 1.9.3`.

This change is breaking for many CI systems running tests for ruby 1.9. Should we release a new minor and mark this version as bad?

pitr-ch:
fix #768


nijikon:
@pitr-ch as I appreciate you fixing this, but I don't think `1.9.3` should still be supported. Support for it ended at 23rd of February 2015, https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/. People should upgrade or just last available version of a gem for their ruby version. We should not support old stuff for forever.

nijikon:
To be honest, we should support version `>= 2.3`, people should not be encouraged to run old ruby versions.

pitr-ch:
With 2 indirections, 20% of all existing gems depend on concurrent-ruby. I have to be extra careful regarding compatibility. I plan to drop the support for old Ruby version eventually but I want to communicate it first with authors of the libraries depending on concurrent-ruby. Not dropping it all of a sudden as I did in 1.1.

nijikon:
Agreed, I would opt for bumping to 2.0 and drop support for `< 2.3`.

nijikon:
Or maybe `2.0` is too big of a bump, I guess `1.1` would be fine for me if we would communicate it in the changelog as breaking changes.

PR #779, opened 2018-11-15
[Travis] Add latest JRuby version with JDK-11

prashantvithani:


pitr-ch:
Thanks! I've updated it. We test only latest of each minor version. Also the older version are usually placed further down in the file.

pitr-ch:
I've also added kept tests for both Java 8 and 11.

.travis.yml
```diff
@@ -9,8 +9,11 @@ matrix:
   include:
     - name: MRI 2.5.1 Latest
       rvm: 2.5.1
-    - name: JRuby 9.2.0.0 Latest
-      rvm: jruby-9.2.0.0
+    - name: JRuby 9.2.4.0 Latest on Java 11
```
olleolleolle:
```suggestion
    - name: JRuby 9.2.4.1 Latest on Java 11
```

.travis.yml
```diff
@@ -9,8 +9,11 @@ matrix:
...
+      rvm: jruby-9.2.4.1
+      jdk: oraclejdk11
+    - name: JRuby 9.2.4.0 Latest on Java 8
```
olleolleolle:
```suggestion
    - name: JRuby 9.2.4.1 Latest on Java 8
```

.travis.yml
pitr-ch: Thanks.

pitr-ch:
Merging the commit into #780. The fix in #780 needs java 11 to be tested and this does not pass without it.

PR #787, opened 2018-12-19
let's not (potentially) leak context on Java 7

kares:
holding a context instance will avoid RubyThread from being gc-d


ext/concurrent-ruby/com/concurrent_ruby/ext/SynchronizationLibrary.java
```diff
@@ -23,6 +23,7 @@
 public class SynchronizationLibrary implements Library {
 
     private static final Unsafe UNSAFE = loadUnsafe();
+    private static final boolean FULL_FENCE = supportsFences();
```
pitr-ch: good catch

pitr-ch:
rebased on master to pass CI

PR #788, opened 2018-12-26
Update rake-compiler-dock to 0.7.0

larskanis:
It adds support for ruby-2.6 and JRuby. See also https://github.com/rake-compiler/rake-compiler-dock/issues/21


pitr-ch:
Thanks @larskanis, testing locally.

PR #803, opened 2019-02-19
link to soure-code from gemspec

grosser:
- rubygems removed the edit UI, so it should be in the gemspec now
 - makes it easy for automated tools to link to github for comparisons


pitr-ch:
Thanks for the PR @grosser, what 'edit UI' are you referring to?

grosser:
rubygems used to have a edit UI where source/docs links could be entered ... but that's gone now

PR #804, opened 2019-03-07
Add respond_to_missing? for async/await

cavinkwon:
Async/Await object must respond correctly to the `respond_to?` method. As you can see below, the echo method exists but returns false.
```ruby
async_class = Class.new do
  include Concurrent::Async
  attr_accessor :accessor

  def initialize(*args)
  end

  def echo(msg)
    msg
  end
end

async_obj = async_class.new
async_obj.async.respond_to?(:echo) #=> false
async_obj.await.respond_to?(:echo) #=> false
```
So,  I added `respond_to_missing?` methods. The above code will work as follows
```ruby
async_obj.async.respond_to?(:echo) #=> true
async_obj.await.respond_to?(:echo) #=> true
async_obj.async.respond_to?(:not_exist_method) #=> false
async_obj.await.respond_to?(:not_exist_method) #=> false
```

### DIFF
--- a/lib/concurrent/async.rb
+++ b/lib/concurrent/async.rb
@@ -333,6 +333,13 @@ module Concurrent
         ivar
       end
 
+      # Check whether the method is responsive
+      #
+      # @param [Symbol] method the method being called
+      def respond_to_missing?(method, include_private = false)
+        @delegate.respond_to?(method) || super
+      end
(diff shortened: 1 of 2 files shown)


pitr-ch:
Nice find thank you for the fix!

PR #812, opened 2019-06-21
Move edge bits from lib-edge to lib/concurrent-edge

adamruzicka:
Move edge bits from `lib-edge` to `lib/concurrent-edge`

Debian packaging tooling expects the source of the gem to live inside `lib` and is profoundly confused in the code is in a different place, such as `lib-edge`.

This commit moves the contents of `lib-edge` into `lib/concurrent-edge` and places proxy files into `lib-edge`. The proxy files forward requires to the correct files in `lib/concurrent-edge`.

So far the files in `lib/concurrent-edge` use the original requires. I'd be glad if we could change that eventually to use the new paths, but I'll leave it as is for now. At least the tests will tell us the backward compatibility works.

This attemps to resolve debian packaging issues we were hitting in https://github.com/theforeman/foreman-packaging/pull/3438


adamruzicka:
As far as I can tell the test failure should be unrelated

adamruzicka:
Now thinking about it, we will need to change the files in `lib/concurrent-edge` to use the "new" require paths to resolve the issue we originally had

infomatico:
hey @adamruzicka , the problem with the topic [#3438](https://github.com/theforeman/foreman-packaging/pull/3438) is already fixed?

adamruzicka:
This is an attempt to make the edge parts be more easily package-able in deb. By itself it won't do much, but it is the first step towards being able to package it.

mmoll:
@pitr-ch any movement here? :)

pitr-ch:
@mmoll some :) The fix is ready, I need to time it with few other PRs to minimize conflicts.

ehelms:
@adamruzicka looks like this will need a rebase

@pitr-ch possible for you to give a timeline for pulling this and the other fixes you mentioned in for a release? We're facing quite a long time of having broken Foreman Proxy's with dynflow capabilities on Debian leading to quite diverged deployments on RPM vs. Debian based systems. And the whole Foreman community would greatly appreciate any expedited help here.

pitr-ch:
@ehelms I am taking care of this PR, I'll rebase it. Waiting for few more PRs to land first though. I've put aside some free time in December to work on concurrent-ruby. I expect a release then, I don't think I will have free time sooner. Sorry to hear that, I thought there is a reasonable workaround in place, I did not realise the implications of the problem.

mmoll:
What's the status here? :)

pitr-ch:
Hm, It take too long I will not wait for the PRs I wanted. I will get this going and released.

ehelms:
@pitr-ch thanks for the merge! any timeline on a release to help us plan?

pitr-ch:
@ehelms I will do a preview release this weekend for you to make sure that everything is fine. Release could be out the week after that.

pitr-ch:
pre-release is out, @ehelms please test and let me know if the new layout is ok.

mmoll:
From what I can say from a quick look, this is good :+1:

pitr-ch:
1.1.6 was released with the changed layout

momolog:
Is it possible that this change may lead to problems like: `LoadError: cannot load such file -- concurrent/map`? 
We have reason to believe this is the case because we are getting this error after upgrading from `1.1.5` to `1.1.6`.

pitr-ch:
@momolog I cannot think of any reason why it would break
```
[2] pry(main)> Concurrent::Map
NameError: uninitialized constant Concurrent::Map
Did you mean?  Math
from (pry):2:in `__pry__'
[3] pry(main)> require "concurrent/map"
=> true
[4] pry(main)> Concurrent::Map
=> Concurrent::Map
[5] pry(main)>
```
It is fine for me. Could you file an issue with more details, ideally a reproducer.

Issue #44, opened 2014-04-01
Pre-release with new thread pools?


jdantonio:
On April 9th I will be presenting (remotely) to the [Copenhagen Ruby Brigade](http://lanyrd.com/2014/cphrb/). I would like to build a pre-release version of the gem before then so that people can start working with the new thread pools. This pre-release will include everything in the [changelog](https://github.com/jdantonio/concurrent-ruby/wiki/Changelog) up to that point. I would like to build the pre-release on April 8.

The stable version will still be 0.5.0. There is currently no planned release date for stable 0.6.0.

@mighe @chrisseaton @lucasallan Any objections?

PR #814, opened 2019-07-06
Use Ruby's Etc.nprocessors for processor_count if available

jhawthorn:
In 0371ed4ad5c910fdcf569c8e85d7de96abf5903e from the discussions in #576, we switched the default method of detecting processor count from `/usr/bin/nproc` to reading `/proc/cpuinfo`. This avoids a fork, but means that we no longer respect CPU affinity, which we should want if using this value to control parallelism. I imagine it should work on Alpha too.

MRI has had `Etc.nprocessors` since Ruby 2.2, which uses:
- sched_getaffinity(): Linux
- sysconf(_SC_NPROCESSORS_ONLN): GNU/Linux, NetBSD, FreeBSD, OpenBSD, DragonFly BSD, OpenIndiana, Mac OS X, AIX
- GetSystemInfo(): Win32

This should be faster than the alternatives of reading from `/proc/cpuinfo`, using WIN32OLE, or spawning a new process (though all should be plenty fast since this is memoized). It also will take into account CPU affinity.

**Before:**
```
$ taskset 0x3 ruby -Ilib -rconcurrent-ruby -e 'puts Concurrent.processor_count'
12
$ be ruby -Ilib benchmark_processor_count.rb
compute_processor_count
                          5.581k (± 0.9%) i/s -     27.999k in   5.017628s
```

**After:**
```
$ taskset 0x3 bundle exec ruby -Ilib -rconcurrent-ruby -e 'puts Concurrent.processor_count'
2
$ be ruby -Ilib benchmark_processor_count.rb
compute_processor_count
                          1.685M (± 1.0%) i/s -      8.442M in   5.010886s
```

cc @pitr-ch @matthewd @graaff

### DIFF
--- a/lib/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent/utility/processor_counter.rb
@@ -1,3 +1,4 @@
+require 'etc'
 require 'rbconfig'
 require 'concurrent/delay'
 
@@ -78,7 +79,7 @@ module Concurrent
       def compute_processor_count
         if Concurrent.on_jruby?
           java.lang.Runtime.getRuntime.availableProcessors
-        elsif defined?(Etc) && Etc.respond_to?(:nprocessors) && (nprocessor = Etc.nprocessors rescue nil)
+        elsif Etc.respond_to?(:nprocessors) && (nprocessor = Etc.nprocessors rescue nil)
           nprocessor
         else
           os_name = RbConfig::CONFIG["target_os"]
--- a/lib/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent/utility/processor_counter.rb
@@ -22,6 +22,8 @@ module Concurrent
       # occasionally poll this property." Subsequently the result will NOT be
       # memoized under JRuby.
       #
+      # Ruby's Etc.nprocessors will be used if available (MRI 2.2+).
+      #
       # On Windows the Win32 API will be queried for the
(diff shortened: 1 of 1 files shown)


graaff:
$ ruby -r rubygems -r etc -e 'puts Etc.nprocessors'
4

$ ruby -v
ruby 2.4.6p354 (2019-04-01 revision 67394) [alpha-linux]

$ uname -a
Linux monolith 4.19.37+ #44 SMP Wed Jun 19 13:07:21 CEST 2019 alpha EV68AL Tsunami GNU/Linux

lib/concurrent/utility/processor_counter.rb
```diff
@@ -76,6 +78,8 @@ def physical_processor_count
       def compute_processor_count
         if Concurrent.on_jruby?
           java.lang.Runtime.getRuntime.availableProcessors
+        elsif defined?(Etc) && Etc.respond_to?(:nprocessors) && (nprocessor = Etc.nprocessors rescue nil)
```
graaff: Etc is not included by default, so should this probe also include a `require "etc"`?

jhawthorn: 👍 Good catch. I added it as a top-level require (I believe all supported rubies should include `etc`, older ones just don't implement `nprocessors`).

PR #823, opened 2019-09-13
FEATURE: add interface compatible constant version of thread local var

SamSaffron:
ThreadLocalVar can be expensive as it spins a thread in finalizers. In some
cases consumers may want to swap in a constant value implementations that
is very cheap.

`Concurrent::ConstantThreadLocalVar` is interface compatible with
AbstractThreadLocalVar, however does not allow any changing of .value

See more context in: https://github.com/rails/rails/pull/36949

@pitr-ch how do you feel about this? Would you prefer this here or in Rails?


jdantonio:
I'm not actively involved in this gem anymore, but I think this PR would be a good addition, FWIW.

pitr-ch:
@SamSaffron Sorry for overlooking the PR. I'll update it and have a look at merging it.

lib/concurrent-ruby/concurrent/atomic/constant_thread_local_var.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    protected
+
+    def allocate_storage
+      # nothing to do
+    end
+
+    def default
+      if @default_block
+        @default_block.call
```
pitr-ch: This evaluates the `@default_block` on each value read, so it may return different values each time value is called. I think that if the `default_block` is provided the ConstantThreadLocalVar should still store the computed value per thread (since it might differ). So the implementation could inherit from ThreadLocalVar, disallow using `value=` , and it would be good if it optimises `@default` case (the value is equal for all threads).

chrisseaton:
@SamSaffron is it still worth me getting this merged?

SamSaffron:
I am mixed Chris ... I ended up working around the issue here: https://github.com/rails/rails/pull/37200 

The finalizer approach though is a bit problematic in many ways, but the workaround in pure Ruby is simple enough. 

The problem with adding the API here though is that we have no real world consumers for it anymore.

chrisseaton:
Closing as no immediate need and we're trying to get on top of the backlog of this gem.

Feel free to re-suggest in the future.

PR #829, opened 2019-10-06
Reimplement TimerTask using promises

adamruzicka:


lib/concurrent/timer_task2.rb
```diff
@@ -0,0 +1,297 @@
...
+    #   @yieldparam cancellation a cancellation object representing the joined cancellation
+    #     of the timer task and this run's timeout
+    #
+    #   @return [TimerTask] the new `TimerTask`
+    def initialize(opts = {}, &task)
+      raise ArgumentError.new('no block given') unless block_given?
+      raise ArgumentError.new('reschedule must be either :before or :after') unless [nil, :before, :after].include?(opts[:reschedule])
+      @execution_interval = AtomicReference.new nil
+      @timeout_interval = AtomicReference.new nil
+      @cancellation = AtomicReference.new nil
```
pitr-ch: Please use Capitalized names for the instance variables, since they are final.

lib/concurrent/timer_task2.rb
```diff
@@ -0,0 +1,297 @@
...
+  #     observe.call(channel)
+  #   end
+  #   #=> (false) Execution failed with error StandardError
+  #   #=> (false) Execution failed with error StandardError
+  #   #=> (false) Execution failed with error StandardError
+  #   task.shutdown
+  #   channel.pop
+  #
+  # @see http://docs.oracle.com/javase/7/docs/api/java/util/TimerTask.html
+  class TimerTask2
```
pitr-ch: It should inherit from `Concurrent::Object` so it can use `safe_initialization!`

lib/concurrent/timer_task2.rb
```diff
@@ -0,0 +1,297 @@
...
+    #   task = Concurrent::TimerTask2.new(execution_interval: 10) { print "Hello World\n" }
+    #   task.running? #=> false
+    #   task.execute
+    #   task.running? #=> true
+    #
+    # @example Instance and execute in one line
+    #   task = Concurrent::TimerTask2.new(execution_interval: 10) { print "Hello World\n" }.execute
+    #   task.running? #=> true
+    def execute
+      return if running?
```
pitr-ch: should return self as well

pitr-ch: However, it would be food to have `execute?` which returns true of false as well.

adamruzicka:
The final changes are pretty straightforward, I just need to find time to actually implement them, hopefully sometime during this week.

lib/concurrent/timer_task2.rb
adamruzicka: Do you mean `::Concurrent::Synchronization::Object`?

lib/concurrent/timer_task2.rb
adamruzicka: Should they use capitalized snake case or camel/pascal case?

pitr-ch: `@TimeoutInterval` same as class, thanks

lib/concurrent/timer_task2.rb
pitr-ch: yeah, my bad

lib/concurrent/timer_task2.rb
```diff
@@ -0,0 +1,324 @@
...
+      raise IllegalOperationError.new('TimerTask2 cannot be resumed after it was stopped') if @Executor.nil?
+      @Cancellation.set(Cancellation.new)
+      execute_task cancellation, @RunNow
+      true
+    end
+
+    def shutdown
+      # TODO: Implement soft shutdown
+      kill
+    end
```
adamruzicka: How exactly should soft shutdown behave?

lib/concurrent/timer_task2.rb
```diff
@@ -0,0 +1,325 @@
...
+      @ExecutionInterval = AtomicReference.new nil
+      @TimeoutInterval = AtomicReference.new nil
+      @Cancellation = AtomicReference.new nil
+
+      self.execution_interval = opts[:execution] || opts[:execution_interval] || EXECUTION_INTERVAL
+      self.timeout_interval = opts[:timeout] || opts[:timeout_interval] || TIMEOUT_INTERVAL
+
+      @Channel = opts[:channel]
+
+      @Stopped = false
```
pitr-ch: This one is not immutable. It got renamed by mistake.

lib/concurrent/timer_task2.rb
```diff
@@ -0,0 +1,325 @@
...
+      raise ArgumentError.new('no block given') unless block_given?
+      raise ArgumentError.new('reschedule must be either :before or :after') unless [nil, :before, :after].include?(opts[:reschedule])
+      @ExecutionInterval = AtomicReference.new nil
+      @TimeoutInterval = AtomicReference.new nil
+      @Cancellation = AtomicReference.new nil
+
+      self.execution_interval = opts[:execution] || opts[:execution_interval] || EXECUTION_INTERVAL
+      self.timeout_interval = opts[:timeout] || opts[:timeout_interval] || TIMEOUT_INTERVAL
+
+      @Channel = opts[:channel]
```
pitr-ch: I think we discussed some problems which could arise if we allow the channel to be shared between two timer-tasks. Like the arriving messages would not have clear sender. Is that right? It would be better if it would accept arguments for creating the channel, where new channel would be created on dup, or to add dup support on the channel and dup it as well in timer-task's dup.

lib/concurrent/timer_task2.rb
pitr-ch: Hm I think there are actually 3 levels to consider.
- stop: finish task if running, do not schedule a new tasks
- shutdown: also cancel running task, do not schedule a new tasks
- kill: kill running tasks by killing it's thread, do not schedule a new tasks
What did the old implementation support? 
For the new one I think all 3 make sense, could you implement them? Please free to experiment with the names or API for this, I am not sure myself what will be the best option.

lib/concurrent/timer_task2.rb
adamruzicka: Right, how about we drop [the reuse of channel in #dup](https://github.com/ruby-concurrency/concurrent-ruby/pull/829/files#diff-5d2534adbdf5de2a2ec965af44ecc206R200). This way `#dup`ed timers won't have a channel, unless a new channel is provided when `#dup` is called?

lib/concurrent/timer_task2.rb
adamruzicka: Not sure if we can do `kill` in a sensible way, will take a look what we can do

lib/concurrent/timer_task2.rb
pitr-ch: That would work but I think it would more surprising. Assuming that a duped timer-task is most often used in the same place/functional-context as the template than it would be expected the new timer-task to have a channel as the parent has.

lib/concurrent/timer_task2.rb
pitr-ch: The kill will always be dangerous. However, has to be there since the other methods can fail :/ It should be only used as a last resort when the previous methods failed to finish. If you run out of time just put raise NotImplemented there, or we could also meet for coffee and figure something together.

mhenrixon:
Any news on this? Also, will this PR prevent the problem that TimerTask spins up a limitless number of threads in certain scenarios?

chrisseaton:
Are people still interested in this PR? If so we can work on it.

mhenrixon:
@chrisseaton well at this point it would really complicate things for me since I now maintain a version of timer task that fixes the thread leak issue: https://github.com/mhenrixon/sidekiq-unique-jobs/blob/main/lib/sidekiq_unique_jobs/timer_task.rb

I still feel like the timer task shouldn't be leaking threads (and therefor memory) so yeah, please fix it! 😂

PR #832, opened 2019-10-31
Allow semaphore permit acquisition with blocks

sambostock:
Ruby often idiomatically uses blocks to open a resource and close it again automatically.

For example:

```ruby
file = File.open('file', 'r')
# do work with file
file.close

# vs

File.open('file', 'r') do |file|
  # do work with file
end
# file closed automatically
```

**This teaches semaphores a similar technique:**

```ruby
semaphore = Concurrent::Semaphore.new(1)

semaphore.acquire
# do work
semaphore.release

# vs

semaphore.acquire do
  # do work
end
# permit automatically released
```

### DIFF
--- a/ext/concurrent-ruby/com/concurrent_ruby/ext/JavaSemaphoreLibrary.java
+++ b/ext/concurrent-ruby/com/concurrent_ruby/ext/JavaSemaphoreLibrary.java
@@ -45,9 +45,13 @@ public class JavaSemaphoreLibrary {
         }
 
         @JRubyMethod
-        public IRubyObject acquire(ThreadContext context, IRubyObject value) throws InterruptedException {
-            this.semaphore.acquire(rubyFixnumToPositiveInt(value, "permits"));
-            return context.nil;
+        public IRubyObject acquire(ThreadContext context) throws InterruptedException {
+            return this.acquire(context, 1);
+        }
+
+        @JRubyMethod
+        public IRubyObject acquire(ThreadContext context, IRubyObject permits) throws InterruptedException {
+            return this.acquire(context, rubyFixnumToPositiveInt(permits, "permits"));
         }
 
         @JRubyMethod(name = "available_permits")
@@ -60,30 +64,32 @@ public class JavaSemaphoreLibrary {
             return getRuntime().newFixnum(this.semaphore.drainPermits());
         }
 
-        @JRubyMethod
-        public IRubyObject acquire(ThreadContext context) throws InterruptedException {
-            this.semaphore.acquire(1);
-            return context.nil;
-        }
-
         @JRubyMethod(name = "try_acquire")
         public IRubyObject tryAcquire(ThreadContext context) throws InterruptedException {
-           return getRuntime().newBoolean(semaphore.tryAcquire(1));
+            int permitsInt = 1;
+            boolean acquired = semaphore.tryAcquire(permitsInt);
+
+            return triedAcquire(context, acquired);
         }
 
         @JRubyMethod(name = "try_acquire")
         public IRubyObject tryAcquire(ThreadContext context, IRubyObject permits) throws InterruptedException {
-           return getRuntime().newBoolean(semaphore.tryAcquire(rubyFixnumToPositiveInt(permits, "permits")));
+            int permitsInt = rubyFixnumToPositiveInt(permits, "permits");
+            boolean acquired = semaphore.tryAcquire(permitsInt);
+
+            return triedAcquire(context, acquired);
         }
 
         @JRubyMethod(name = "try_acquire")
         public IRubyObject tryAcquire(ThreadContext context, IRubyObject permits, IRubyObject timeout) throws InterruptedException {
-             return getRuntime().newBoolean(
-                        semaphore.tryAcquire(
-                                rubyFixnumToPositiveInt(permits, "permits"),
-                                rubyNumericToLong(timeout, "timeout"),
-                                java.util.concurrent.TimeUnit.SECONDS)
-                );
+            int permitsInt = rubyFixnumToPositiveInt(permits, "permits");
+            boolean acquired = semaphore.tryAcquire(
+                    permitsInt,
+                    rubyNumericToLong(timeout, "timeout"),
+                    java.util.concurrent.TimeUnit.SECONDS,
+                    );
+
+            return triedAcquire(context, acquired);
         }
 
         @JRubyMethod
@@ -93,8 +99,8 @@ public class JavaSemaphoreLibrary {
         }
 
         @JRubyMethod
-        public IRubyObject release(ThreadContext context, IRubyObject value) {
-            this.semaphore.release(rubyFixnumToPositiveInt(value, "permits"));
+        public IRubyObject release(ThreadContext context, IRubyObject permits) {
+            this.semaphore.release(rubyFixnumToPositiveInt(permits, "permits"));
             return getRuntime().newBoolean(true);
         }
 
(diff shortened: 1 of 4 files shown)


pitr-ch:
@adamruzicka you have originally contributed semaphore implementation, would you have time to review this PR?

adamruzicka:
Hopefully tomorrow

pitr-ch:
Yeah the chain is not that clear. @sambostock could you please inline the acquire_and_release and the other method `try_...`.

chrisseaton:
@sambostock are you interested in making the small change that Petr suggested? Then we can merge this.

lib/concurrent-ruby/concurrent/atomic/mutex_semaphore.rb
```diff
@@ -23,8 +23,17 @@ def acquire(permits = 1)
 
       synchronize do
         try_acquire_timed(permits, nil)
-        nil
```
sambostock, line 26: I haven't worked with `synchronize` before, so I'm not sure if there was a reason this was in the same block. Didn't seem like it.

lib/concurrent-ruby/concurrent/atomic/mutex_semaphore.rb
```diff
@@ -23,8 +23,17 @@ def acquire(permits = 1)
...
+
+      if block_given?
+        begin
+          yield
+        ensure
+          release(permits)
+        end
+      end
+
+      nil
```
sambostock, line 36: I figured it was safe to not put this in the `synchronize` block, since we don't need the contents of the block to run in a synchronized context, and `release` is itself synchronized.

lib/concurrent-ruby/concurrent/atomic/mutex_semaphore.rb
```diff
@@ -48,13 +57,23 @@ def try_acquire(permits = 1, timeout = nil)
...
         end
       end
+
+      if acquired && block_given?
+        begin
+          yield
+        ensure
+          release(permits)
+        end
+      end
```
sambostock, line 74: As I was inlining this I wondered about the API here.

Given this is `TRY_acquire`, which is correct:

- `yield` if acquired
- always `yield(acquired)` and let the block decide what to do

My gut is the **former**, as it is more intuitive, and less likely someone would forget to check if the lock was acquired or not, but I thought I should check, as there could be cases where it might be useful to do something in the block if the lock failed to be acquired. Though maybe that use case should just use the non-block version... 🤷 

If we did want the latter, we could do something like

```suggestion
      if block_given?
        begin
          yield(acquired)
        ensure
          release(permits) if acquired
        end
      end
```

sambostock (commented):
Wow, it's been a while since I looked at this. Thanks for the ping @chrisseaton!

I think I've addressed the inlining. Let me know how it looks, and in particular if everything looks correct with the use of `synchronize` and my API question for `try_acquire`.

I'll circle back and add docs soon.

sambostock:
Actually, the docs were trivial to write. I don't know why I thought it would take longer.

I've squashed the inlining and docs into a single commit, so it's ready to go if you're happy with it!

lib/concurrent-ruby/concurrent/atomic/mutex_semaphore.rb
chrisseaton: It's there to stop an internal value being accidentally returned by the API - nothing to do with `synchronized`.

ext/concurrent-ruby/com/concurrent_ruby/ext/JavaSemaphoreLibrary.java
```diff
@@ -54,7 +54,7 @@ public IRubyObject acquire(ThreadContext context, IRubyObject value) throws Inte
         @JRubyMethod
         public IRubyObject acquire(ThreadContext context, IRubyObject value, final Block block) throws InterruptedException {
             this.acquire(context, value);
-            block.yield(context, null);
+            block.yieldSpecific(context);
```
sambostock, line 57: I've replaced `yield(context, null)` with `yieldSpecific(context)` in the hopes that will fix the [error I was seeing in CI](https://github.com/ruby-concurrency/concurrent-ruby/runs/4826321806?check_suite_focus=true), however the [docs I was able to find on `Block`](https://www.javadoc.io/static/org.jruby/jruby-core/9.2.0.0/org/jruby/runtime/Block.html#yieldSpecific-org.jruby.runtime.ThreadContext-) provide no info as to the difference between them...

Looks like I need a maintainer to kick off another CI run though.

chrisseaton: > Looks like I need a maintainer...

Done - sorry not sure why that's set - it's just free GH Actions.

sambostock: Looks like it only limits first time contributors. Probably so I don't mine bitcoin or whatever...

sambostock: Looks like it's still failing, but I'm afraid I'm not familiar enough with JRuby to figure it out...

```

LoadError:
  Java extensions are required for JRuby.
  load error: concurrent/concurrent_ruby -- java.lang.RuntimeException: (LoadError) Mismatched block parameters for method com.concurrent_ruby.ext.JavaSemaphoreLibrary$JavaSemaphore.tryAcquire
# ./lib/concurrent-ruby/concurrent/utility/native_extension_loader.rb:35:in `load_native_extensions'
# ./lib/concurrent-ruby/concurrent/synchronization.rb:5:in `<main>'
# ./lib/concurrent-ruby/concurrent/atomic/event.rb:2:in `<main>'
# ./lib/concurrent-ruby/concurrent/concern/obligation.rb:4:in `<main>'
# ./lib/concurrent-ruby/concurrent/delay.rb:2:in `<main>'
# ./lib/concurrent-ruby/concurrent/configuration.rb:2:in `<main>'
# ./lib/concurrent-ruby/concurrent.rb:4:in `<main>'
# ./spec/spec_helper.rb:36:in `<main>'
# ------------------
# --- Caused by: ---
# Java::OrgJrubyExceptions::LoadError:
#   (LoadError) Mismatched block parameters for method com.concurrent_ruby.ext.JavaSemaphoreLibrary$JavaSemaphore.tryAcquire
#   RUBY.load_native_extensions(/home/runner/work/concurrent-ruby/concurrent-ruby/lib/concurrent-ruby/concurrent/utility/native_extension_loader.rb:35)
```

Unfortunately, even the detailed backtrace just points to the line loading the class.

sambostock: Okay, actually, I think I may have figured it out by looking at the [JRuby implementation of `IO.open`](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyIO.java#L1164-L1181), which does something similar to what we want to do here.

Looks like instead of overloading the methods (one with `final Block block`, and one without), I need to just have the one with the block argument and check `block.isGiven()`.

I'll push an update soon, but looking into that I realized it would probably be more idiomatic to return the result of the `yield`, so I'm going to change that too.

sambostock:
Okay, I'm pushed an update:

- I extracted some small refactors into a separate commit, ahead of the main one
- I'm still using `yieldSpecific`. Not sure if that's what it should be using or not. I'm trying to yield to the block, without any arguments.
- I added a `final Block block` argument to the existing `acquire` and `tryAcquire` JRuby method, instead of overloading them. Combining this with `block.isGiven()` appears to be the way to handle optional blocks in JRuby Java files, based on scouring the JRuby source.
- I've changed the API slightly from what I originally proposed, to match other similar APIs (such as `IO.open`):
  - `acquire` **with a block** will return the return value of the block, instead of `nil`
  - `try_acquire` **with a block**
    - if the permits are acquired will return the return value of the block, instead of the acquired boolean
    - if the permits are not acquired will return `nil`, instead of the acquired boolean

I've updated the docs and specs to line up. Hopefully I haven't made any mistakes, and hopefully I'm finally right about my JRuby block usage, but I can't run the tests locally, so I'll have to wait until someone kicks off CI for me to find out.

sambostock:
Alright, looks like I've finally got JRuby working 🎉 

There are failures on JRuby HEAD, but they all seem unrelated to this PR.

PR #838, opened 2019-12-19
Implement #dup and #clone for Structs

### LINKED ISSUE #836: Dup and Clone with (Mutable)Structs
```
* Ruby implementation:             Ruby
* `concurrent-ruby` version:       1.1.5
```

Hey folks,

I've started using this awesome gem in an effort to make one of my employer's apps thread-safe. Thank you for all your hard work, this makes my life so much easier! :)

While I'd preferably not even use the `MutableStruct` I have to do so for now. And in doing so I noticed that `dup` and `clone` do not behave as they do with Ruby's normal `Struct`. I.e., since values are stored in an array and dup/clone are not overwritten  this happens:

```ruby
greeting = Concurrent::MutableStruct.new(:text).new("Hello, World!")
# => #<concurrent/mutable_struct {:text=>"Hello, World!"}>
greeting.dup.text = "Goodbye, World!"
# => "Goodbye, World!"
greeting.text
# => "Goodbye, World!"
```

While in a normal Ruby `Struct` a call to dup also duplicates the value table:

```ruby
greeting = Struct.new(:text).new("Hello, World!")
# => #<struct text="Hello, World!">
greeting.dup.text = "Goodbye, World!"
# => "Goodbye, World!"
greeting.text
# => "Hello, World!"
```

I don't know if this is something you'd deem a bug, I guess it depends on how much of a drop-in replacement mutable struct is supposed to be. However, it is something that surprised me, luckily in my test suite only.

Best regards,
Kai

kaikuchn:
This fixes #836 

### Why are the changes necessary?

When calling `#dup` or `#clone` on a plain old Ruby Struct the member values
are copied along with the Struct's instance. This is not true for MutableStruct,
SettableStruct or ImmutableStruct.

This is especially troublesome for MutableStruct since mutating a member value
of a copy would also mutate the original. It is also a problem when any of
Concurrent Ruby's Struct classes has a mutable member value, e.g., an array.
Changes to that array would be reflected in the original and the copy since they
essentially share the same array.

### What does this pull request cover?

* Implement `initialize_copy` for Concurrent Ruby's Struct classes to copy the
  member values when `dup` or `clone` are called.
* Add tests to ensure Struct classes behave just like plain old Ruby Structs:
  * dup does not preserve taints (singleton class, frozen state)
  * clone does preserve taints
  * clone and dup both preserve taints of member values
  * copies are shallow
* Add tests to ensure dup and clone of `MutableStruct` and `SettableStruct`
  are protected (synchronization)

### DIFF
--- a/lib/concurrent/immutable_struct.rb
+++ b/lib/concurrent/immutable_struct.rb
@@ -70,11 +70,11 @@ module Concurrent
       ns_select(&block)
     end
 
-    # @!macro struct_initialize_copy
-    #
+    private
+
     # @!visibility private
     def initialize_copy(original)
-      super
+      super(original)
       ns_initialize_copy
     end
 
--- a/lib/concurrent/immutable_struct.rb
+++ b/lib/concurrent/immutable_struct.rb
@@ -70,6 +70,14 @@ module Concurrent
       ns_select(&block)
     end
 
+    # @!macro struct_initialize_copy
+    #
+    # @!visibility private
+    def initialize_copy(original)
+      super
+      ns_initialize_copy
+    end
+
     # @!macro struct_new
     def self.new(*args, &block)
       clazz_name = nil
--- a/lib/concurrent/mutable_struct.rb
+++ b/lib/concurrent/mutable_struct.rb
@@ -169,16 +169,6 @@ module Concurrent
       synchronize { ns_select(&block) }
     end
 
-    # @!macro struct_initialize_copy
-    #
-    # @!visibility private
-    def initialize_copy(original)
-      synchronize do
-        super
-        ns_initialize_copy
-      end
-    end
-
     # @!macro struct_set
     #
     #   Attribute Assignment
@@ -204,6 +194,16 @@ module Concurrent
       end
     rescue NoMethodError
       raise NameError.new("no member '#{member}' in struct")
+    end
+
+    private
+
+    # @!visibility private
+    def initialize_copy(original)
+      synchronize do
+        super(original)
+        ns_initialize_copy
+      end
     end
 
     # @!macro struct_new
--- a/lib/concurrent/mutable_struct.rb
+++ b/lib/concurrent/mutable_struct.rb
@@ -169,6 +169,16 @@ module Concurrent
       synchronize { ns_select(&block) }
     end
 
+    # @!macro struct_initialize_copy
+    #
+    # @!visibility private
+    def initialize_copy(original)
+      synchronize do
+        super
+        ns_initialize_copy
+      end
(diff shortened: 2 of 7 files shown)


kaikuchn:
I am unsure whether or not `ImmutableStruct` should synchronize on `initialize_copy` as well.

Assume someone does this:

```ruby
TaskList = Concurrent::ImmutableStruct.new(:todos)
foo = TaskList.new([])
# Thread 1
foo.todos << 'create bugs'
# Thread 2
foo.clone
```

In my book using an immutable struct and then mutating it's members is just wrong and people should not do that. But, WDYT?

pitr-ch:
Thanks for the very well prepared PR. I've pushed one commit with minor changes. When the tests are green I will merge it.

pitr-ch:
Good question. Even though it's good to be defensive I would not add it here for the following reasons that out-weight defensiveness for me. The race condition is in the array, the pattern is discouraged, and it would synchronise and therefore slow down immutable struct.

pitr-ch:
Please look at the failing tests. There is an issue with older Ruby versions.

kaikuchn:
Ok. The problem was that in Ruby versions prior to 2.4 some objects would throw a TypeError when you cannot clone/dup them. Nowadays they simply return themselves (Fixnum, NilClass, TrueClass, FalseClass, Method, ...).

I fixed that the easy way by rescuing `TypeError` when calling clone. Another way would be to use predicates to check whether or not an object can be cloned. Rails did this by patching object with a `duplicable?` method. Apparently that way is much faster (for older Versions of Ruby) and can be relevant on hot code paths.
I don't think that is a worthwhile trade, IMHO it's better to use the rescue idiom and just accept that for older Ruby versions this will be quite a bit slower, than to monkey around in object and make it a tiny bit slower for every Ruby version.

If you want to read more about dup/clone in context of immediate objects and Rails approach to the problem, this blog post gives a nice overview: https://blog.arkency.com/2017/03/prototypes-in-ruby/

kaikuchn:
I think the JRuby failures may be unrelated to my changes. Seems like there is bug in the RakeCompiler JavaExtensionTask? Must've been introduced recently though, by this commit https://github.com/rake-compiler/rake-compiler/commit/f808dd7a74893d6269c6cc72e049d7bb85b84b65#diff-74d7db1134bce6e2c234c28722496689 I'll raise the issue with them.

pitr-ch:
@kaikuchn Thanks for for the fix for old Rubies, the investigation and the follow up with rake-compiler! They already fixed it and there is a new release. CI is green, merging.

PR #840, opened 2020-01-08
Introduce DaemonThreadFactory

### LINKED ISSUE #817: Executor threads not being created as daemon
There appears to be an executor service somewhere in concurrent-ruby that's not marking the threads as "daemon" on JVM. As a result, this causes that executor to keep the process from terminating.

This issue has been affecting builds of JRuby for some time; the CRuby standard library test suite we run appears to trigger concurrent-ruby to create this executor, resulting in the entire suite hanging at the end when it should gracefully shut down.

https://github.com/jruby/jruby/issues/5718#issuecomment-492372765

I am not sure where the code is for launching such an executor...someone else might be able to find this quickly. I will try.

### LINKED ISSUE #839: TimerTask is starting a non-daemon thread
I have a small reproduction example:
```ruby
  def test_timer_task
    task = Concurrent::TimerTask.new(execution_interval: 1) do
      puts "Here"
    end
    task.execute
  end
```

The threads started by TimerTask **should be** daemons however the JVM does not shutdown.
There seems to be some other _concurrent-ruby_ thread which is not a **daemon**

The problem is that at the end of the test, the JVM remains active and does not shutdown.

(The following is from _jstack_)
```
"pool-6-thread-1" #61 prio=5 os_prio=31 tid=0x00007fe599823800 nid=0xb103 in Object.wait() [0x0000700013a56000]
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
        at java.lang.Object.wait(Native Method)
        at java.lang.Object.wait(Object.java:460)
        at org.jruby.RubyThread$SleepTask.run(RubyThread.java:1410)
        - locked <0x000000076e124a08> (a com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject)
        at org.jruby.RubyThread.executeBlockingTask(RubyThread.java:1463)
        at org.jruby.RubyThread.wait_timeout(RubyThread.java:1933)
        at com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject.nsWait(SynchronizationLibrary.java:254)
        at com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject$INVOKER$i$0$1$nsWait_DBG.call(SynchronizationLibrary$JRubyLockableObject$INVOKER$i$0$1$nsWait_DBG.gen)
        at org.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:796)
        at org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:202)
        at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:153)
        at org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:315)
        at org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:159)
        at org.jruby.ir.interpreter.Interpreter.INTERPRET_BLOCK(Interpreter.java:127)
        at org.jruby.runtime.InterpretedIRBlockBody.yieldDirect(InterpretedIRBlockBody.java:109)
        at org.jruby.runtime.IRBlockBody.yieldSpecific(IRBlockBody.java:80)
        at org.jruby.runtime.Block.yieldSpecific(Block.java:134)
        at org.jruby.RubyKernel.loop(RubyKernel.java:1316)
        at org.jruby.RubyKernel$INVOKER$s$0$0$loop_DBG.call(RubyKernel$INVOKER$s$0$0$loop_DBG.gen)
        at org.jruby.internal.runtime.methods.JavaMethod$JavaMethodZeroBlock.call(JavaMethod.java:536)
        at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:77)
        at org.jruby.runtime.callsite.CachingCallSite.callIter(CachingCallSite.java:84)
        at org.jruby.ir.instructions.CallBase.interpret(CallBase.java:429)
        at org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:360)
        at org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:159)
        at org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:84)
        at org.jruby.internal.runtime.methods.InterpretedIRMethod.INTERPRET_METHOD(InterpretedIRMethod.java:182)
        at org.jruby.internal.runtime.methods.InterpretedIRMethod.call(InterpretedIRMethod.java:165)
        at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:163)
        at org.jruby.runtime.callsite.CachingCallSite.callIter(CachingCallSite.java:170)
        at org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:336)
        at org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:159)
        at org.jruby.ir.interpreter.Interpreter.INTERPRET_BLOCK(Interpreter.java:127)
        at org.jruby.runtime.InterpretedIRBlockBody.yieldDirect(InterpretedIRBlockBody.java:109)
        at org.jruby.runtime.BlockBody.yield(BlockBody.java:114)
        at org.jruby.runtime.Block.yield(Block.java:165)
        at com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject.rubySynchronize(SynchronizationLibrary.java:232)
        - locked <0x000000076e124a08> (a com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject)
        at com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject$INVOKER$i$0$0$rubySynchronize_DBG.call(SynchronizationLibrary$JRubyLockableObject$INVOKER$i$0$0$rubySynchronize_DBG.gen)
```

```
* Operating system:               mac
* Ruby implementation:             jruby 9.1.17.0 (2.3.3)
* `concurrent-ruby` version:       1.1.5
```

fzakaria:
Introduce a ThreadFactory that will set `Thread#setDaemon` according
to the _auto_terminate_ value.

This makes the need for the `AtExit` functionality uneeded since all
threads can now be started daemonized.

Furthermore, the gross **hack** of `Configuration#disable_at_exit_handlers!`
is especially not needed since this allows the JVM to properly terminate
even with minitest.

You can test this with:
```
ruby test/concurrent/executor/test_cached_thread_pool.rb
```

This fixes issues reported by others wherein the JVM fails to shutdown when used with Minitest.

fixes #839
fixes #817


fzakaria:
tag @kaikuchn  @headius

fzakaria:
With this change I would rip out the _AtExit_ hack completely.
All threads **by default** in Ruby are daemonized and _concurrent-ruby_ should respect this by default.

Anyone who uses the library and sets `auto_terminate` to false, is then responsible for termination themselves.

test/concurrent/executor/test_cached_thread_pool.rb
```diff
@@ -0,0 +1,14 @@
+$LOAD_PATH.unshift File.expand_path("../../../lib", __dir__)
```
fzakaria: Unfortunately, testing the functionality is only possible with minitest.
You can see the comment surround [disable_at_exit_handlers!](https://ruby-concurrency.github.io/concurrent-ruby/1.1.4/Concurrent.html#disable_at_exit_handlers!-class_method) to read why it can't be tested in RSpec.

For now I've included a single raw minitest file -- I'm open to suggestion as to how best to incorporate it into the Rakefile according to this build process.

fzakaria:
The Travis CI failure seems to be related to use of `Unsafe` and not my code change.

headius:
I pushed a tweak to the travis configs to use JRuby 9.2.9.0 instead of 9.2.8.0, which seems to have fixed the two hanging "latest" builds.

fzakaria:
@headius can you restart the travis ?
I'd love to get this in + whatever changes necessary.

Seems pretty **high** considering it makes using Minitest & this library a PITA

headius:
@fzakaria I'm looking at master to try to get everything green there, but with at least JRuby 9.2.9 and 9.1.17 green I think we'd be ok to merge this in. Could you rebase your PR on master and repush, just so I can confirm 9.2.9 on Java 8 still passes?

There are still failures for JRuby 9.2.9 on Java 11 and for other JRuby versions (on master) that aren't your fault.

headius:
FWIW I do not have push rights for the gem, so I'll need help from @pitr-ch or @jdantonio to get a release out.

fzakaria:
Sounds good -- I'll rebase in ~30 minutes. Going to pick up my son + few other errands
etc..

As an aside; if this gets merged; what do you think of removing most of the `AtExit` code ?
Seems that it will be superseded by just making all the threads daemon

headius:
@fzakaria Possibly... I have to inspect that logic a bit more closely before I make a decision. The key point will be whether these daemon threads need any sort of at_exit cleanup. If they don't and they can just die, then it's probably fine.

There may still be a need for at_exit logic in other places, though.

fzakaria:
@headius just rebased

We can move the discussion else where regarding AtExit if there's an IRC / Slack also.

My thoughts on the matter are highly influenced from Java's [ExecutorService](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html) which does not have a "auto clean shutdown" -- rather individually the user is responsible for adding a [shutdown hook](https://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)) themselves.
 
concurrent-ruby seems heavily inspired by Java's API; so leaving that up to the user sounds reasonable and avoids nasty problems light this in the library 
(where another library such as minitest overrides the shutdown behavior)

headius:
> We can move the discussion else where regarding AtExit if there's an IRC / Slack also.

There's an active concurrent-ruby channel on Gitter: https://gitter.im/ruby-concurrency/concurrent-ruby. Let's discuss there.

headius:
Strangely, the JRuby 9.2.9 on Java 11 suite passes just fine on my local system (MacOS).

fzakaria:
The build failures aren't that easy to grok however they don't look to be related to the change AFAICT @headius .

jdantonio:
@fzakaria The failing test has a comment `FIXME this leads to weird message processing ordering` and is one of the "edge" features. So this may just be a buggy test and probably shouldn't prevent us from merging. Can you please comment out the test on line 191 of `spec/concurrent/actor_spec.rb` and see if it passes CI? If that works we may just temporarily disable the test and merge your change.

spec/concurrent/actor_spec.rb
```diff
@@ -188,7 +188,7 @@ def on_message(message)
         end
       end
 
-      it 'links' do
+      skip 'links' do
```
fzakaria: @jdantonio  added the skip here.

fzakaria:
Okay! I added some new code to conditionally skip the whole actor_spec.rb for JDK11

fzakaria:
Okay @headius & @jdantonio looks like JDK 11 build is just FUBAR

```
TypeError: illegal access on 'isShutdown': class org.jruby.javasupport.JavaMethod (in module org.jruby.dist) cannot access a member of class java.util.concurrent.Executors$DelegatedExecutorService (in module java.base) with modifiers "public"
        ns_shutdown? at /home/travis/build/ruby-concurrency/concurrent-ruby/lib/concurrent/executor/java_executor_service.rb:70
         ns_running? at /home/travis/build/ruby-concurrency/concurrent-ruby/lib/concurrent/executor/java_executor_service.rb:58
```

I think it should be moved to **allowed failures** section in the CI

headius:
I believe the Java 11 errors are due to a bug in JRuby 9.2.9 that's already fixed for 9.2.10.

headius:
Confirmed: JRuby 9.2.9 (on Java 11) presents the same errors and hangs the same way locally, while JRuby 9.2.10 runs to completion every time.

I believe it is safe to disregard the JRuby 9.2.9 on Java 11 failure.

headius:
I have added concurrent-ruby to JRuby's travis-ci run to make sure it stays green in the future: https://travis-ci.org/jruby/jruby/jobs/637495980

lib/concurrent/executor/cached_thread_pool.rb
```diff
@@ -51,11 +50,11 @@ def initialize(opts = {})
     def ns_initialize(opts)
       super(opts)
       if Concurrent.on_jruby?
+        self.auto_terminate = opts.fetch(:auto_terminate, true)
         @max_queue = 0
-        @executor = java.util.concurrent.Executors.newCachedThreadPool
+        @executor = java.util.concurrent.Executors.newCachedThreadPool Concurrent::DaemonThreadFactory.new(self.auto_terminate?)
```
headius: Maybe this should be two lines?

test/concurrent/executor/test_cached_thread_pool.rb
headius: It would seem better to spin off a separate process and ensure that it terminates within some timeout. The test is fine but I think it should run in its own subprocess.

headius:
I've tweaked CI on master to hide the JRuby 9.2.9 + Java 11 failure (fixed in JRuby 9.2.10/head) and move jruby-head out of allowed-failures. If you remove your spec skips and rebase we should see a green result.

test/concurrent/executor/test_cached_thread_pool.rb
fzakaria: There were tests that essentially did that in RSpec -- the failure however was that Minitest was messing with the AtExit handlers.
Given that it's a very popular test framework; I thought it would be reasonable to incorporate it sparingly for such cases. I'm not too strongly attached one way or another.

jdantonio:
@fzakaria @headius  Thanks for all your effort on this? Let's give @pitr-ch a few days to weigh in. If he isn't available early next week I'll approve the merge.

pitr-ch:
Thank you very much for the contribution! I'll have time to look at it on the weekend.

test/concurrent/executor/test_cached_thread_pool.rb
kaikuchn: Could we instead start a simple process that messes with the `AtExit` handlers? That might be easier to understand and we wouldn't introduce another development dependency. WDYT?

lib/concurrent/executor/cached_thread_pool.rb
```diff
@@ -1,6 +1,5 @@
 require 'concurrent/utility/engine'
 require 'concurrent/executor/thread_pool_executor'
-
```
kaikuchn: Style: add empty line between require and module/class definition.

lib/concurrent/executor/java_daemon_thread_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    def newThread(runnable)
+      thread = java.util.concurrent.Executors.defaultThreadFactory().newThread(runnable)
+      thread.setDaemon(@daemonize)
+      return thread
+    end
+
+  end
+
```
kaikuchn: style: don't add empty lines between `end`s.

test/concurrent/executor/test_cached_thread_pool.rb
```diff
@@ -0,0 +1,14 @@
...
+require 'minitest/autorun'
+require 'concurrent'
+class CachedThreadPoolTest < Minitest::Test
+
+  def test_cached_thread_pool_does_not_impede_shutdown
+    pool = Concurrent::CachedThreadPool.new
+    pool.post do
+      sleep
+    end
+
```
kaikuchn: style: don't add empty lines between `end`s.

fzakaria:
@kaikuchn thanks for the feedback -- I applied all the style changes.
(Future issue: it would be great to incorporate rubocop to just automatically apply style)

test/concurrent/executor/test_cached_thread_pool.rb
fzakaria: A sub-process could be written which calls `disable_at_exit_handlers!`
Personal opinion on the matter though is that a test that verifies that it works with minitest is pretty useful given the  popularity of the gem.

This PR has seem some churn -- I'd love to get this fix in and follow-up with ammendments/improvements to the testing facility also ?

That would also play well with the idea I've been chatting about on _gitter_ channel about even potentially removing **AtExit** completely. (It's uneeded now if all the threads are daemon)

pitr-ch:
Looks good. Thanks everyone involved for the contribution!
I have a request though. Could you create a subprocess to test this in spec instead of introducing minitest? Removing AtExit in a followup PR makes sense.

pitr-ch:
`master` changed file layout, I've rebased this PR for you and force pushed.

spec/concurrent/executor/cached_thread_pool_spec.rb
```diff
@@ -201,6 +201,30 @@ module Concurrent
         end
       end
 
+      context 'auto terminate' do
+
+        # https://github.com/ruby-concurrency/concurrent-ruby/issues/817
+        # https://github.com/ruby-concurrency/concurrent-ruby/issues/839
+        it 'does not stop shutdown ' do
+          Timeout::timeout(10) do
+            `
```
fzakaria: @pitr-ch I re-worked the test and have it running via spec.

I validated that this test on _master_ causes the test to hang.
Unfortunately, I can't seem to get it to gracefully fail even with Timeout. The timeout error raising _Kernel#raise_ does not seem to interrupt it waiting for the subprocess. Not sure how much it matters since we expect the test to pass...

note: You can't use Process#fork on JRuby

fzakaria:
The JRuby head failure does not seem related to the pull-request:
```rspec ./spec/concurrent/atomic/event_spec.rb:93 # Concurrent::Event#wait returns immediately when the event has been set```

(Could be from the rebase from master?)

headius:
Green!

fzakaria:
@pitr-ch appreciate a release when you can.

pitr-ch:
Thanks again for the fix. I am doing pre-release this weekend and a release probable next one.

pitr-ch:
pre-release is out

pitr-ch:
released in 1.1.6

headius:
@pitr-ch Thank you!

Issue #796, opened 2019-02-03
Concurrent::Set is not thread-safe in CRuby


fxn:
`Concurrent::Set` returns `Set` in CRuby with this argument:

```ruby
# Because MRI never runs code in parallel, the existing
# non-thread-safe structures should usually work fine.
::Set
```

As discussed in #793, that is not enough for thread-safety, you also need the interface to be atomic somehow. For `Hash` and `Array` this holds because they are written in C, but `Set` is [partially written in Ruby](https://github.com/ruby/ruby/blob/trunk/lib/set.rb).

To demonstrate `Concurrent::Set` is not thread-safe I have created this artificial example:

```ruby
require "concurrent-ruby"

class MySet < Concurrent::Set
  # Note include? is written in Ruby in Set, like delete?
  def include?(o)
    puts "include? call issued, 1 belongs to self, but there is a context switch"
    sleep 2
    super
  end
end

set = MySet[1]

t = Thread.new do
  sleep 1
  puts "clearing the set in a different thread"
  set.clear
end

p set.delete?(1) # should be self, but it is nil

t.join

```

By the time `delete?` is invoked, the set has the element 1, therefore, that method should return `self` (the set). However, it returns `nil` because `clear` was called during a context switch in the middle of the method execution.

fxn:
I have revised the example to have the context switch happen in the middle of the execution of `delete?`.

thedarkone:
Hi @fxn,

you are right, the comment should be dropped (it is incorrect). https://github.com/ruby-concurrency/concurrent-ruby/blob/b20a56296e6ef6dd85d9eb07100271f369787509/lib/concurrent/set.rb#L26-L27

---

Furthermore, currently it looks like Set is not safe to use beyond basic `add`/`remove`/`include?` methods. I knew Set was written in Ruby, and I skimmed through the methods while reviewing #666, it looked fine to me 😞 .

pitr-ch:
@fxn thanks for filing the issue!
@haridutt12 would you be interested in taking a stab at fixing it?

anuragshopify:
I'd like to work on this. What would the fix entail though? Changing the implementation in mri ruby or providing a thread safe interface here?

pitr-ch:
@anuragshopify thank you! I would recommend to inherit from the MRI's Set in concurrent-ruby and fixing the methods in the new class here. Then you could try to contribute it back to MRI.

fzakaria:
[...]

PR #841, opened 2020-01-27
Remove AtExit code

fzakaria:
Following #840 which allowed all threads created by concurrent-ruby
to be marked as daemon -- specifically in JRuby; the "AtExit" handling
is no longer needed.

Delete all references to "AtExit" and replace it with clearer
documentation that the users can select whether the threads started
shall be marked as daemon threads.

This would be considered a _breaking_ change since the API surface is being modified.
I don't expect consumers of _concurrent-ruby_ however to have used it. The reliance on the _:auto-terminate_ option continues to work.

Conceptually, relying on the fact that threads are marked as daemon is much simpler & maps closer to the _ExecutorService_ API which the library seems to model itself after (or the Java concurrent libraries in general).

_Information regarding what it means to mark a thread as daemon:_
> When a Java Virtual Machine starts up, there is usually a single non-daemon thread (which typically calls the method named main of some designated class).
> The Java Virtual Machine continues to execute threads until either of the following occurs:
>
> 1. The exit method of class Runtime has been called and the security manager has permitted the exit operation to take place.
> 2. All threads that are not daemon threads have died, either by returning from the call to the run method or by throwing an exception that propagates beyond the run method.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Current
 
+concurrent-ruby:
+
+* Concurrent.disable_at_exit_handlers! is not longer needed and was deprecated.
```
LikeLakers2:
```suggestion
* Concurrent.disable_at_exit_handlers! is no longer needed and was deprecated.
```

lib/concurrent-ruby/concurrent/configuration.rb
```diff
@@ -96,21 +98,40 @@ def self.global_logger=(value)
...
+  #   E.g. Minitest's test-suite runs itself in `at_exit` callback which
+  #   executes after the pools are already terminated. Then auto termination
+  #   needs to be disabled and called manually after test-suite ends.
+  # @note This method should *never* be called
+  #   from within a gem. It should *only* be used from within the main
+  #   application and even then it should be used only when necessary.
+  # @deprecated Has no effect since it is not longer needed, see https://github.com/ruby-concurrency/concurrent-ruby/pull/841.
+  #
+  def self.disable_at_exit_handlers!
+    deprecated "Method #disable_at_exit_handlers! has no effect since it is not longer needed, see https://github.com/ruby-concurrency/concurrent-ruby/pull/841."
```
fzakaria: thanks -- didn't know there was a _deprecated_ style to the library.

lib/concurrent-ruby/concurrent/configuration.rb
```diff
@@ -149,7 +170,6 @@ def self.executor(executor_identifier)
   def self.new_fast_executor(opts = {})
     FixedThreadPool.new(
         [2, Concurrent.processor_count].max,
-        auto_terminate:  opts.fetch(:auto_terminate, true),
```
fzakaria: @pitr-ch don't you want to pass down the `auto_terminate` option?
I guess someone _may_ want one non-daemonized

CHANGELOG.md
fzakaria: Add @LikeLakers2 suggestion.

lib/concurrent-ruby/concurrent/configuration.rb
```diff
@@ -125,9 +126,10 @@ def self.global_logger=(value)
   # @note This method should *never* be called
   #   from within a gem. It should *only* be used from within the main
   #   application and even then it should be used only when necessary.
+  # @deprecated Has no effect since it is not longer needed, see https://github.com/ruby-concurrency/concurrent-ruby/pull/841.
```
fzakaria:
```suggestion
# @deprecated Has no effect since it is no longer needed, see https://github.com/ruby-concurrency/concurrent-ruby/pull/841.
```

lib/concurrent-ruby/concurrent/configuration.rb
```diff
@@ -125,9 +126,10 @@ def self.global_logger=(value)
...
   #
   def self.disable_at_exit_handlers!
-    AT_EXIT.enabled = false
+    deprecated "Method #disable_at_exit_handlers! has no effect since it is not longer needed, see https://github.com/ruby-concurrency/concurrent-ruby/pull/841."
```
fzakaria:
```suggestion
    deprecated "Method #disable_at_exit_handlers! has no effect since it is no longer needed, see https://github.com/ruby-concurrency/concurrent-ruby/pull/841."
```

lib/concurrent-ruby/concurrent/configuration.rb
pitr-ch: thanks

lib/concurrent-ruby/concurrent/configuration.rb
pitr-ch: Indeed. I missed it does not merge with the opts. Good catch.

fzakaria:
@pitr-ch looking the failures the builds are passing but not terminating.
Looks like one of the JRuby threads are not daemonizing.

lib/concurrent-ruby/concurrent/executor/java_single_thread_executor.rb
```diff
@@ -17,12 +17,13 @@ def initialize(opts = {})
       end
 
       private
-      
+
       def ns_initialize(opts)
-        @executor = java.util.concurrent.Executors.newSingleThreadExecutor
+        @executor = java.util.concurrent.Executors.newSingleThreadExecutor(
```
fzakaria: @pitr-ch this was the missing piece.
Locally my `rake spec` now completes & doesn't hang.
I had to move **DaemonThreadFactory** to the parent class.

fzakaria:
Sweet the **Travis CI** build passed with my latest commit @pitr-ch

lib/concurrent-ruby/concurrent/executor/java_single_thread_executor.rb
pitr-ch: Thank you for investigating and the fix!

lib/concurrent-ruby/concurrent/executor/java_executor_service.rb
```diff
@@ -96,6 +96,8 @@ def newThread(runnable)
         return thread
       end
     end
-    
+
+    private_constant :DaemonThreadFactory
```
fzakaria: can `private_constant` be accessed by subclasses?

pitr-ch: yeah that works fine. (It has to be private otherwise it would show up in the documentation.)

spec/concurrent/executor/cached_thread_pool_spec.rb
```diff
@@ -212,6 +212,8 @@ module Concurrent
               pid       = spawn RbConfig.ruby, test_file
               Process.waitpid pid
               expect($?.success?).to eq true
+            rescue Errno::ECHILD
+              # child already gone
```
fzakaria: Ah! I was also fixing this :)

```ruby
            rescue Errno::ECHILD
              # the child could quit out fast enough before we try to wait for it!
```

I think we could _TODO_ move this test to `executor_service_shared` & have `pool_quits` start every type of executor service: cached & single executor

pitr-ch: yeah that sounds like a good idea! Would you be willing to do it?

fzakaria: done

fzakaria:
Failure does not seem related -- i re-triggered build with empty commit.

```
Failures:
  1) Concurrent::Map updates dont block reads
     Failure/Error: expect(joined_thread).not_to eq nil
     
       expected: value != nil
            got: nil
     
       (compared using ==)

...
     # ./spec/concurrent/map_spec.rb:268:in `block in Concurrent'
```

pitr-ch:
Thanks. I will remove the last commit and merge. (You can also trigger the build again by amending the last commit and force by force-pushing.)

fzakaria:
Just squash merge it @pitr-ch ?

pitr-ch:
Yeah that works as well, I planned `git reset --hard HEAD^; git push -f`. I won't get to it until tomorrow though.

fzakaria:
I just did that & it's using the previous failed check of that commit :(

fzakaria:
Had to amend the HEAD commit anyways

pitr-ch:
Thanks for the PR Farid!

fzakaria:
thanks @pitr-ch 
I'm enjoying working on this project.
Is there another issue you think worth looking at to help?

pitr-ch:
Awesome, thanks for the offer 👍 This project definitely needs more contributors! This https://github.com/ruby-concurrency/concurrent-ruby/issues/796 deserves some attention. Let me know if that interests you and I will assign it to you. If not please have a look at https://github.com/ruby-concurrency/concurrent-ruby/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Alooking-for-contributor

fzakaria:
@pitr-ch on it.

PR #843, opened 2020-01-31
CI: update rake

ahorek:
fixes the ruby-head build

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 require File.join(File.dirname(__FILE__), 'lib/concurrent-ruby/concurrent/version')
 require File.join(File.dirname(__FILE__ ), 'lib/concurrent-ruby-edge/concurrent/edge/version')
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'concurrent-ruby-edge', Concurrent::EDGE_VERSION, options
 gem 'concurrent-ruby-ext', Concurrent::VERSION, options.merge(platform: :mri)
 
 group :development do
-  gem 'rake', '~> 12.0'
+  gem 'rake', (RUBY_VERSION < '2.2.0') ? '~> 12.0' : '~> 13.0'
```
pitr-ch: There is a utility file for this which can be safely loaded here `lib/concurrent-ruby/concurrent/utility/engine.rb`, then `Concurrent.ruby_version :<, 2, 2, 0` can be used. It does proper version comparison.

pitr-ch: Thanks!

ahorek:
thanks Peter!

PR #845, opened 2020-02-02
Ruby 2.7+: delegate kwags explicitely

ahorek:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

fixes
<code>
/home/travis/build/ruby-concurrency/concurrent-ruby/lib/concurrent-ruby-edge/concurrent/edge/erlang_actor.rb:535: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
</code>
<code>
/home/travis/build/ruby-concurrency/concurrent-ruby/lib/concurrent-ruby-edge/concurrent/edge/erlang_actor.rb:492: warning: The called method `spawn_actor' is defined here
</code>


pitr-ch:
Thanks for the fix!

PR #847, opened 2020-02-10
Make CRuby Set thread safe

### LINKED ISSUE #796: Concurrent::Set is not thread-safe in CRuby

fzakaria:
Follow similar pattern for Rbx to wrap every superclass (::Set) instance
method with a Monitor (which is re-entrant).

Included is a test that I believe *would* occasionally fail
if Set was not thread safe.

fixes #796

### DIFF
--- a/lib/concurrent-ruby/concurrent/set.rb
+++ b/lib/concurrent-ruby/concurrent/set.rb
@@ -23,9 +23,21 @@ module Concurrent
   # @!macro internal_implementation_note
   SetImplementation = case
                       when Concurrent.on_cruby?
-                        # Because MRI never runs code in parallel, the existing
-                        # non-thread-safe structures should usually work fine.
-                        ::Set
+                        # The CRuby implementation of Set is written in Ruby itself and is
+                        # not thread safe for certain methods.
+                        require 'monitor'
+                        require 'concurrent/thread_safe/util/data_structures'
+
+                        class CRubySet < ::Set
+                          def initialize(*args)
+                            self.send(:_mon_initialize)
+                            super(*args)
+                          end
+                        end
+
+                        # make use of the same RBX
+                        ThreadSafe::Util.make_synchronized_on_rbx Concurrent::CRubySet
+                        CRubySet
 
                       when Concurrent.on_jruby?
                         require 'jruby/synchronized'
--- a/spec/concurrent/set_spec.rb
+++ b/spec/concurrent/set_spec.rb
@@ -42,7 +42,7 @@ module Concurrent
     end
 
     context 'concurrency' do
-      it do
+      it 'simple' do
         (1..Concurrent::ThreadSafe::Test::THREADS).map do |i|
           in_thread do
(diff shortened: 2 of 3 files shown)


lib/concurrent-ruby/concurrent/set.rb
```diff
@@ -23,9 +23,21 @@ module Concurrent
...
-                        # non-thread-safe structures should usually work fine.
-                        ::Set
+                        # The CRuby implementation of Set is written in Ruby itself and is
+                        # not thread safe for certain methods.
+                        require 'monitor'
+                        require 'concurrent/thread_safe/util/data_structures'
+
+                        class CRubySet < ::Set
+                          def initialize(*args)
+                            self.send(:_mon_initialize)
```
fzakaria: Without this -- certain Sets are being created not through _initialize_
I would see:
```
      when initializing with a block argument
concurrent-ruby/concurrent/thread_safe/util/data_structures.rb:43: warning: instance variable @_monitor not initialized
```

I'm not certain how the Set is being initialized without going through:
```ruby
          def self.new(*args)
            obj = super(*args)
            obj.send(:_mon_initialize)
            obj
          end
```

TBH i'm

spec/concurrent/set_spec.rb
```diff
@@ -55,6 +55,33 @@ module Concurrent
         end.map(&:join)
         expect(set).to be_empty
       end
+
+      it 'force context switch' do
```
fzakaria: writing a test for this is actually not easy lol.

What this ended up testing was that during a method of set it grabs a monitor and locks out any other changes.
```ruby
expect(set.add?(1)).to eq set
```
the above line _should_ occasionally return **nil** if the methods were not synchronized.

pitr-ch:
Replaced by https://github.com/ruby-concurrency/concurrent-ruby/pull/911

PR #852, opened 2020-02-13
Remove un-needed line in TimerTask

fzakaria:
Removed `self.value = value` in TimerTask#timeout_task

This line was probably added as a copy-mistake from the above
method (execute_task).

I ran `brake spec SPEC=spec/concurrent/timer_task_spec.rb` and the tests pass.
I can't see why the need to self-assign the variable here.

co-author: @kalenp

### DIFF
--- a/lib/concurrent-ruby/concurrent/timer_task.rb
+++ b/lib/concurrent-ruby/concurrent/timer_task.rb
@@ -325,7 +325,6 @@ module Concurrent
     def timeout_task(completion)
       return unless @running.true?
       if completion.try?
-        self.value = value
         schedule_next_task
         observers.notify_observers(Time.now, nil, Concurrent::TimeoutError.new)
       end


fzakaria:
Some more notes on TimerTask we are discussing:
1. `try?` on event is confusing or rather has no documentation.
2. there should be explicit mention that TimerTask is a **fixed delay** rather than **fixed rate**
3. If the timeout_interval is shorter than the length of time the task takes; concurrent tasks will run.

pitr-ch:
Thank you for taking an interest. There is ongoing effort to rewrite TimerTask to something more simpler and maintainable https://github.com/ruby-concurrency/concurrent-ruby/pull/829 the code there is incomplete though, I have more changes locally.

pitr-ch:
Please wait until I finish and push the changes I have. Let's address points you had on the new implementation.

fzakaria:
@pitr-ch I think you commented on the wrong PR

pitr-ch:
Actually I did not but the comment is not understandable 😁sorry
I meant to ask you not to invest more time in the TimerTask until I push the changes to the other PR, I am trying to prevent effort duplication.

fzakaria:
Gotcha.

PR #866, opened 2020-04-24
Child promise state not set to :pending immediately after #execute when parent has completed

pelly:
1. Create a parent promise and wait for it to complete
2. Create a child promise chain off the parent (via a call to `then`, `on_success`, etc)
3. Call `execute` on the child promise
4. Immediately check the state of the child promise
5. This state will be `:unscheduled`.

Expectation: It should be `:pending`.


pitr-ch:
Thank you for the fix!

PR #869, opened 2020-05-07
Removes references to static Ruby versions in docs

baweaver:
This change removes references to Ruby 2.2.0 in favor of referencing the
latest version of Ruby. By referencing just `core` rather than
`core-x.y.z` the Ruby docs site will forward to the latest applicable
version.


baweaver:
Documentation will have to be regenerated, but avoided that to keep the diff more minimal.

pitr-ch:
Doc updated in https://github.com/ruby-concurrency/concurrent-ruby/commit/cd59563ed19eae0cd10fa9ed9f76eece3d025cd1

PR #874, opened 2020-06-30
Use default require path for concurrent-ruby gem

### LINKED ISSUE #873: RubyMine integration fails due to custom gem require path
Feel free to close this as 'not a bug', but I thought I'd raise awareness of a RubyMine IDE integration issue I stumbled upon affecting consumers of this dependency: [`RUBY-23350`](https://youtrack.jetbrains.com/issue/RUBY-23350).

The root cause of the issue seems to be that the `concurrent-ruby` gem uses a non-standard require path: https://github.com/ruby-concurrency/concurrent-ruby/blob/7dc6eb04142f008ffa79a59c125669c6fcbb85a8/concurrent-ruby.gemspec#L23

Unfortunately, due to the issue linked above the RubyMine IDE currently does not index these non-standard paths correctly, causing IDE features such as code completion, jump to class definition and quick documentation to not work with this library: 

Obviously it would be great if JetBrains quickly fixed this issue on their end, but this could also be worked around in this gem if the built gem's directory structure was changed to the default `lib` require path. I ~would be happy to attempt~ opened a PR along these lines, please let me know if you would be open to such a change.

Otherwise, I would encourage anyone affected by this issue to encourage JetBrains to update their IDE to prioritize a fix for this bug (e.g., by up-voting the linked issue).

wjordan:
This PR is one approach to refactoring the `concurrent-ruby` gem to use the default `['lib']` require path. This approach bulk-moves all files in `lib/concurrent-ruby/concurrent/` to `lib/concurrent/`, updating all references to match.

Fixes #873.

Note that this doesn't update the require path in `concurrent-ruby-edge` gem as well- that would also be possible if desired, but in order to keep that path separate from `concurrent-ruby` it would need to be moved to a separate `lib` directory, e.g., from `lib/concurrent-ruby-edge/concurrent` to `edge/lib/concurrent/` and `concurrent-ruby-edge.gemspec` moved into `edge/`.


chrisseaton:
Thanks for this sensible idea.

Your PR is heavily conflicted now so I will close this particular PR, but we'll still look to fix this issue.

PR #886, opened 2020-09-20
Fix race condition in TVar for stale reads, fixes #885

### LINKED ISSUE #885: Race condition in TVar
When a `TVar` is first written to, the fact that it's been locked isn't recorded until *after* the check if any previous reads are now stale. If the transaction aborts at this point due to stale reads the `TVar` is left locked.

chrisseaton:


headius:
You've got a rogue `puts` in there...

lib/concurrent-ruby/concurrent/tvar.rb
```diff
@@ -224,6 +227,7 @@ def commit
 
       unlock
 
+      puts 'committed'
```
headius, line 230: Remove. 😀

headius (changes requested):

headius (approved):

Issue #887, opened 2020-09-21
STM guarantees clarification


pitr-ch:
The current implementation seems to allow https://en.wikipedia.org/wiki/Software_transactional_memory#Implementation_issues
see the example below. I think it should be clarified in the readme.

```ruby
require 'concurrent-ruby'

v1 = Concurrent::TVar.new(0)
v2 = Concurrent::TVar.new(0)

2.times.map{
   Thread.new{
     while true
       Concurrent::atomically do
         t1 = v1.value
         t2 = v2.value
         raise [t1, t2].inspect if t1 != t2
       end
     end
   }

   Thread.new{
     100_000.times{|i|
       Concurrent::atomically do
         v1.value += 1
         v2.value += 1
       end
     }
   }
}.each{|t| p t.join}

__END__

$ ~/tmp/jruby-9.2.13.0/bin/jruby -v ~/whome/src/rb/t.rb
jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM
11.0.8+10-post-Ubuntu-0ubuntu120.04 on
11.0.8+10-post-Ubuntu-0ubuntu120.04 +jit [linux-x86_64]
warning: thread "Ruby-0-Thread-3: /home/ko1/whome/src/rb/t.rb:1"
terminated with exception (report_on_exception is true):
RuntimeError: [8, 9]
   /home/ko1/whome/src/rb/t.rb at /home/ko1/whome/src/rb/t.rb:12
                    atomically at
/home/ko1/tmp/jruby-9.2.13.0/lib/ruby/gems/shared/gems/concurrent-ruby-1.1.7/lib/concurrent-ruby/concurrent/tvar.rb:118
                          loop at org/jruby/RubyKernel.java:1442
                    atomically at
/home/ko1/tmp/jruby-9.2.13.0/lib/ruby/gems/shared/gems/concurrent-ruby-1.1.7/lib/concurrent-ruby/concurrent/tvar.rb:108
   /home/ko1/whome/src/rb/t.rb at /home/ko1/whome/src/rb/t.rb:9
```

@chrisseaton would you have some time to do that?

pitr-ch:
The transaction's `#commit` allows the partial reads of the values.

chrisseaton:
[...]

PR #891, opened 2020-10-12
Simplify TVar implementation

chrisseaton:
With the ongoing discussion in #887 I've decided to simplify `TVar` to provide a more robust set of semantics with a bit lower performance and concurrency. I don't think anyone is using `TVar`, but if they do in the future this will at least more more understandable for them.

@ko1 this new implementation provides opacity.


pitr-ch:
Thanks a lot! I hope I'll have some time on the weekend to review.

docs-source/tvar.md
```diff
@@ -24,16 +24,9 @@ We implement nested transactions by flattening.
...
-See:
-
-1.  T. Harris, J. Larus, and R. Rajwar. Transactional Memory. Morgan & Claypool, second edition, 2010.
-
-Note that this implementation allows transactions to continue in a zombie state
-with inconsistent reads, so it's possible for the marked exception to be raised
-in the example below.
+Our implementation uses a very simple algorithm that locks each `TVar` when it
+is first read or written. If it cannot lock a `TVar` it aborts and retries.
+There is no contention manager so competing transactions may retry eternally.
```
pitr-ch, line 29: I am hesitant to merge it as is because of this trade off. It introduces live-lock. Do you know how to fix it?

I am thinking to define an order between all the `TVar`s, therefore, also between all the locks. Then any retrying transaction will pre-lock its visited tvars in that order. That should avoid any live-locks between transactions. However, that assumes that transactions lock same TVars on the second run.

chrisseaton: > I am hesitant to merge it as is because of this trade off. It introduces live-lock. Do you know how to fix it?

I'm sorry there was never any contention manager. I've just strengthened the warning, not weakened the code!

> Then any retrying transaction will pre-lock its visited tvars in that order.

Yes that's a common contention management strategy - to pre-acquire the locks you needed last time in the hope that you'll need them this time.

> That should avoid any live-locks between transactions.

I don't think it's good enough to do that, due to the assumption you mention.

Issue #872, opened 2020-05-27
Concurrent::Collection::RubyNonConcurrentPriorityQueue#delete works wrong


bolshakov:
```
* Operating system:                mac
* Ruby implementation:             Ruby
* `concurrent-ruby` version:       master
```

Here is the test case:

```ruby
def dequeue_all(queue)
  queue.size.times.inject([]) do |acc, _|
    acc << queue.pop
  end
end

queue = described_class.from_list([2, 1, 2, 0, 1, 1, 2], order: :max)

queue.delete(0)

expect(dequeue_all(queue)).to eq([2, 2, 2, 1, 1, 1])
```

It fails with the following error:

```
       expected: [2, 2, 2, 1, 1, 1]
            got: [2, 2, 1, 2, 1, 1]
```

As you can see it does not maintain the oredr after removing an element in some cases.

PR #905, opened 2021-03-27
Fix RubyNonConcurrentPriorityQueue#delete method

andrykonchin:
Close https://github.com/ruby-concurrency/concurrent-ruby/issues/872

The issue is that `#delete` method assumes that deleted element is always "greater" than the last array element and tries to find a proper place in a subtree. But actually the deleted element can be "smaller" and we need to swap it with some parent node.

### Example

Let's consider an example:
[2, 1, 2, 0, 1, 1, 2]

```
   2
  /  \
 1    2
/ \  / \
0 1  1  2
```

After deleting element `0` we have the following incorrect array:

[2, 1, 2, 2, 1, 1]

```
   2
  /  \
 1    2
/ \  /
2 1  1 
```

### DIFF
--- a/lib/concurrent-ruby/concurrent/collection/ruby_non_concurrent_priority_queue.rb
+++ b/lib/concurrent-ruby/concurrent/collection/ruby_non_concurrent_priority_queue.rb
@@ -30,7 +30,7 @@ module Concurrent
           if @queue[k] == item
             swap(k, @length)
             @length -= 1
-            sink(k)
+            sink(k) || swim(k)
             @queue.pop
           else
             k += 1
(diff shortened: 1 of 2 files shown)


andrykonchin:
Some jobs fail on `bundle install` step with error:

> Retrying fetcher due to error (2/4): Bundler::Fetcher::CertificateFailureError Could not verify the SSL certificate for https://rubygems.org/.
> There is a chance you are experiencing a man-in-the-middle attack, but most likely your system doesn't have the CA certificates needed for verification. For information about OpenSSL certificates, see http://bit.ly/ruby-ssl. To connect without using SSL, edit your Gemfile sources and change 'https' to 'http'.

Looks like it's related only to Windows platform on CI and not related to the changes in the PR

pitr-ch:
@andrykonchin Great fix, thank you!

Issue #900, opened 2021-01-26
[documentation] Concurrent::Set documentation is ambiguous, possibly misleading.


grncdr:
The docs for `Concurrent::Set` state (emphasis mine)

> This version locks against the object itself for every method call, ensuring only one thread can be reading or writing at a time. **This includes iteration methods like `#each`.**

However, on CRuby, the implementation of `Concurrent::Set` is just the built-in `Set` ([ref](https://github.com/ruby-concurrency/concurrent-ruby/blob/v1.1.8/lib/concurrent-ruby/concurrent/set.rb#L25-L28)).

Clearly, the built-in set does *not* perform any locking around method calls. This caused a few bugs for me when I used `Concurrent::Set` it for sharing state between threads. I had one thread iterating the set with `#each` and other threads adding/deleting items from the set. This failed, since CRuby Set doesn't support mutating a set during iteration.

Would a PR updating the docs to call this out explicitly be welcome? I would propose something like:

> This version locks against the object itself for every method call, ensuring only one thread can be reading or writing at a time. This includes iteration methods like `#each`.
>
> > **NOTE:** On CRuby this is simply an alias for `Set` and does not support concurrent modification during iteration. Prefer `Concurrent::Map` instead.

```
* Ruby implementation:             Ruby
* `concurrent-ruby` version:       1.1.8, but this applies to all versions afaict
```

pitr-ch:
Hi @grncdr, thanks for the report. It would be better to fix the implementation on CRuby though. It should also lock during the iteration since that is the behaviour of Concurrent::Set on JRuby and TruffleRuby. Could you contribute the fix please?

PR #906, opened 2021-04-20
Make `Concurrent::Set` thread-safe on CRuby

flavorjones:
See #900 for original issue pointing out that `Concurrent::Set#each` (and other methods as well) are not thread-safe because the implementation was short-circuited to be `::Set`.

This PR adds two commits:

- one introducing test coverage for thread-safe usage of `#each` combined with `#add` and `#delete`
- a second introducing a CRuby-specific `Set` implementation that uses `Monitor` to synchronize across all instance methods

This approach, using a new method `ThreadSafe::Util.make_synchronized_on_cruby`, borrows heavily from the approach used for Rubinius (see `ThreadSafe::Util.make_synchronized_on_rbx`).

Please let me know if this approach is too naïve, I fear I may be missing something obvious.

### DIFF
--- a/lib/concurrent-ruby/concurrent/set.rb
+++ b/lib/concurrent-ruby/concurrent/set.rb
@@ -23,9 +23,13 @@ module Concurrent
   # @!macro internal_implementation_note
   SetImplementation = case
                       when Concurrent.on_cruby?
-                        # Because MRI never runs code in parallel, the existing
-                        # non-thread-safe structures should usually work fine.
-                        ::Set
+                        require 'monitor'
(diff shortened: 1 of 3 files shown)


flavorjones:
The test failures don't seem to be related to this change.

chrisseaton:
Yes sorry it looks like the CI has bit-rotted out from underneath us. We should switch to GitHub Actions quickly.

pitr-ch:
Replaced by https://github.com/ruby-concurrency/concurrent-ruby/pull/911

PR #908, opened 2021-05-06
Add test CI flow for various Rubys on Github actions

wildmaples:
## Change

Travis CI has been down for a while now and we'd like to switch to Github Actions. This change introduces a new CI build on Github Actions that runs tests on Ruby implementations and also mark some Ruby failing implementations as experimental so the build still passes.

It does not contain all the requirements from the previous Travis CI yet, but it's a good start. 

#### Currently, tests pass on these Ruby implementations: 
- 2.7, jruby, jruby-head, truffleruby-head

#### These Ruby implementations are marked experimental as they are failing: 
- head (ruby-head), truffleruby

## Missing Functionalities

There are some functionalities that exist in the previous CI (travis CI) that is missing from this change.

- Specifying jdk for JRuby implementations
- Running order of Ruby implementations based on importance/relevance 
- YARD uptodate in docs (`rake yard:master:uptodate`) 
- Running tests on Rubinius (currently, ruby/setup-ruby does not support Rubinius) 

## Questions

Would this change still be useful? 

The `truffleruby-head` is a flaky build, it passes sometimes but fails most. Should we mark it experimental (so the build will still be green if it fails)?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,38 @@
+name: CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.7, jruby, jruby-head, truffleruby-head]
+        experimental: [false]
+        include:
+          - ruby: head
+            experimental: true
+          - ruby: truffleruby
+            experimental: true
+
+    env:
+      JAVA_OPTS: '-Xmx1024m'
+      RUBYOPT: '-w'
+      JRUBY_OPTS: '--dev'
+
+    name: "Tests: Ruby ${{ matrix.ruby }}"
+    steps:
+    - name: Clone Repo
+      uses: actions/checkout@v2
+    - name: Setup system Ruby ${{ matrix.ruby }}
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby }}
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Run tests
+      run: |
+        gem install bundler --version 1.17.3
+        echo JAVA_OPTS: $JAVA_OPTS
+        bundle exec rake ci


chrisseaton (approved):
Thanks - we should indeed probably be switching to Actions and this is a good first step. It will need some iteration over time but we can do that from this start.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Clone Repo
+      uses: actions/checkout@v2
+    - name: Setup system Ruby ${{ matrix.ruby }}
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby }}
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Run tests
+      run: |
+        gem install bundler --version 1.17.3
```
tomstuart, line 36: I don’t know that `gem install bundler` achieves anything really (other than making CI slower). Per the `ruby/setup-ruby` [docs](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically):

>Note that any step doing `bundle install` (for the root `Gemfile`) or `gem install bundler` can be removed with `bundler-cache: true`.

At this point the gems have already been installed so it doesn’t really matter what version of Bundler we then install.

chrisseaton: Yes it's probably good to copy what Travis did first and then simplify and refine.

tomstuart: Perhaps more helpfully: the right way to control the Bundler version is to [provide a `bundler:` input](https://github.com/ruby/setup-ruby/blob/master/action.yml#L12-L18) to the `ruby/setup-ruby` action.

pitr-ch:
Thank you for the migration!
I am not yet familiar with Github actions so I cannot help with it at the moment, though.
I've also tried to request credits for OSS on travis-ci.com, we'll see if they grant us some allowing to stay c-r on travis avoiding the migration.

pitr-ch:
The old CI runs in travis.com with granted OSS credits for now. Let me know if you run into problems.

pitr-ch:
To clarify, If you would like to set up the Github actions based CI to fully replace the Travis based CI I will be for it. However, I won't be able to help with the migration at the moment. Migrating to Travis.com with granted OSS credits was a simple and pragmatic thing to do at the moment to keep the CI running as is even after travis.org shutdown until everything is ready on GH-actions. Thanks again for working on the migration.

.github/workflows/ci.yml
eregon: Indeed, it's counter-productive to install bundler here, and will probably be more confusing than helpful.
Also Bundler 1 is basically not really supported on Ruby 2.7+ (no longer maintained AFAIK).

eregon: => https://github.com/ruby-concurrency/concurrent-ruby/pull/916

PR #915, opened 2021-06-25
`Concurrent.monotonic_time` accept an optional unit parameter

casperisfine:
This is a direct transposition of MRI's `clock_gettime` unit option.

Tis is a parameter I use quite a lot when writing MRI specific code, but that I miss when using `Concurrent.monotonic_time` for portability.

More often than not, when you reach for monotonic time, you want to minimize the overhead, so it's appreciable to be able to avoid a `/ 1_000` when you want milliseconds in the first place.

NB: I'm not too sure how a good way to test the scale of the return value would be.

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/monotonic_time.rb
+++ b/lib/concurrent-ruby/concurrent/utility/monotonic_time.rb
@@ -2,26 +2,64 @@ require 'concurrent/synchronization'
 
 module Concurrent
 
-  class_definition = Class.new(Synchronization::LockableObject) do
-    def initialize
-      @last_time = Time.now.to_f
-      super()
+  # @!macro monotonic_get_time
+  #
+  #   Returns the current time a tracked by the application monotonic clock.
+  #
+  #   @param [Symbol] unit the time unit to be returned, can be either
+  #     :float_second, :float_millisecond, :float_microsecond, :second,
+  #     :millisecond, :microsecond, or :nanosecond default to :float_second.
+  #
+  #   @return [Float] The current monotonic time since some unspecified
+  #     starting point
+  #
+  #   @!macro monotonic_clock_warning
+  if defined?(Process::CLOCK_MONOTONIC)
+
+    def monotonic_time(unit = :float_second)
+      Process.clock_gettime(Process::CLOCK_MONOTONIC, unit)
     end
 
-    if defined?(Process::CLOCK_MONOTONIC)
-      # @!visibility private
-      def get_time
-        Process.clock_gettime(Process::CLOCK_MONOTONIC)
+  elsif Concurrent.on_jruby?
+
+    # @!visibility private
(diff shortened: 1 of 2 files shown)


chrisseaton:
When is `defined?(Process::CLOCK_MONOTONIC)` false?

It's not false on JRuby 9.2.17.0 for example. `Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)` works fine on JRuby, TruffleRuby, and I'm guessing not even Rubinius but it's hard to build these days.

@headius do you know if this is some kind of ancient fallback?

casperisfine:
Older MRI I think. `Process.clock_gettime` was added in 2.3 if my memory serves me right.

chrisseaton:
Oh right didn't realise it was so recent. I will check against older versions and build up a matrix of what is supported where.

lib/concurrent-ruby/concurrent/utility/monotonic_time.rb
```diff
@@ -10,25 +10,52 @@ def initialize
...
-      def get_time
-        Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      def get_time(unit)
+        Process.clock_gettime(Process::CLOCK_MONOTONIC, unit)
       end
     elsif Concurrent.on_jruby?
       # @!visibility private
-      def get_time
-        java.lang.System.nanoTime() / 1_000_000_000.0
+      self::TIME_UNITS = Hash.new { |_hash, key| raise ArgumentError, "unexpected unit: #{key}" }.compare_by_identity
```
chrisseaton, line 18: Looking at CI I don't think this syntax worked the way it does now on Ruby 1.9. Do you want to look at that and refactor so it works on older Rubies?

chrisseaton:
Seems good to merge when CI passes on older Rubies.

I'll look at the TruffleRuby failure that's unrelated.

lib/concurrent-ruby/concurrent/utility/monotonic_time.rb
casperisfine: Yes, definitely plan to, it's just getting hard to have such old rubies working locally. 

I'll fix that tomorrow and ping you back.

chrisseaton: I don't know if supporting Ruby 1.9 is an intelligent thing to do anymore?

casperisfine: I'd say no, but I have no idea what the support policy of the gem is.

casperisfine:
@chrisseaton I believe I fixed the CI (well it has various failures but unrelated to the current PR).

I also think the implementation is now simpler, as for the simple case of `Process.clock_getime` being present we save one method call.

chrisseaton (approved):

chrisseaton:
I'm sorry that CI is in such a mess. We really need to get a grip of it.

PR #916, opened 2021-07-16
Cleanup GitHub Actions workflow

eregon:
* No need to install Bundler a second time, it was also unused.
* JAVA_OPTS are already shown in the log.


chrisseaton:
These are all existing failures aren't they? We need to get on top of this...

chrisseaton (approved):

chrisseaton:
Can you please rebase?

chrisseaton:
@eregon can you rebase?

eregon:
Rebased, let's check if the CI is still happy, should be since it's only cleanups.

PR #918, opened 2021-08-07
Fix Concurrent.processor_count link

bolshakov:
[ProcessorCounter](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/utility/processor_counter.rb) annotated as `@!visibility private` and hence is not indexed by YARD. Thus, I guess the line should refer to the `Concurrent.processor_count` method.


chrisseaton:
Thanks!

PR #919, opened 2021-08-22
can get write lock after read locked more than once only by myself

turnon:
The condition `@Count == @HeldCount` does not correctly indicate that WE ARE THE ONLY READER.

Because when we lock for read more than once, `@HeldCount` increases but `@Count` dosen't.

```ruby
lock.with_read_lock do
  lock.with_read_lock do
    # blocked here, can not acquire write lock ! Even though no else reader
    lock.with_write_lock {}
  end
end
```

So we just compare `@Count == 1` and try to set RUNNING_WRITER bit on


chrisseaton:
Can you please rebase?

turnon:
> Can you please rebase?

Ok, just rebased

chrisseaton (approved):

PR #920, opened 2021-09-14
remove exec bit from non scripts

kbrock:
a build process picked this up.
it is minor but seems like good practices


chrisseaton (approved):

chrisseaton:
Thanks. Our CI is a mess at the moment, so ignore that.

kbrock:
> Thanks. Our CI is a mess at the moment, so ignore that.

thanks I was digging into it but was coming up empty

The change is trivial. Appreciate you humoring it.

PR #921, opened 2021-09-16
Typo fix in Event docs

andrewdotn:
This also rearranges the output order to highlight the case where t1 will wait before t2 calls set.


chrisseaton:
Thank you for this contribution.

PR #922, opened 2021-09-18
Fix CI and introduce experimental rubies scheduled job

wildmaples:
# Issue

CI is broken on concurrent-ruby. There are also multiple CI pipeline running including Github Actions. In order to keep it simple, I propose sticking with GH Actions.

The current GH Action failures are due to the fact that experimental rubies are being run on the CI. These rubies, `ruby-head`, `truffleruby` and `truffleruby-head` are already failing previously (see [old travisCI build](https://github.com/ruby-concurrency/concurrent-ruby/runs/3626793982)).

Unfortunately, Github Actions doesn't have an option to [allow failure like TravisCI does](https://github.com/actions/toolkit/issues/399#issuecomment-607465968), so all PRs remain red despite marking the rubies as experimental.  

# Proposed Fix

In order to patch these issues and get CI to a reliable green checkmark for each PR, I decided to split up the experimental rubies into a separate scheduled job and keep the remaining rubies on the primary CI that runs on push and PR. That way, people are not confused by the red checkmarks in PRs (from the experimental rubies) and we can "stop the bleeding".  

I've also deleted the old CI pipelines to avoid confusion. 

This is not meant to be a permanent fix. Once we fix the failures from the experimental rubies, we can add them back to the primary CI pipeline.


wildmaples:
The `jruby-head` ruby build is flaky -- I can't rerun the build myself but rerunning it should do the trick.

chrisseaton:
Yes - Travis is clearly a dead-end now. I don't know about Appveyor but I don't see it commonly either.

chrisseaton:
There's still something expecting Travis to run - this might be something we need to fix-up after or will fix after merge.

Screenshot 2021-09-19 at 23 06 38

chrisseaton:
We used to test back to 1.9. I don't know if that's a goal anymore - supporting 2.5 means we cover latest Debian Old Stable, which is pretty extraordinarily old, and the previous Ubuntu LTS, and even the system macOS Ruby.

PR #924, opened 2021-09-20
build: Trust CI system to have a Bundler

olleolleolle:
This PR cleans up an old, now-duplicate installation of Bundler.

We are using the `bundler-cache` feature of ruby/setup-ruby, which does the installing and caching of gems. Perhaps we can skip installing a Bundler.


olleolleolle:
@chrisseaton I think this could go in.

PR #925, opened 2021-10-22
Fix setting AtomicReferenceImplementation on TruffleRuby

eregon:
* This accidentally worked before because until Ruby 2.7 `alias_method` returns the receiver.
* In Ruby 3 however, `alias_method` returns the first argument.

```
$ ruby -ve 'p class C; alias_method :id, :object_id; end'
ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux]
C
$ chruby 3.0.2                                           
$ ruby -ve 'p class C; alias_method :id, :object_id; end'
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
:id
```

This is blocking to implement the change about `alias_method`;s return value in https://github.com/oracle/truffleruby/pull/2499.
Could you make a bug fix release with this fix soon if possible?

cc @pitr-ch

### DIFF
--- a/lib/concurrent-ruby/concurrent/atomic/atomic_reference.rb
+++ b/lib/concurrent-ruby/concurrent/atomic/atomic_reference.rb
@@ -170,6 +170,7 @@ module Concurrent
                                       alias_method :compare_and_swap, :compare_and_set
                                       alias_method :swap, :get_and_set
                                     end
+                                    TruffleRubyAtomicReference
                                   when Concurrent.on_rbx?
                                     # @note Extends `Rubinius::AtomicReference` version adding aliases
                                     #   and numeric logic.


eregon:
@pitr-ch or other concurrent-ruby maintainers, could you merge this?

jdantonio (approved):

eregon:
Thank you!

It would be great to have a release including this when maintainers have some time.
For now I made a patch in TruffleRuby (https://github.com/oracle/truffleruby/commit/5ea0bc28c67938922fb1ce1a7ec868958bfea55d), but it is quite messy, so a release of `concurrent-ruby` would be much better so the patch is not needed/has no effect when using latest `concurrent-ruby`.

jdantonio:
I've been out of the loop for a long time (I only lurk) so I'm hesitant to do a release myself. I'd be fine giving you and @chrisseaton access to RubyGems so that you can both make releases, but I think I'd be overstepping my bounds a bit right now. Hopefully @pitr-ch can jump in soon and give hi sthoughts.

PR #927, opened 2021-11-02
Support kwargs in initializer when mixing in Concurrent::Async

mlarraz:
Classes with keyword arguments in their initializers cannot be initialized today if they include `Concurrent::Async`, as of Ruby 3.0.


jdantonio (approved):

jdantonio:
@mlarraz Can you please revisit this and add the necessary guards to allow this to build on all supported Ruby versions? Thank you!

jdantonio (approved):

PR #930, opened 2022-01-07
Add Ruby 3.1 to CI

petergoldstein:


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.4, 2.5, 2.6, 2.7, 3.0, jruby, jruby-head]
+        ruby: [2.4, 2.5, 2.6, 2.7, '3.0', 3.1, jruby, jruby-head]
```
Paprikas, line 11: @petergoldstein why quotes?

petergoldstein: Quotes are required around 3.0 because of how YAML parses floats.  Without quotes, 3.0 gets treated as 3 - the additional precision is lost.  And so It will attempt to load the latest Ruby version matching '3', which will be a 3.1 version.

To ensure 3.0 actually tests a Ruby 3.0.x you need to wrap it in quotes.

petergoldstein:
I'm not sure exactly why jruby-head is failing, but it appears unrelated.

chrisseaton:
Thanks for the contribution. Sorry yes CI isn't always in a great state.

Issue #931, opened 2022-01-08
undefined method `exception' for nil:NilClass with Ruby 3.1.0


aaronjensen:
```
* Operating system:                 mac 
* Ruby implementation:             Ruby 3.1.0
* `concurrent-ruby` version:       1.1.9
* `concurrent-ruby-ext` installed:  no
* `concurrent-ruby-edge` used:     no
```

After upgrading to Ruby 3.1.0 I'm getting a very strange, very hard (for me) to debug error with premailer+sprockets that ultimately manifests with concurrent failing to provide a reason for a promise rejection.

The error is strange enough that it seems like it could be a ruby bug.

When stepping through with byebug I get here:

https://github.com/rails/sprockets/blob/v4.0.2/lib/sprockets/manifest.rb#L143

Then, drilling into `asset.source`, it returns the source fine. The `yield` however causes a jump all the way out of the `begin` in `SafeTaskExecutor#execute`. If I put an `ensure` on that begin, it runs. Nothing after `@task.call` in the block runs. No exception can be caught. 

I then get this:

```
        project/gems/ruby/3.1.0/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/obligation.rb:128:in `exception': undefined method `exception' for nil:NilClass

        reason.exception(*args)
              ^^^^^^^^^^ (NoMethodError)
                from project/gems/ruby/3.1.0/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/obligation.rb:87:in `raise'
                from project/gems/ruby/3.1.0/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/obligation.rb:87:in `block in wait!'
                from <internal:kernel>:90:in `tap'
                from project/gems/ruby/3.1.0/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/obligation.rb:87:in `wait!'
                from project/gems/ruby/3.1.0/gems/sprockets-4.0.2/lib/sprockets/manifest.rb:141:in `each'
                from project/gems/ruby/3.1.0/gems/sprockets-4.0.2/lib/sprockets/manifest.rb:141:in `find'
                from project/gems/ruby/3.1.0/gems/sprockets-4.0.2/lib/sprockets/manifest.rb:153:in `each'
[...]

PR #932, opened 2022-01-13
SafeTaskExecutor respects local jump errors

aaronjensen:
See: https://bugs.ruby-lang.org/issues/18474
And: https://github.com/ruby-concurrency/concurrent-ruby/issues/931

### DIFF
--- a/lib/concurrent-ruby/concurrent/executor/safe_task_executor.rb
+++ b/lib/concurrent-ruby/concurrent/executor/safe_task_executor.rb
@@ -16,10 +16,10 @@ module Concurrent
 
     # @return [Array]
     def execute(*args)
-      synchronize do
-        success = false
-        value   = reason = nil
+      success = true
+      value   = reason = nil
 
+      synchronize do
         begin
           value   = @task.call(*args)
           success = true
@@ -27,9 +27,9 @@ module Concurrent
           reason  = ex
           success = false
         end
-
-        [success, value, reason]
       end
+
+      [success, value, reason]
     end
   end
 end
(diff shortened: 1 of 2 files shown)


aaronjensen:
A LocalJumpError will cause the synchronize block to break, meaning that result will never be set and success will never be set to true. Furthermore, it is not caught by the rescue so no reason is caught. 

I consider it a “success” by setting success to true initially. It will only be considered a failure if an exception is caught. 

Even though the synchronize block is short circuited, the return still evaluates, so this completes the promise with a success and nil value.

chrisseaton:
Note that JRuby has a bug with local jumps in general, which will also impact this functionality.

```ruby
task = -> { yield 42 }
puts :here
```

This is fixed in their master branch.

lib/concurrent-ruby/concurrent/executor/safe_task_executor.rb
```diff
@@ -16,20 +16,20 @@ def initialize(task, opts = {})
...
           success = true
         rescue @exception_class => ex
           reason  = ex
           success = false
         end
-
-        [success, value, reason]
       end
+
+      [success, value, reason]
```
eregon, line 32: I don't understand this change, it should never make a difference as far as I can tell.
All possibilities are:
* @task.call does not raise an exception, so `success = true`
* @task.call raises something caught by `rescue @exception_class`, so `success = false`
* @task.call raises something NOT caught by `rescue @exception_class` (e.g., like a break or return or exception higher in the hirearchy), then there should be no way we would ever reach the end of this method, or even the end of the synchronize block.

So the only case I could see this would make a difference is a bug in CRuby where it break or return to the semantically incorrect place, such as after that synchronize call (which should only be possible with a `break` syntactically inside the synchronize block, and there isn't any).

That said, this change is not hurting either, so I think it's fine to keep the change, but we should have a better understanding why it's needed, and a clearer test (see https://github.com/ruby-concurrency/concurrent-ruby/issues/931#issuecomment-1069094690).

PR #936, opened 2022-01-31
Move jruby-head to the experimental workflow

eregon:
It fails currently, probably due to some issues with kwargs: https://github.com/ruby-concurrency/concurrent-ruby/runs/5010387674?check_suite_focus=true.
Other head rubies are already in the experimental workflow.


chrisseaton:
Followed up by 0fe5618a42a59b2130fb94196c5c62484f0e37ad to make it manually runnable.

PR #939, opened 2022-03-13
caller_runs execution should not block the work queue

### LINKED ISSUE #933: Worker threads get stuck when using caller_runs policy
I'm attempting to use a threadpool to implement bounded concurrency with the the `caller_runs` fallback_policy to add back pressure. This doesn't work as I'd expect, after the queue is full and the fallback policy is triggered subsequent jobs all get processed on the caller thread rather than being delegated to the workers.

Reproduction script:
```ruby
require 'concurrent-ruby'

tasks = (1..10).to_a

threadpool = Concurrent::FixedThreadPool.new(
  1,
  max_queue: 2,
  fallback_policy: :caller_runs
)

tasks.each do |t|
  threadpool.post do
    puts "Task #{t} running on '#{Thread.current.name || 'main'}'"
    sleep 1
    puts "Task #{t} complete"
  end
end

threadpool.shutdown
threadpool.wait_for_termination
```

Output:
```
Task 4 running on 'main'
Task 1 running on 'worker-1'
Task 1 complete
Task 4 complete
Task 5 running on 'main'
Task 5 complete
Task 6 running on 'main'
Task 6 complete
Task 7 running on 'main'
Task 7 complete
Task 8 running on 'main'
Task 8 complete
Task 9 running on 'main'
Task 9 complete
Task 10 running on 'main'
Task 10 complete
Task 2 running on 'worker-1'
Task 2 complete
Task 3 running on 'worker-1'
Task 3 complete
```

I suspect this is because the threadpool remains locked while the caller task is running, preventing worker threads from continuing. Adding a `sleep` between posting each task to the pool unblocks it and allows worker threads to continue processing.

I don't know if this is considered a bug or not, but I found it very surprising and can't see an obvious reason it needs to work this way, I'd expect the worker threads to be able to pull items from the queue regardless of how long the caller task blocks.

```
* Operating system:                   linux and mac
* Ruby implementation:             Ruby 
* `concurrent-ruby` version:       1.1.9
* `concurrent-ruby-ext` installed:  no
* `concurrent-ruby-edge` used:    no
```

chrisseaton:
`caller_runs` execution was being done while the executor lock was held. This caused the entire executor’s work queue to be blocked, with worker threads not able to take new work. We’ve restructured so that the execution is done as deferred work, outside of the lock.

Fixes #933.

### DIFF
--- a/lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb
+++ b/lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb
@@ -16,10 +16,19 @@ module Concurrent
 
     def post(*args, &task)
       raise ArgumentError.new('no block given') unless block_given?
+      deferred_action = nil
       synchronize do
         # If the executor is shut down, reject this task
-        return handle_fallback(*args, &task) unless running?
-        ns_execute(*args, &task)
+        unless running?
+          deferred_action = fallback_action(*args, &task)
+          break
+        end
+        deferred_action = ns_execute(*args, &task)
+      end
+
+      if deferred_action
+        deferred_action.call
+      else
         true
       end
     end
--- a/lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb
+++ b/lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb
@@ -16,16 +16,13 @@ module Concurrent
 
     def post(*args, &task)
       raise ArgumentError.new('no block given') unless block_given?
-      deferred_action = nil
-      synchronize do
-        # If the executor is shut down, reject this task
-        unless running?
-          deferred_action = fallback_action(*args, &task)
-          break
+      deferred_action = synchronize {
+        if running?
+          ns_execute(*args, &task)
+        else
+          fallback_action(*args, &task)
         end
-        deferred_action = ns_execute(*args, &task)
-      end
-
+      }
       if deferred_action
         deferred_action.call
       else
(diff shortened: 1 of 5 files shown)


chrisseaton:
Ignore the JRuby failures - they're unrelated. Fixing on another PR.

lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb
```diff
@@ -16,10 +16,19 @@ def initialize(*args, &block)
...
     def post(*args, &task)
       raise ArgumentError.new('no block given') unless block_given?
+      deferred_action = nil
       synchronize do
         # If the executor is shut down, reject this task
-        return handle_fallback(*args, &task) unless running?
-        ns_execute(*args, &task)
+        unless running?
+          deferred_action = fallback_action(*args, &task)
+          break
```
eregon, line 24: `break` looks weird to me here, how about a simply if/else?
Also assuming `synchronize` returns the value of the block it could be `deferred_action = synchronize do`

eregon (commented):
Looks good, just 1 comment

lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb
chrisseaton: Fixed.

chrisseaton:
Please re-review @eregon.

eregon (approved):

PR #942, opened 2022-03-13
Prep for v1.1.10 release

chrisseaton:


chrisseaton:
- [ ] Change log
- [ ] Test on some example applications
- [ ] See what can be done easily to improve TruffleRuby support
- [ ] See what can be done easily to improve JRuby support

chrisseaton:
There's a little JRuby flakiness - possibly bad tests or arguable MRI bugs - but I don't think this is anything new in this release, so I've ticked off JRuby as ready go. @headius

chrisseaton:
TruffleRuby is waiting on #943.

chrisseaton:
I'll likely be releasing this later today.

headius:
Which tests specifically appear to be flaky on JRuby? I have run into some tests here that are bad or that are sensitive to threads from previous tests not shutting down immediately, but I'm not sure of the current status. We do not currently have any known bugs that should affect this library.

chrisseaton:
There's a small number of tests that are either flaky on JRuby or are written with false assumptions based on MRI. We've just don't a new release to clear the backlog, so I now have some time to investigate. I will involve you in what we find.

PR #943, opened 2022-03-13
Improve TruffleRuby compatibility

chrisseaton:


lib/concurrent-ruby/concurrent/configuration.rb
```diff
@@ -1,7 +1,14 @@
...
 require 'concurrent/delay'
 require 'concurrent/errors'
-require 'concurrent/atomic/atomic_reference'
+
+if Concurrent.on_truffleruby?
+  # Avoid TruffleRuby's patch
+  require File.expand_path('atomic/atomic_reference', __dir__)
+else
+  require 'concurrent/atomic/atomic_reference'
+end
```
eregon, line 10: This workaround is unnecessary if we use truffleruby-head or can wait the next release

eregon: It's https://github.com/oracle/truffleruby/issues/2605 BTW

eregon:
From the look of some failures on https://github.com/ruby-concurrency/concurrent-ruby/pull/943 some seem due to kwargs handling, so I'll get that work merged in truffleruby and then look at this again.

lib/concurrent-ruby/concurrent/configuration.rb
chrisseaton: What is the actual issue with the patch? I couldn't find it out.

chrisseaton: I more mean - why do we need the patch anymore?

eregon: TruffleRuby needs the patch for older versions of concurrent-ruby, which seems important they still work.
The patch should have no effect whatsoever on newer concurrent-ruby versions, so it's fine to "always apply it".

This workaround OTOH feels hacky and might not work reliably so I don't want to merge that.

.github/workflows/experimental.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby: [head, jruby-head, truffleruby, truffleruby-head]
+        ruby: [head, jruby-head, truffleruby-head]
```
chrisseaton, line 15: @eregon surely it's in either one or the other but not both?

eregon: The idea is it'd be `truffleruby` again in `ci.yml` once that's good enough to pass CI, in my understanding this is just a draft PR to get any version of TruffleRuby to pass the CI.

eregon (approved):
@chrisseaton This is ready now, and the CI [passed](https://github.com/ruby-concurrency/concurrent-ruby/runs/5573808150?check_suite_focus=true)

chrisseaton:
GitHub is back up, still some CI failures. I want to release tomorrow if I can. What do you think?

eregon:
I pushed what I had locally now that GitHub works.

In any case, feel free to release, I think concurrent-ruby works fine on truffleruby (and not worse than before).
Regarding #931 I'll take a look to improve the test, but I don't think it's a release blocker either.

eregon:
All green, I'm merging.

If there are new flaky test runs we can skip tests like in [e4b4ff2](https://github.com/ruby-concurrency/concurrent-ruby/pull/943/commits/e4b4ff287cdd43b0147cacf98a51561c1b317058), or if it's not specific tests then temporarily remove it from the push/PR CI (it's already run daily anyway) + open issue to investigate them.

PR #946, opened 2022-03-15
Travis is done and never coming back

chrisseaton:


spec/spec_helper.rb
```diff
@@ -2,14 +2,7 @@
   require 'simplecov'
   require 'coveralls'
 
-  if ENV['TRAVIS']
```
eregon, line 5: Should it be `ENV['CI']` instead?

eregon (approved):

spec/spec_helper.rb
chrisseaton: The branch hasn't been run the whole time we've been using GitHub Actions. If we want a different coverage output format in the future we can pick it then.

spec/concurrent/actor_spec.rb
```diff
@@ -148,7 +148,7 @@ def on_message(message)
           end
         end
 
-        it 'terminates with all its children', notravis: true do
```
eregon, line 151: I wonder if this tag was somehow avoiding running transient tests in CI.
Maybe the failure in https://github.com/ruby-concurrency/concurrent-ruby/runs/5553488676?check_suite_focus=true is related (but maybe not)

eregon: Given the Rakefile above it sounds like they were excluded, even on GitHub Actions before this PR

PR #951, opened 2022-03-15
Update information on supported Ruby versions

chrisseaton:
Let's commit to 2.4.10 as a minimum.

That lets us run on:

* macOS system Ruby
* Previous LTS Ubuntu
* Debian old stable

That's what we're actually testing in practice, so what we can genuinely claim. It's two major versions of unmaintained.

I'll say we support latest JRuby and TruffleRuby, which seems to be the case in practice based on what's being test. We should maybe add more JRuby versions to CI and keep supporting a few old versions. @headius for any opinion.

I'll leave the Rubinius code in for now, because I don't want to rip out quite a lot of code in this release, but I've warned off that it will likely go in the future.

@jeremyevans


chrisseaton:
> In general, the required_ruby_version should be set to the minimum Ruby version that will run the gem, not the minimum version you support.

Aren't these the same thing? If we aren't running a Ruby version in CI then I don't know if the gem will run on it or not.

jeremyevans:
> > In general, the required_ruby_version should be set to the minimum Ruby version that will run the gem, not the minimum version you support.
> 
> Aren't these the same thing? If we aren't running a Ruby version in CI then I don't know if the gem will run on it or not.

If you don't know, be optimistic, not pessimistic, and leave the required version at 1.9.3.  If you get a report that it doesn't work on an unsupported version, then bump required_ruby_version at that point.  required_ruby_version should indicate you are 100% sure the gem doesn't work in earlier versions, because of the Ruby features it uses.

If you want, I can test CI for you on earlier Ruby versions (GitHub CI supports back to 1.9.3) to tell you when it actually fails.

chrisseaton:
I think I will leave 2.2 and above running in CI, and mark the gem as supporting 2.2 and above, as that is the factual reality.

Can re-address in a major release.

concurrent-ruby-edge.gemspec
```diff
@@ -23,7 +23,7 @@ be obeyed though. Features developed in `concurrent-ruby-edge` are expected to m
 Please see http://concurrent-ruby.com for more information.
   TXT
 
-  s.required_ruby_version = '>= 1.9.3'
+  s.required_ruby_version = '>= 2.2.10'
```
eregon, line 26: Should be just `2.2` as Jeremy said, it's very uncommon to specify 3rd component for required_ruby_version, and this could be a problem for an alternative implementation which might not use the 3rd component in the same or it might not matter (unlikely for this case since 2.2 is old, but still)

eregon:
> I think I will leave 2.2 and above running in CI, and mark the gem as supporting 2.2 and above, as that is the factual reality.

Agreed.

> Can re-address in a major release.

IMHO no need for a major release, that's the point: `required_ruby_version` would automatically prevent installing a newer concurrent-ruby which needs a newer Ruby version on an older Ruby version. Specifically this requirement is an issue as major releases of a gem are rare and lots of work, and completely unrelated to the pace of CRuby releases, so they should IMHO not be tight together.

eregon:
One argument for requiring 2.3 minimum is Thread#{name,name=} support.
But it seems not hard to keep working around the lack of it in 2.2.

eregon:
We will require Ruby 2.3 in the next release: https://github.com/ruby-concurrency/concurrent-ruby/pull/975

PR #954, opened 2022-03-18
Only run the CRuby bug-specific test on CRuby

eregon:
* See https://github.com/ruby-concurrency/concurrent-ruby/issues/931#issuecomment-1072579667


eregon:
The failures seem unrelated transients.

chrisseaton (approved):

PR #955, opened 2022-03-18
Cleanup dead code for Ruby < 2.2

eregon:


eregon:
The only ruby version guard left I could find is: https://github.com/ruby-concurrency/concurrent-ruby/blob/8ffa7bbb75a0065246ea7b3ca82ee59a385da83b/spec/concurrent/executor/ruby_thread_pool_executor_spec.rb#L160

chrisseaton (approved):

PR #961, opened 2022-08-19
Fixes/improvements to promises.in.md.

keithrbennett:
Great documentation, thank you!

I noticed some things that I thought were unintended, and "fixed" them. ;)

### DIFF
--- a/docs-source/promises.in.md
+++ b/docs-source/promises.in.md
@@ -117,31 +117,31 @@ Let's define an inspection helper for methods.
 ```ruby
 def inspect_methods(*methods, of:)
   methods.reduce({}) { |h, m| h.update m => of.send(m) }
-end #
+end
 ```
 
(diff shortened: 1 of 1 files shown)


chrisseaton:
Does anyone have a clue why we'd have written `#` at the end of these lines like this? Must have been some reason...

chrisseaton:
These `#` are definitely supposed to be doing something - they're stripped when the file is processed but I have no idea why.

pitr-ch:
If I remember correctly. They tell the tool not to add the value of the expression at the end as it does for all the other lines.

keithrbennett:
@pitr-ch What tool is that?

pitr-ch:
It uses https://github.com/pitr-ch/md-ruby-eval as far as I remember.

chrisseaton:
Closed based on explanation. Thanks for the thought, though.

Issue #967, opened 2022-10-24
Semaphore doesn't raise exception when negative int passed to constructor


flash-gordon:
The [docs](https://ruby-concurrency.github.io/concurrent-ruby/1.1.4/Concurrent/Semaphore.html#initialize-instance_method) say it'll raise an error but it doesn't:
```ruby
Concurrent::Semaphore.new(-1)
```
I use cruby.
There's no check for negative number in the code:
https://github.dev/ruby-concurrency/concurrent-ruby/blob/385156117c110808875e1e87d5267147f696f796/lib/concurrent-ruby/concurrent/atomic/mutex_semaphore.rb#L12

eregon:
Indeed, could you make a PR to add the check?

eregon:
Looking more into this, it seems it's intended to allow a negative initial count since https://github.com/ruby-concurrency/concurrent-ruby/commit/736ebd16ff9570534527cf04c1af81ce7e2541bc, which also adds a spec for `Concurrent::Semaphore.new(-1)`.
So the docs are wrong and I'll update that.
Semaphore can go negative, that seems a common feature of semaphores.

eregon:
Docs are up to date now at https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Semaphore.html#initialize-instance_method

PR #973, opened 2022-12-21
Ensure positive int passed to `Concurrent::Semaphore` constructor

khaile:
Address #967


eregon:
Thank you for the PR, that made me realize the behavior is intended and instead the docs need to be updated (I'll do that), see https://github.com/ruby-concurrency/concurrent-ruby/issues/967#issuecomment-1361182570.

PR #975, opened 2022-12-21
Require Ruby 2.3+ and drop Ruby 2.2 support

eregon:
* Set `required_ruby_version` to 2.3.
* `rb_data_object_alloc()` is deprecated since a long time by `rb_data_object_wrap()` but that is [not available in Ruby 2.2](https://github.com/ruby/ruby/blob/ruby_2_2/include/ruby/ruby.h#L1218), so drop Ruby 2.2 support.

Example failed CI run on 2.2:
https://github.com/ruby-concurrency/concurrent-ruby/actions/runs/3749315062/jobs/6367656392
```
compiling ../../../../ext/concurrent-ruby-ext/atomic_reference.c
../../../../ext/concurrent-ruby-ext/atomic_reference.c: In function ‘ir_alloc’:
../../../../ext/concurrent-ruby-ext/atomic_reference.c:47:10: warning: implicit declaration of function ‘rb_data_object_wrap’; did you mean ‘rb_data_object_get’? [-Wimplicit-function-declaration]
   47 |   return rb_data_object_wrap(klass, (void *) Qnil, ir_mark, NULL);
      |          ^~~~~~~~~~~~~~~~~~~
      |          rb_data_object_get
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.2, 2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, jruby, truffleruby]
+        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, jruby, truffleruby]
 
     env:
       JAVA_OPTS: '-Xmx1024m'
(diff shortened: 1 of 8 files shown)


eregon:
Bundler 2 also needs Ruby 2.3+ so this also simplifies the experience for local development (also 2.2 is quite messy to install).

Issue #963, opened 2022-10-21
`Promises.any_fulfilled_future` does not accept an Event


hanazuki:
Documentation on `Promises.any_fulfilled_future_on` says:

> If event is supplied, which does not have value and can be only resolved, it's represented as :fulfilled with value nil.

But actually it fails when you pass an Event to the function.

Reproducing code:
```ruby
require 'concurrent'
p Concurrent::VERSION

p Concurrent::Promises.any_fulfilled_future(Concurrent::Promises.resolved_event).value!  # => Expect nil
```

Output:
```
"1.1.10"
~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:2078:in `resolvable?': undefined method `fulfilled?' for #<Concurrent::Promises::Event:0x00007f672f3cfa30 resolved> (NoMethodError)

        future.fulfilled? ||
              ^^^^^^^^^^^
        from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:1658:in `on_blocker_resolution'
        from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:808:in `callback_notify_blocked'
        from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:786:in `call_callback'
        from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:751:in `add_callback'
        from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:732:in `add_callback_notify_blocked'
        from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/promises.rb:1637:in `block in new_blocked_by'
[...]

PR #976, opened 2022-12-21
Let `Promises.any_fulfilled_future` take an `Event`

hanazuki:
This patch teaches `Promises.any_fulfilled_future` to treat a resolved `Event` as a fulfilled `Future` value of `nil`.

Closes: https://github.com/ruby-concurrency/concurrent-ruby/issues/963


eregon (approved):
Thank you!

Issue #962, opened 2022-10-17
Ruby locking is per-fiber, not per-thread.


ioquatix:
Ruby mutex implementation is per-fiber and generally speaking, all locking mechanisms should be per-fiber for the purpose of mutual exclusion.

As such, it appears at least some parts of concurrent ruby might not be correct, according to this specification.

I've been looking at <https://github.com/ruby-concurrency/concurrent-ruby/blob/385156117c110808875e1e87d5267147f696f796/lib/concurrent-ruby/concurrent/atomic/reentrant_read_write_lock.rb#L51> because we had some reports here <https://github.com/socketry/falcon/issues/166>

Specifically

```
      "kind":"Concurrent::IllegalOperationError",
      "message":"Cannot release a read lock which is not held",
```

It appears that multi-fiber re-entrancy is not supported or handled correctly, or used incorrectly.

I'm not sure what the correct solution is here, but at least we can start having a discussion about how it should work and what the solution is.

chrisseaton:
This is since Ruby 3.0 isn't it?

Not massively looking to working through the significance of this change...

Would you be willing to start by working with me to redefine what you think the specs should be now?

ioquatix:
IIRC, different versions of Ruby did different things and 3.0 standardised the behaviour. On some platforms where fibers were backed by threads, mutex was already per-fiber.

I checked the implementation a little bit and it looks okay to me, you are using `Mutex` correctly.

Maybe it's a genuine usage error on the part of rails, let me check their code.

ioquatix:
I've updated the original discussion of this issue, because I'm no longer sure that the issue is with this code.

ioquatix:
This seems to reproduce the original issue, at least the error is the same:

```ruby
require 'concurrent'

lock = Concurrent::ReentrantReadWriteLock.new
lock.acquire_read_lock

puts "Lockity lock!"

Thread.new do
  lock.release_read_lock
end.join
```

The question is, is this what's happening?

chrisseaton:
> Cannot release a read lock which is not held

That thread doesn't hold the lock so can't release it.

ioquatix:
Yes, I understand that, but why is it not happening on the same thread?

chrisseaton:
Are you saying you're aware that the last code snippet you posted is broken - you're saying that something else in some other code is making the same mistake as this example code, and that's what you're trying to fix?

Ok I think we're on the same page now. Yeah sounds like an application bug.

[...]

PR #983, opened 2023-01-11
Fix `ReentrantReadWriteLock` implementation when Mutex is per-fiber.

ioquatix:
Ruby 3.0+ is per-fiber Mutex. Currently, it's possible to get errors when using `ReentrantReadWriteLock` on different fibers. Let's fix that.

Fixes https://github.com/ruby-concurrency/concurrent-ruby/issues/962


ioquatix:
This PR also enables the tests under JRuby.

eregon:
> Use Thread.current.local_variable... for implementing thread locals. 

Unfortunately I think we cannot just do that, and this is the reason that `Concurrent::ThreadLocalVar` exists (AFAIK):
* When the `Concurrent::ThreadLocalVar` is no longer reachable, the values held on each thread are released, and so the values corresponding to that ThreadLocalVar can be GC'd and are not leaked, unlike when using `Thread#thread_variable_set` (same semantics as [Java ThreadLocal](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html)). https://github.com/ruby-concurrency/concurrent-ruby/blob/eac0237f54b18a9bdee3fb24ae25ee6b2136acfb/lib/concurrent-ruby/concurrent/atomic/ruby_thread_local_var.rb#L10 explains some of that.
* Those instances are anonymous, the key is the instance, and they cannot be accessed in any way but through the ThreadLocalVar instance, which means there isn't the problem of a thread local accessed by another which `Thread#thread_variable_get` allows. That's easy to workaround though.

Maybe the first point can be addressed with some `ObjectSpace.define_finalizer` on the ThreadLocalVar. But we would also need a weak set of Thread instances or something. That could be an array/set of Threads + a finalizer per Thread, a bit similar to what `RubyThreadLocalVar` does.

Some pointers I could find about `Concurrent::ThreadLocalVar`:
* https://github.com/ruby-concurrency/concurrent-ruby/commit/8dd7e175d04b87dd7534054db1d229274c0a2f31 https://github.com/ruby-concurrency/concurrent-ruby/pull/22 : this actually used `thread_variable_get`
* https://github.com/ruby-concurrency/concurrent-ruby/pull/164 is about fixing the leaks
* https://github.com/ruby-concurrency/concurrent-ruby/pulls?q=is%3Apr+ThreadLocalVar

eregon:
The rest of the PR looks good from a quick look.

To give an example issue with the current PR: each `ReentrantReadWriteLock` instance leaks a number of bytes per thread (i.e. a Ruby core thread-local/fiber-local variable, so at least the key which is a Symbol, the value and the internal hash entry), even after the ReentrantReadWriteLock is GC'd. That leak only goes away when the Thread instance is GC'd, which might never happen (with thread pools, etc).

ioquatix:
That makes sense. I don’t think there is any way in Ruby to delete a thread local or fiber local variable
In general. Maybe we can nullify it using a finalizer.

eregon:
> That makes sense. I don’t think there is any way in Ruby to delete a thread local or fiber local variable
> In general. Maybe we can nullify it using a finalizer.

Indeed, there is no such thing as `thread_variable_remove` or for fiber locals in Ruby (currently at least).
So we cannot even use one Ruby thread-local (i.e., `thread_variable_get`) per ThreadLocalVar instance, because that would leak the internal Hash entry on the Thread and the key Symbol (and per Thread which used that ThreadLocalVar), even if we would set the value to `nil` via a finalizer.
That's probably the main reason why the current implementation uses a single Ruby thread-local for all ThreadLocalVar instances, to avoid that leak.

Aside: I wonder if this is partly why I didn't mind the `:"foo_#{object_id}"` workaround for Fiber storage variables, because anyway those also don't support removal and keep a strong reference to their key + value until the Fiber is gone, regardless of whether the key would become unreachable (at least currently, maybe something that could be improved in Ruby, but unlikely because incompatible and would likely have a big perf trade-off).
Hence if being able to GC a thread/fiber-local is wanted, then one needs concurrent-ruby's ThreadLocalVar/FiberLocalVar.

ioquatix:
One way to deal with this would be to assume assigning nil deletes the value - i.e. nil is indistinguishable from unset.

ioquatix:
I've reverted back to the shared array implementation, and reworked it into `class AbstractLocals`, `class ThreadLocals` and `class FiberLocals`. We could perhaps make the name a little more specific, e.g. `AbstractLocalsArray` or something but I honestly don't have a strong opinion about it.

Generally, it seems okay. I'm a little concerned about the O(number of ECs) cost of deleting a local. If we are expecting people to have 100s or 100,000s of threads/fibers etc, this is a non-trivial cost.

A better solution might be to use a linked list for the `@all_locals`. Each EC would use a linked list node which is removed when it goes out of scope. With a little cunning, it might be possible to retain the similar synchronisation free approach.

lib/concurrent-ruby/concurrent/atomic/thread_local_var.rb
```diff
@@ -1,104 +1,61 @@
...
-  #   @!method value
-  #     @!macro thread_local_var_method_get
-  #
-  #   @!method value=(value)
-  #     @!macro thread_local_var_method_set
-  #
-  #   @!method bind(value, &block)
-  #     @!macro thread_local_var_method_bind
-
-  ###################################################################
```
eregon, line 56: The documentation above should be kept, and also added for FiberLocalVar

lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb
```diff
@@ -0,0 +1,20 @@
...
+      # Check if the mutex is still owned in a child fiber:
+      Fiber.new{mutex.owned?}.resume
+    end
+  end
+
+  if mutex_owned_per_thread?
+    LockLocalVar = ThreadLocalVar
+  else
+    LockLocalVar = FiberLocalVar
+  end
```
eregon, line 19: I'm unsure about this name, especially a public constant like this under `Concurrent`.
I think for now it might be better to just use that logic in ReentrantReadWriteLock.

lib/concurrent-ruby/concurrent/atomic/locals.rb
```diff
@@ -0,0 +1,184 @@
...
+      weak_synchronize do
+        # This is O(N) number of ECs with locals... I'm concerned about the
+        # performance cost. But I don't see a better way to do it. Using
+        # `each_value` can blow up if @all_locals is modified during iteration.
+        # Maybe using a linked list for `@all_locals` would be better?
+        @all_locals.values.each do |locals|
+          locals[index] = nil
+        end
+
+        @free << index
```
eregon, line 81: We should keep the original implementation's comments, notably here there was:
> `# free index has to be published after the arrays are cleared`

lib/concurrent-ruby/concurrent/atomic/locals.rb
```diff
@@ -0,0 +1,184 @@
...
+
+      return index
+    end
+
+    def free_index(index)
+      weak_synchronize do
+        # This is O(N) number of ECs with locals... I'm concerned about the
+        # performance cost. But I don't see a better way to do it. Using
+        # `each_value` can blow up if @all_locals is modified during iteration.
+        # Maybe using a linked list for `@all_locals` would be better?
```
eregon, line 76: The previous comment here had more information:
```
          # The cost of GC'ing a TLV is linear in the number of threads using TLVs
          # But that is natural! More threads means more storage is used per TLV
          # So naturally more CPU time is required to free more storage
          #
          # DO NOT use each_value which might conflict with new pair assignment
          # into the hash in #value= method
```
Could you add it back on top or below your comment?

lib/concurrent-ruby/concurrent/atomic/locals.rb
```diff
@@ -0,0 +1,184 @@
...
+  # get bigger and bigger with time), and 2) we need to null out all the
+  # references held in the now-unused slots (both to avoid blocking GC of those
+  # objects, and also to prevent "stale" values from being passed on to a new
+  # local when the index is reused).
+  #
+  # Because we need to null out freed slots, we need to keep references to
+  # ALL the locals arrays, so we can null out the appropriate slots in all of
+  # them. This is why we need to use a finalizer to clean up the locals array
+  # when the EC goes out of scope.
+  class AbstractLocals
```
eregon, line 33: There are quite a few changes here compared to `RubyThreadLocalVar`, I'll need to review those differences/the new logic carefully.

lib/concurrent-ruby/concurrent/atomic/thread_local_var.rb
ioquatix: Done.

lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb
ioquatix: Feel free to suggest a better name?

I am happy to move this logic but it seems like external consumers might run into the same problem and thus exposing it seems pretty reasonable to me. i.e. we use it in the tests too.

lib/concurrent-ruby/concurrent/atomic/locals.rb
ioquatix: It (should be) semantically identical, just without global variables.

lib/concurrent-ruby/concurrent/atomic/locals.rb
ioquatix: Done.

lib/concurrent-ruby/concurrent/atomic/locals.rb
ioquatix: I'll remove my comment as they basically say the same thing. However, I fundamentally dislike:

> So naturally more CPU time is required to free more storage.

Essentially, freeing the storage is O(number of threads/fibers). On a system with lots of instances, that will be slow. However, it should be an infrequent operation. I personally think we should be aiming for constant time if possible.

eregon: I don't think constant time is possible here, which is the case of a ThreadLocalVar/FiberLocalVar being GC'd. We need to clean (up to) N memory locations, one for each Thread/Fiber which used it (and since this is globalized, one for each Thread/Fiber which ever used a ThreadLocalVar/FiberLocalVar).

ioquatix: Using `@all_locals.values` is also a potentially expensive memory allocation, and O(2N) because it will iterate the list twice (once for the initial copy into the temporary array, and once for the actual iteration). Well, I'm not trying to solve that problem in this PR, but I don't think it's good for scalability.

ioquatix:
@eregon I've done another pass over this, squashed all the commits, etc. Let me know if you want any further changes made.

lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb
```diff
@@ -0,0 +1,20 @@
...
+module Concurrent
+  def self.mutex_owned_per_thread?
+    mutex = Mutex.new
+
+    # Lock the mutex:
+    mutex.synchronize do
+      # Check if the mutex is still owned in a child fiber:
+      Fiber.new{mutex.owned?}.resume
+    end
+  end
```
eregon, line 13: I'm thinking we should cache that and only do it once, and just return true on truffleruby/jruby because that's been always true for them (and creating a Fiber there is expensive)

eregon: Or just do that logic at the top-level and declare a constant instead/in addition of the method

ioquatix: What about making the method private because it’s only used once and probably shouldn’t be a public interface anyway?

ioquatix: Okay, I marked it as private.

ioquatix: Also, it's already only used once, and the result is already cached (which constant is defined) so I'm not sure what you want to change?

ioquatix: Maybe we can merge this PR and then follow up with a JRuby/TruffleRuby specific changes as makes sense since I don't really know what the code should look like (nor do I think we should hard code anything if it might change in the future).

eregon:
@ioquatix Could you rebase? (feel free to squash your changes to make the rebase easier)
Then I'll review.

ioquatix:
Okay, rebase done.

lib/concurrent-ruby/concurrent/atomic/java_thread_local_var.rb
```diff
@@ -1,37 +0,0 @@
-require 'concurrent/atomic/abstract_thread_local_var'
-
-if Concurrent.on_jruby?
-
-  module Concurrent
-
-    # @!visibility private
-    # @!macro internal_implementation_note
-    class JavaThreadLocalVar < AbstractThreadLocalVar
```
eregon, line 9: There were various discussions in the past whether it's better to keep this or not.
I think given that's actually per-Fiber on JRuby it's just so confusing, so agreed let's simplify and use the Ruby ones for all.

eregon (commented):
I don't see any specs for FiberLocalVar, this is necessary as we introduce a new public class.

lib/concurrent-ruby/concurrent/atomic/thread_local_var.rb
```diff
@@ -1,104 +1,90 @@
...
-  #       v.value #=> 14
-  #       v.value = 2
-  #       v.value #=> 2
-  #     end
-  #
-  #     v.value #=> 14
-  #
-  #   @see https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html Java ThreadLocal
-  #
-  # @!macro thread_local_var_public_api
```
eregon, line 101: This class doc seems removed, I'll restore it.

lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb
eregon: It seems reasonable to expose it and low enough risk.
Could be MutexLocalVar but it also applies to Monitor, so LockLocalVar sounds fine.

eregon (approved):
I added specs for FiberLocalVar.
Should be good to go now, just waiting CI.
Thank you for the PR!

lib/concurrent-ruby/concurrent/atomic/locals.rb
```diff
@@ -146,23 +139,23 @@ def locals!
...
   class ThreadLocals < AbstractLocals
     def locals
-      Thread.current.thread_variable_get(@name)
+      Thread.current.thread_variable_get(:concurrent_thread_locals)
     end
 
     def locals!
       thread = Thread.current
-      locals = thread.thread_variable_get(@name)
+      locals = thread.thread_variable_get(:concurrent_thread_locals)
```
ioquatix, line 147: This change means that multiple instances of the class will clobber each other.

eregon: Yes but since we only need 1 instance we support just that and not more

ioquatix: Okay, fair enough.

PR #990, opened 2023-02-20
Add `require 'fiber'` for `Fiber.current` usage.

ioquatix:
Fixes <https://github.com/ruby-concurrency/concurrent-ruby/issues/962>.


matthewd (approved):
At a quick glance it looks like the prevailing style is to require stdlib before `concurrent/*`, FWIW... though it obviously doesn't matter.

ioquatix:
I think we should also consider moving `Fiber.current` into core ruby.

ioquatix:
https://bugs.ruby-lang.org/issues/19453

eregon (approved):
Thanks!

PR #991, opened 2023-02-21
Prefer `Process.pid` over `$$`

casperisfine:
It's way more readable, but also it allows `Process.pid` to be cached by things like https://github.com/Shopify/pid_cache

Hopefully in Ruby 3.3 it would matter anymore though (https://bugs.ruby-lang.org/issues/19443).

cc @eregon is this something you'd be OK with?

Alternatively I can implement a PID cache for Ruby 3.1+ like I did in `dalli` https://github.com/petergoldstein/dalli/pull/953 and `redis-client` https://github.com/redis-rb/redis-client/pull/91

### DIFF
--- a/lib/concurrent-ruby/concurrent/async.rb
+++ b/lib/concurrent-ruby/concurrent/async.rb
@@ -290,7 +290,7 @@ module Concurrent
         @delegate = delegate
         @queue = []
         @executor = Concurrent.global_io_executor
-        @ruby_pid = $$
+        @ruby_pid = Process.pid
       end
 
       # Delegates method calls to the wrapped object.
@@ -346,9 +346,9 @@ module Concurrent
       end
 
       def reset_if_forked
-        if $$ != @ruby_pid
+        if Process.pid != @ruby_pid
           @queue.clear
-          @ruby_pid = $$
+          @ruby_pid = Process.pid
         end
       end
     end
--- a/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
@@ -138,7 +138,7 @@ module Concurrent
       @completed_task_count = 0
       @largest_length       = 0
       @workers_counter      = 0
-      @ruby_pid             = $$ # detects if Ruby has forked
+      @ruby_pid             = Process.pid # detects if Ruby has forked
 
       @gc_interval  = opts.fetch(:gc_interval, @idletime / 2.0).to_i # undocumented
       @next_gc_time = Concurrent.monotonic_time + @gc_interval
@@ -287,7 +287,7 @@ module Concurrent
     end
 
     def ns_reset_if_forked
-      if $$ != @ruby_pid
+      if Process.pid != @ruby_pid
         @queue.clear
         @ready.clear
         @pool.clear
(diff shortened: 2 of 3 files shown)


eregon (approved):
This looks good, thanks

eregon:
```
$ ruby -rbenchmark/ips -e 'Benchmark.ips { |x| x.report("Process.pid") { Process.pid }; x.report("$$") { $$ }; x.compare! }'
Warming up --------------------------------------
         Process.pid   966.730k i/100ms
                  $$     1.071M i/100ms
         Process.pid      9.552M (± 2.4%) i/s -     48.336M in   5.062964s
                  $$     10.616M (± 2.2%) i/s -     53.559M in   5.047529s

Comparison:
                  $$: 10615866.3 i/s
         Process.pid:  9552479.6 i/s - 1.11x  (± 0.00) slower
```
On `ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]`.

So it's not ideal from a perf POV on CRuby.

eregon:
On (existing releases of) TruffleRuby `Process.pid` is much slower:
```
$ ruby -rbenchmark/ips -e 'Benchmark.ips { |x| x.report("Process.pid") { Process.pid }; x.report("$$") { $$ }; x.compare! }'
Warming up --------------------------------------
         Process.pid   112.386k i/100ms
                  $$   357.722M i/100ms
         Process.pid      4.953M (±26.4%) i/s -     22.477M in   5.018213s
                  $$      3.539B (± 2.0%) i/s -     17.886B in   5.055332s

Comparison:
                  $$: 3539461429.1 i/s
         Process.pid:  4952808.9 i/s - 714.64x  slower
```
because the current implementation is:
```ruby
$$ = Process.pid if Truffle::POSIX::NATIVE

module Process
  def pid
    Truffle::POSIX.getpid
  end
end
```

So I think we should revert this and consider this something that should be fixed in Ruby implementations, because it slows down existing Ruby releases.

https://bugs.ruby-lang.org/issues/19443 and https://github.com/oracle/truffleruby/issues/2882

casperisfine:
Interesting. I assume being a magic variable it's faster to call.

I'm fine with reverting, it's not a huge deal. Or alternatively I can implement a manual PIDCache like I did in Dalli / redis-client / etc.

eregon:
I think the cache in the Ruby implementations is the simplest and best solution here.
And IMHO it would make sense to backport that to e.g. CRuby 3.2 or so.

Maybe we could use a `_fork` hook to avoid those pids checks entirely on the fast path, but that's probably quite messy while supporting Ruby < 3.1.

casperisfine:
> Maybe we could use a `_fork` hook to avoid those pids checks entirely on the fast path, but that's probably quite messy while supporting Ruby < 3.1.

It's not that bad:
  - https://github.com/petergoldstein/dalli/pull/953
  - https://github.com/redis-rb/redis-client/pull/91

You could do it on pre 3.1, but that is messy:
  - https://github.com/rails/rails/blob/7036bc2b633463e8ce1011f32ec5eb87ce435d06/activesupport/lib/active_support/fork_tracker.rb

Anyway, happy to throw a PR if you want it.

eregon:
Do you know how `$$` is implemented on CRuby? Maybe it already caches the pid? I guess not, but just wondering.
I guess the difference is just in avoiding an extra Ruby call on CRuby.

eregon:
I'll think a bit more about it, but if we can solve this on the Ruby impls side it seems a lot cleaner.

casperisfine:
> Do you know how $$ is implemented on CRuby? Maybe it already caches the pid? 

It's the same implementation than `Process.pid`, just saves a method call semantic. See https://github.com/ruby/ruby/pull/7326

PR #992, opened 2023-02-22
Revert "Prefer `Process.pid` over `$$`"

eregon:
Reverts ruby-concurrency/concurrent-ruby#991
Since it's significantly slower on existing Ruby releases.
See comments there for details.


casperisfine:
👍

PR #996, opened 2023-05-24
Fix TimerTask docs to not refer to `#execute` as "blocking"

### LINKED ISSUE #794: TimerTask#execute is not a blocking call
Hi,

I was looking through the TimerTask examples and noticed that the one below isn't working as documented:

```ruby
# timer_task_boom_test.rb

# from https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/timer_task.rb#L94-L111
timer_task = Concurrent::TimerTask.new(execution_interval: 1) do |task|
  task.execution_interval.to_i.times { print 'Boom! ' }
  print "\n"
  task.execution_interval += 1
  if task.execution_interval > 5
    puts 'Stopping...'
    task.shutdown
  end
end

timer_task.execute # blocking call - this task will stop itself
#=> Boom!
#=> Boom! Boom!
#=> Boom! Boom! Boom!
#=> Boom! Boom! Boom! Boom!
#=> Boom! Boom! Boom! Boom! Boom!
#=> Stopping...
```

If I put those contents into a file and run it via `ruby timer_task_boom_test.rb`, it exits immediately. If I add a `sleep` right after the timer_task.execute line then I see the `Boom!`s printed out.

Is timer_task.execute supposed to be a blocking call? Or has the behavior changed and this example is no longer valid?

## more info
```
* Ruby implementation:             Ruby 2.2.3
* `concurrent-ruby` version:       1.1.4
```

bensheldon:
Also fixes some small other oddities I found when running the code example.

Closes #794

### DIFF
--- a/lib/concurrent-ruby/concurrent/timer_task.rb
+++ b/lib/concurrent-ruby/concurrent/timer_task.rb
@@ -87,7 +87,7 @@ module Concurrent
   #
   # @example Controlling execution from within the block
   #   timer_task = Concurrent::TimerTask.new(execution_interval: 1) do |task|
-  #     task.execution_interval.times{ print 'Boom! ' }
+  #     task.execution_interval.to_i.times{ print 'Boom! ' }
   #     print "\n"
   #     task.execution_interval += 1
   #     if task.execution_interval > 5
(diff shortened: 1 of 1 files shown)


ioquatix (approved):

lib/concurrent-ruby/concurrent/timer_task.rb
```diff
@@ -96,12 +96,11 @@ module Concurrent
   #     end
   #   end
   #
-  #   timer_task.execute # blocking call - this task will stop itself
+  #   timer_task.execute
   #   #=> Boom!
   #   #=> Boom! Boom!
   #   #=> Boom! Boom! Boom!
   #   #=> Boom! Boom! Boom! Boom!
-  #   #=> Boom! Boom! Boom! Boom! Boom!
```
ioquatix, line 104: Can you double check this removal is correct?

When I ran the code, I got:

```
Boom! 
Boom! Boom! 
Boom! Boom! Boom! 
Boom! Boom! Boom! Boom! 
Boom! Boom! Boom! Boom! Boom! 
Stopping...
```

ioquatix:
Thanks!

PR #997, opened 2023-05-25
Add `TimerTask#interval_type` option to configure interval calculation

### LINKED ISSUE #678: TimerTask should schedule next run before doing job
As far as I understand the logic behind the TimerTask, it should run a job every *execution* seconds and kill it if it running too long.
The current logic is a little bit different. It causes a job to be run *execution* seconds *after* it did the previous iteration.
Short example. If I need to poll something every 30 seconds. I'll create a timer task with execution_interval set to 30 and say timeout set to 15 seconds. My code is polling a slow thing which can reply in 20 seconds or more. 
What I expect:
```
0:00 Poll
0:10 Poll finished
0:30 Poll
0:45 Poll finished
1:00 Poll
1:05 Poll finished
1:30 Poll
1:35 Poll finished
...
```
What really happens with TimerTask:
```
0:00 Poll
0:10 Poll finished
0:40 Poll
0:55 Poll finished
1:25 Poll
1:30 Poll finished
2:00 Poll
2:05 Poll finished
...
```
Now TimerTask doesn't look like a good candidate for polling.

```ruby
#!/usr/bin/env ruby
require "concurrent"

puts "Creating tasks"
10.times do |x|
  task = Concurrent::TimerTask.new(run_now: true, execution_interval: 20) do
    sleep x*2
    puts "#{Time.now} #{x*2} finished"
  end
  task.execute
end

sleep 60
puts "Terminated"
```
The output:
```
Creating tasks
2017-10-19 01:41:48 +0300 0 finished
2017-10-19 01:41:50 +0300 2 finished
2017-10-19 01:41:52 +0300 4 finished
2017-10-19 01:41:54 +0300 6 finished
2017-10-19 01:41:56 +0300 8 finished
2017-10-19 01:41:58 +0300 10 finished
2017-10-19 01:42:00 +0300 12 finished
2017-10-19 01:42:02 +0300 14 finished
2017-10-19 01:42:04 +0300 16 finished
2017-10-19 01:42:06 +0300 18 finished
2017-10-19 01:42:08 +0300 0 finished
2017-10-19 01:42:12 +0300 2 finished
2017-10-19 01:42:16 +0300 4 finished
2017-10-19 01:42:20 +0300 6 finished
2017-10-19 01:42:24 +0300 8 finished
2017-10-19 01:42:28 +0300 0 finished
2017-10-19 01:42:28 +0300 10 finished
2017-10-19 01:42:32 +0300 12 finished
2017-10-19 01:42:34 +0300 2 finished
2017-10-19 01:42:36 +0300 14 finished
2017-10-19 01:42:40 +0300 4 finished
2017-10-19 01:42:40 +0300 16 finished
2017-10-19 01:42:44 +0300 18 finished
2017-10-19 01:42:46 +0300 6 finished
2017-10-19 01:42:48 +0300 0 finished
Terminated
```
Consider the increasing intervals.

******
-   Operating system: linux
-   `concurrent-ruby-ext` installed: no

bensheldon:
Can be either `:fixed_delay` or `:fixed_rate`, default to `:fixed_delay`.

Closes #678

Notes:

- I would love a better suggestion for the parameter than `interval_type`
- The option names come from @eregon's [suggestion](https://github.com/ruby-concurrency/concurrent-ruby/issues/678#issuecomment-1559751900) to mirror Java's options

### DIFF
--- a/lib/concurrent-ruby/concurrent/timer_task.rb
+++ b/lib/concurrent-ruby/concurrent/timer_task.rb
@@ -32,6 +32,17 @@ module Concurrent
   # be tested separately then passed to the `TimerTask` for scheduling and
   # running.
   #
+  # A `TimerTask` supports two different types of interval calculations.
+  # A fixed delay will always wait the same amount of time between the
+  # completion of one task and the start of the next. A fixed rate will
+  # attempt to maintain a constant rate of execution regardless of the
+  # duration of the task. For example, if a fixed rate task is scheduled
+  # to run every 60 seconds but the task itself takes 10 seconds to
+  # complete, the next task will be scheduled to run 50 seconds after
+  # the start of the previous task. If the task takes 70 seconds to
+  # complete, the next task will be start immediately after the previous
+  # task completes. Tasks will not be executed concurrently.
+  #
   # In some cases it may be necessary for a `TimerTask` to affect its own
   # execution cycle. To facilitate this, a reference to the TimerTask instance
   # is passed as an argument to the provided block every time the task is
@@ -73,6 +84,12 @@ module Concurrent
   #   task.execute
   #
   #   #=> 'Boom!'
+  #
+  # @example Configuring `:interval_type` with either :fixed_delay or :fixed_rate, default is :fixed_delay
+  #   task = Concurrent::TimerTask.new(execution_interval: 5, interval_type: :fixed_rate) do
+  #          puts 'Boom!'
+  #        end
+  #   task.interval_type #=> :fixed_rate
   #
   # @example Last `#value` and `Dereferenceable` mixin
   #   task = Concurrent::TimerTask.new(
@@ -152,8 +169,16 @@ module Concurrent
     # Default `:execution_interval` in seconds.
     EXECUTION_INTERVAL = 60
 
-    # Default `:timeout_interval` in seconds.
-    TIMEOUT_INTERVAL = 30
+    # Maintain the interval between the end of one execution and the start of the next execution.
+    FIXED_DELAY = :fixed_delay
+
+    # Maintain the interval between the start of one execution and the start of the next.
+    # If execution time exceeds the interval, the next execution will start immediately
+    # after the previous execution finishes. Executions will not run concurrently.
+    FIXED_RATE = :fixed_rate
+
+    # Default `:interval_type`
+    DEFAULT_INTERVAL_TYPE = FIXED_DELAY
 
     # Create a new TimerTask with the given task and configuration.
     #
@@ -164,6 +189,9 @@ module Concurrent
     #   @option opts [Boolean] :run_now Whether to run the task immediately
     #     upon instantiation or to wait until the first #  execution_interval
     #     has passed (default: false)
+    #   @options opts [Symbol] :interval_type method to calculate the interval
+    #     between executions, can be either :fixed_rate or :fixed_delay.
+    #     (default: :fixed_delay)
     #   @option opts [Executor] executor, default is `global_io_executor`
     #
     #   @!macro deref_options
@@ -243,6 +271,10 @@ module Concurrent
       end
     end
 
+    # @!attribute [r] interval_type
+    # @return [Symbol] method to calculate the interval between executions
+    attr_reader :interval_type
+
     # @!attribute [rw] timeout_interval
     # @return [Fixnum] Number of seconds the task can run before it is
     #   considered to have failed.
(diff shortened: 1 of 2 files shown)


lib/concurrent-ruby/concurrent/timer_task.rb
```diff
@@ -242,6 +267,24 @@ def execution_interval=(value)
...
+    # @return [Symbol] method to calculate the interval between executions, can be either
+    #   :fixed_rate or :fixed_delay; default to :fixed_delay.
+    def interval_type
+      synchronize { @interval_type }
+    end
+
+    # @!attribute [rw] interval_type
+    # @return [Symbol] method to calculate the interval between executions, can be either
+    #   :fixed_rate or :fixed_delay; default to :fixed_delay.
+    def interval_type=(value)
```
ioquatix, line 280: Consider removing this as discussed.

lib/concurrent-ruby/concurrent/timer_task.rb
```diff
@@ -152,8 +169,16 @@ class TimerTask < RubyExecutorService
...
+    # Maintain the interval between the end of one execution and the start of the next execution.
+    FIXED_DELAY = :fixed_delay
+
+    # Maintain the interval between the start of one execution and the start of the next.
+    # If execution time exceeds the interval, the next execution will start immediately
+    # after the previous execution finishes. Executions will not run concurrently.
+    FIXED_RATE = :fixed_rate
+
+    # Default `:interval_type`
+    INTERVAL_TYPE = FIXED_DELAY
```
ioquatix, line 181:
```suggestion
    DEFAULT_INTERVAL_TYPE = FIXED_DELAY
```

lib/concurrent-ruby/concurrent/timer_task.rb
```diff
@@ -152,8 +169,16 @@ class TimerTask < RubyExecutorService
     # Default `:execution_interval` in seconds.
     EXECUTION_INTERVAL = 60
 
-    # Default `:timeout_interval` in seconds.
-    TIMEOUT_INTERVAL = 30
+    # Maintain the interval between the end of one execution and the start of the next execution.
+    FIXED_DELAY = :fixed_delay
```
ioquatix, line 173: As discussed, we see a third mode, which could be `fixed_rate_drop` which drops subsequent executions if the current execution violates the maximum time slice given the rate specified.

lib/concurrent-ruby/concurrent/timer_task.rb
```diff
@@ -74,6 +85,12 @@ module Concurrent
   #
   #   #=> 'Boom!'
   #
+  # @example Configuring `:interval_type` with either :fixed_delay or :fixed_rate, default is :fixed_delay
+  #   task = Concurrent::TimerTask.new(execution_interval: 5, interval_type: :fixed_rate) do
```
ioquatix, line 89: I personally consider the `interval_type` to be a static configuration of the behaviour of the application rather than something decided at run-time. Therefore, I wonder if this would be better implemented using sub-classes which implement the specific behaviour required, e.g. `FixedRateTimerTask`.

ioquatix (changes requested):
I agree with the proposed idea but I have some suggestions about the implementation.

lib/concurrent-ruby/concurrent/timer_task.rb
eregon: Yeah that makes sense to me.

lib/concurrent-ruby/concurrent/timer_task.rb
eregon: Subclasses seem rather heavy for this, and I am concerned that for compatibility we would need to keep Concurrent::TimerTask so it would be both a leaf class and an abstract class, that sounds very confusing.
So I think a constructor argument like here makes most sense and is consistent with what is already done in concurrent-ruby APIs.

lib/concurrent-ruby/concurrent/timer_task.rb
eregon: Agreed, it should be only set when building a new TimerTask and not changed after

lib/concurrent-ruby/concurrent/timer_task.rb
```diff
@@ -296,10 +340,12 @@ def schedule_next_task(interval = execution_interval)
     # @!visibility private
     def execute_task(completion)
       return nil unless @running.true?
+      start = Concurrent.monotonic_time
       _success, value, reason = @executor.execute(self)
       if completion.try?
         self.value = value
-        schedule_next_task
+        interval = interval_type == FIXED_DELAY ? execution_interval : [execution_interval - (Concurrent.monotonic_time - start), 0].max
```
eregon, line 347: This is quite a long line, how about a `if/else` or a `case/when`?

lib/concurrent-ruby/concurrent/timer_task.rb
ioquatix: Why would it be an abstract class?

lib/concurrent-ruby/concurrent/timer_task.rb
ioquatix: I think it would be simpler if it was a sub-class. This PR would only be a 2-3 line change in that case.

ioquatix:
@bensheldon if you have time, I suggest we make a 2nd PR which uses a derived class, so we can compare the level of complexity.

bensheldon:
@ioquatix no prob! I'll try to get to that next week.

lib/concurrent-ruby/concurrent/timer_task.rb
eregon: > Why would it be an abstract class?

If you make subclasses the superclass should be abstract. Otherwise I think it's pretty confusing.

eregon (approved):

eregon:
This looks great, merging, thank you!

PR #998, opened 2023-05-25
Have ScheduledTask raise ArgumentError if provided executor has fallback policy of `:abort`

### LINKED ISSUE #889: ScheduledTask triggering ThreadPoolExecutor's `fallback_policy: :abort` breaks ScheduledTasks
When `ScheduledTask`s are being run on a ThreadPoolExecutor with `fallback_policy: :abort`, triggering the abort breaks all subsequent usage of ScheduledTasks, they never execute, even when running on a different executor.

Apologies that I don't quite understand the internals of ScheduledTasks, but I'm hypothesizing that when their delay expires, they attempt to execute, and then the executor raises the abort exception, and the ScheduledTask internal is not recovering from the exception.

```ruby
executor_options = {
  name: 'test',
  min_threads: 0,
  max_threads: 2,
  auto_terminate: true,
  idletime: 60,
  max_queue: -1,
  fallback_policy: :abort, # <= this configuration is the problem; :discard works fine
}.freeze

executor = Concurrent::ThreadPoolExecutor.new(executor_options)

(1..4).to_a.each do |i|
  # Schedule more tasks to execute at the same time than there are available threads
  4.times do
    task = Concurrent::ScheduledTask.new(i, executor: executor) { puts Time.now }
    task.add_observer(proc { |time, output, thread_error| puts "Finished: #{time}, #{output}, #{thread_error}" }, :call)
    task.execute
  end
end

# No scheduled tasks work from now on, not even ones of the global pool
task = Concurrent::ScheduledTask.execute(1) { puts Time.now } # <= never executes
task.value # <= never completes
```

```
* Ruby implementation:             MRI Ruby
* `concurrent-ruby` version:       v1.1.7
```

bensheldon:
Closes #889

### DIFF
--- a/lib/concurrent-ruby/concurrent/scheduled_task.rb
+++ b/lib/concurrent-ruby/concurrent/scheduled_task.rb
@@ -189,6 +189,8 @@ module Concurrent
         @task = task
         @time = nil
         @executor = Options.executor_from_options(opts) || Concurrent.global_io_executor
+        raise ArgumentError.new("cannot use an executor with fallback_policy: :abort") if @executor.fallback_policy == :abort
+
         self.observers = Collection::CopyOnNotifyObserverSet.new
       end
     end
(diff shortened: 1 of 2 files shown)


bensheldon:
Closing because I think I want to approach this differently given how many places seem to assume that fallback policy `:abort` is fine.

PR #1000, opened 2023-05-28
Allow TimerTask to be initialized with a specified Executor

### LINKED ISSUE #691: Shutting down TimerTask is stopping already executing tasks
Shutting down a timer task is not waiting for already executing tasks to finish. I talked to @headius and he encouraged me to open an issue since it is not clear if this is a feature or bug.

Example code:
~~~ruby
require "rubygems"
require "bundler/setup"
require "concurrent"

tasks = 10.times.map do |s|
  Concurrent::TimerTask.new(execution_interval: 1) do
    sleep s
    puts "s: #{s}\n"
  end
end

begin
  tasks.each { |t| t.execute }
  sleep
rescue Interrupt
  puts "Received INT"
  puts "Shutting down tasks"
  tasks.each { |t| t.shutdown }
ensure
  puts "Releasing resources"
end
~~~

bensheldon:
The resulting ScheduledTasks will be initialized with that executor.

Closes #691

### DIFF
--- a/lib/concurrent-ruby/concurrent/timer_task.rb
+++ b/lib/concurrent-ruby/concurrent/timer_task.rb
@@ -164,6 +164,7 @@ module Concurrent
     #   @option opts [Boolean] :run_now Whether to run the task immediately
     #     upon instantiation or to wait until the first #  execution_interval
     #     has passed (default: false)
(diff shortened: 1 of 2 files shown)


eregon (approved):
Looks good, sorry for the very late review.

PR #1001, opened 2023-05-29
Allow TimerTask to be safely restarted after shutdown and avoid duplicate tasks

### LINKED ISSUE #672: shutdown and execute a timertask results in double runs
Script as PoC:

```ruby
puts "Start: #{Time.now}"
task = Concurrent::TimerTask.execute(execution_interval: 3) { puts "In task: #{Time.now}" }
sleep 4
task.shutdown
puts "Restart: #{Time.now}"
task.execute
sleep 6
task.shutdown
puts "End: #{Time.now}"
```

Example output:
```
Start: 2017-09-06 12:24:54 +0200
In task: 2017-09-06 12:24:57 +0200
Restart: 2017-09-06 12:24:58 +0200
In task: 2017-09-06 12:25:00 +0200
In task: 2017-09-06 12:25:01 +0200
In task: 2017-09-06 12:25:03 +0200
End: 2017-09-06 12:25:04 +0200
```

The tekst "In task: 2017-09-06 12:25:00 +0200" is not what I would expect here, this belongs to the first execution. The combination of shutdown and execute results in 1 extra execution of the first timer.
if you create a new TimerTask instead, the pending task is not executed.

******
-   Operating system: linux
-   `concurrent-ruby-ext` installed: yes

bensheldon:
Fixes #672

Creates a new @running object on shutdown while the ScheduledTask is checking the previously set @running object which will remain false


ioquatix:
I have several concerns about the implementation of `TimerTask` and the usage of `@running`. While this PR on the surface looks okay to me, @matthewd do you mind taking a closer look? I feel like we need to make sure the semantics of `execute` and `shutdown` are clear.

matthewd:
Yeah, I think I agree there are some semantic questions here.

Particularly given the explicitly documented "run-then-sleep-N loop" implementation (as opposed to "run every N"), I think it's reasonable to expect that only one instance of the block will be running at any given time. (Notwithstanding the internal use of the `completion` event, which almost feels like a nod toward at-least-once behaviour, but looks to me like it actually became redundant when the timeout behaviour [which involved two racing tasks] was removed.)

The linked issues are clearly violations of that principle, and this PR reduces the circumstances where that can happen... but an `execute; shutdown; execute` could still leave two copies overlapping if the first run starts just before the shutdown. 

Setting aside whether it's a complete fix, in terms of whether this PR is safe in isolation: I'm not certain of the semantics of an unsynchronized read of an instance variable that's potentially being written by another (synchronized) thread. Without going back to research it properly, the local implementation of `execution_interval` seems to support my vague feeling that it may be problematic.

On implementation, I note that ScheduledTask has a potentially-relevant `cancel` method. If we have to switch to an AtomicReference anyway, perhaps we can just remember the upcoming ScheduledTask and then cancel it directly when necessary?

bensheldon:
In consultation with @matthewd, I tried using an AtomicFixnum to store and check the "age" of the TimerTask (incremented when started/restaretd) to try to avoid any variable reassignment.

eregon:
Is this ready to be merged, if so @bensheldon could you merge it?
(I plan to make a release soon)

Issue #684, opened 2017-12-06
Thread pools behaving oddly with `max_queue` and `remaining_capacity`


remofritzsche:
We're in need for a thread pool that has a maximum number of parallel threads, no queue and fails if too many threads are added. We also need to be able to detect how many threads are "free" at a given time.

We've tried almost all of the ThreadPool and ThreadPoolExecutor classes but can't get it to work:

```
require 'concurrent'

pool = Concurrent::ThreadPoolExecutor.new(
  min_threads: 0,
  max_threads: 5,
  max_queue: 0,
  fallback_policy: :abort,
  auto_terminate: false
)

6.times do
  pool.post { sleep 5 }
  puts pool.remaining_capacity
end

pool.wait_for_termination
```

If we interpret the documentation right, this should create a thread pool with 0 threads at start that can grow up to 5 threads which are handled in parallel. If all 5 threads are occupied, posting new work units should result in an exception. There should be no additional queue. We'd also expect `remaining_capacity` to drop by 1 each time something has been posted and the previous threads are still busy.

The output of the above script:

```
-1
-1
-1
-1
-1
-1
```

Expected would be:

```
4
3
2
1
0
=> EXCEPTION
```

Did we interpret the documentation wrong and how can we achieve what we need? We're also need to be able to detect the number of idle threads so that we can post the right number of work units to it.

On a side-note: The above example stops with a number of exceptions like the following:

```
/usr/local/lib/ruby/gems/2.4.0/gems/concurrent-ruby-1.0.5/lib/concurrent/synchronization/mri_lockable_object.rb:43:in `sleep': No live threads left. Deadlock? (fatal)
6 threads, 6 sleeps current:0x00007fa6936e27d0 main thread:0x00007fa6937044c0
* #<Thread:0x00007fa69487f0b8 sleep_forever>
   rb_thread_t:0x00007fa6937044c0 native:0x00007fff7b53e000 int:1
   /usr/local/lib/ruby/gems/2.4.0/gems/concurrent-ruby-1.0.5/lib/concurrent/synchronization/mri_lockable_object.rb:43:in `sleep'
   /usr/local/lib/ruby/gems/2.4.0/gems/concurrent-ruby-1.0.5/lib/concurrent/synchronization/mri_lockable_object.rb:43:in `wait'
[...]

PR #1002, opened 2023-05-29
Create method ThreadPoolExecutor#active_count to expose the number of threads that are actively executing tasks

bensheldon:
Connects to #684.

~~I called it "available" because `ready` means the thread has already been created but is idle, which isn't inclusive of uncreated workers that could still be added to the pool.~~

This exposes the equivalent of the Java ThreadPoolExecutor's [`getActiveCount`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#getActiveCount--): "Returns the approximate number of threads that are actively executing tasks."

### DIFF
--- a/lib/concurrent-ruby/concurrent/executor/java_thread_pool_executor.rb
+++ b/lib/concurrent-ruby/concurrent/executor/java_thread_pool_executor.rb
@@ -73,6 +73,11 @@ if Concurrent.on_jruby?
         @executor.getCompletedTaskCount
       end
 
+      # @!macro thread_pool_executor_method_active_count
+      def active_count
+        @executor.getActiveCount
+      end
+
       # @!macro thread_pool_executor_attr_reader_idletime
       def idletime
         @executor.getKeepAliveTime(java.util.concurrent.TimeUnit::SECONDS)
--- a/lib/concurrent-ruby/concurrent/executor/fixed_thread_pool.rb
+++ b/lib/concurrent-ruby/concurrent/executor/fixed_thread_pool.rb
@@ -39,6 +39,10 @@ module Concurrent
   #   The number of tasks that have been completed by the pool since construction.
   #   @return [Integer] The number of tasks that have been completed by the pool since construction.
 
+  # @!macro thread_pool_executor_method_active_count
+  #   The number of threads that are actively executing tasks.
+  #   @return [Integer] The number of threads that are actively executing tasks.
+
   # @!macro thread_pool_executor_attr_reader_idletime
   #   The number of seconds that a thread may be idle before being reclaimed.
   #   @return [Integer] The number of seconds that a thread may be idle before being reclaimed.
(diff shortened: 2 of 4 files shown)


lib/concurrent-ruby/concurrent/executor/java_thread_pool_executor.rb
```diff
@@ -73,6 +73,11 @@ def completed_task_count
         @executor.getCompletedTaskCount
       end
 
+      # @!macro thread_pool_executor_attr_reader_available_worker_count
+      def available_worker_count
```
ioquatix, line 77: As discussed, should this just be `available_count` or is the `worker_count` taxonomy established here/elsewhere?

ioquatix:
```ruby

5.times do
  Thread.new do
    if pool.available_worker_count > 0 # query
      pool << proc {...} # use
    end
  end
end
```

This PR encourages time-of-check / time-of-use race conditions.

While I see the value of this interface/method, I don't think it solves the original problem.

Given that `max_queue: 0` == unbounded queue is probably impossible to change, maybe we should implement the desired behaviour with `max_queue: nil`.

Alternatively, if we think that is a terrible idea (consistency `max_queue: 1` vs `max_queue: 0` is definitely confusing), then another option is to deprecate the current syntax, release that, rework the interface, and then release that some time later.

1. Make `max_queue: 0` and `max_queue: nil` the same but issue a deprecation warning for the former.
2. Release minor.
3. Make `max_quueue: 0` do the logical thing.

eregon:
I'm OK to add this as a monitoring method, and of course yeah it's always potentially stale info, as for other monitoring methods.

It doesn't address #684 though, we'd probably need to fix the ThreadPoolExecutor to handle having a not-unlimited queue.

ioquatix:
@eregon `max_queue: 0` => unbounded. To me, that's pretty confusing. However, code search reveals at least some usage. My concern is, those people may have been expecting a queue of size 0.

What do you think of using `max_queue: nil` to mean no queue?

bensheldon:
oh! I think I understand this queuing thing better. The `synchronous: true` option is actually synonymous with "no queue", and easier to understand in the JavaThreadPoolExecutor because a `SynchronousQueue` is no queue at all (["A synchronous queue does not have any internal capacity, not even a capacity of one."](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html)):

https://github.com/ruby-concurrency/concurrent-ruby/blob/9f40827be9a8a192a6993a8d157bd3ed0662ada0/lib/concurrent-ruby/concurrent/executor/java_thread_pool_executor.rb#L117-L125

...though confusingly it _must_ be combined with `max_queue: 0`. 

I don't object if we wanted to say `max_queue: nil` was just shorthand for `max_queue: 0, synchronous: true`

ioquatix (approved):

PR #1003, opened 2023-07-05
Fix TimerTask example output

bensheldon:
The correct number of booms _is_ 5. Noted in https://github.com/ruby-concurrency/concurrent-ruby/pull/996#discussion_r1252858025


ioquatix (approved):

ioquatix:
Thanks!

PR #1005, opened 2023-07-17
Use executor from arg in then_on/rescue_on/chain_on for Promises

tgwizard:
When doing some spelunking through the code I saw that these 3 methods don't use the executor passed in through the argument. I believe they should, as the other `*_on` methods do.


eregon (approved):
I'm not very familiar with this code but yeah looking at it it makes a lot of sense, thanks for the fix!

PR #1013, opened 2023-11-20
Drop dependency on mutex_m

casperisfine:
In Ruby 3.3.0 `mutex_m` is promoted to a default gem, which requires to add it to the gemspec/gemfile.

But given how little usage it has in `concurrent-ruby` I think it might as well just be not used.

Additionally including `Mutex_m` cause many undesirable methods to be exposed, so simply using a Mutex instance variable is preferable.

cc @eregon


eregon (approved):
Looks good, thanks!

eregon:
Released in https://github.com/ruby-concurrency/concurrent-ruby/releases/tag/v1.2.3

Issue #1015, opened 2023-12-07
`Concurrent::Promises::Future#value!` can return `nil` even when promise is not resolved yet


907th:
### Details

```
* Operating system:                mac
* Ruby implementation:             Ruby
* `concurrent-ruby` version:       1.2.2
* `concurrent-ruby-ext` installed: no
* `concurrent-ruby-edge` used:     no
```

When main process received `SIGINT` currently running `future.value!` returns `nil` but `future.pending?` is true and it can be resolved later.

The documentation tells us:
> Returns:
> (Object, nil, timeout_value) — the value of the Future when fulfilled, or nil on rejection, or timeout_value on timeout.

So I think either the documentation needs to be updated or the signal handling needs to be fixed.

### Background

I worked on a code base which used chunks of code like:
```ruby
task1, task2, ... = Concurrent::Promises.future { ... }, ...
value1, value2, ... = task1.value!, ...
work_with(value1, value2, ...)
```

It was written in the assumption that `nil` value means rejection.

I wanted to implement `SIGINT` handling in the main thread and provide this code with the gracefull shutdown feature. But this code started to report future rejections (false positives) after sending `SIGINT` to the main process.

eregon:
Thanks for the report.
That's very surprising, as `Signal.trap("INT") {}` should mean sending SIGINT should do nothing.

But actually with a little debugging what happens is ConditionVariable#wait has a spurious wakeup on CRuby (.wait returns even though neither .broadcast nor .signal has been called) in that case, probably because CRuby does SIGVTALRM/interrupt logic anyway it seems, even with the trap overridden.
Although even though the trap is overridden, it still has to execute that empty block on the main thread, so that's probably why we get that internal interrupt.
That's maybe a CRuby issue, because for instance there is no such issue on TruffleRuby.

Regardless,
https://github.com/ruby-concurrency/concurrent-ruby/blob/068212f0f0e0497d0ea0dc37bc25dfbaddffc378/lib/concurrent-ruby/concurrent/promises.rb#L775
should be `until`, not `unless`, i.e., there should always be a loop around a ConditionVariable#wait.

eregon:
CRuby issue for this: https://bugs.ruby-lang.org/issues/20047

bensheldon:
[...]

PR #1017, opened 2023-12-07
Loop around ConditionVariable#wait in ReentrantReadWriteLock

eregon:
* To workaround spurious wakeups in CRuby: https://bugs.ruby-lang.org/issues/20047

Related: https://github.com/ruby-concurrency/concurrent-ruby/issues/1015

Unfortunately, this fails `spec/concurrent/atomic/reentrant_read_write_lock_spec.rb`.
I'm not sure why, help to investigate appreciated.


eregon:
OK the comments in the file around these `ns_wait if` explain why it's safe and should remain that way, so it's safe to not change this.

PR #1026, opened 2024-01-14
docs: update promises grammar

G-Rath:
Noticed this while reviewing the docs :)

### DIFF
--- a/lib/concurrent-ruby/concurrent/promises.rb
+++ b/lib/concurrent-ruby/concurrent/promises.rb
@@ -64,7 +64,7 @@ module Concurrent
         resolvable_event_on default_executor
       end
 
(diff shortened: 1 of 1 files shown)


lib/concurrent-ruby/concurrent/promises.rb
```diff
@@ -64,8 +64,8 @@ def resolvable_event
         resolvable_event_on default_executor
       end
 
-      # Created resolvable event, user is responsible for resolving the event once by
-      # {Promises::ResolvableEvent#resolve}.
+      # Created a resolvable event, user is responsible for resolving the event once
```
eregon, line 67:
```suggestion
      # Creates a resolvable event, user is responsible for resolving the event once
```

eregon (approved):

PR #1034, opened 2024-01-17
Add `mutex_m` as a runtime dependancy | Run CI against Ruby head

joshuay03:
Context: https://github.com/puma/puma/pull/3320

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


eregon:
As you can see https://github.com/ruby-concurrency/concurrent-ruby/actions/runs/7563147565/job/20594974217?pr=1034 is successful.
concurrent-ruby doesn't depend on mutex_m anymore in the latest release and since https://github.com/ruby-concurrency/concurrent-ruby/pull/1013.

eregon:
I'm +1 to add ruby-head in CI though.

eregon:
But it's already tested in https://github.com/ruby-concurrency/concurrent-ruby/blob/master/.github/workflows/experimental.yml so then I think it's safe to close this.

PR #1035, opened 2024-01-21
cgroup support for physical_processor_count

usiegl00:
As rails depends on `physical_processor_count` to spin up the right number of puma workers, rails will fall over in a containerized environment where cgroups are used to limit cpu time. This pr is an attempt to accurately reflect the available cpu cores when the ruby process is running inside a cgroup.

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -36,8 +36,9 @@ module Concurrent
               when /darwin\d\d/
                 IO.popen("/usr/sbin/sysctl -n hw.physicalcpu", &:read).to_i
               when /linux/
-                if Dir.exist?("/sys/fs/cgroup/cpu,cpuacct") && (cfs_quota_us = IO.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i) > 0
-                  (cfs_quota_us / IO.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us").to_i.to_f).ceil
+                # https://kernel.googlesource.com/pub/scm/linux/kernel/git/glommer/memcg/+/cpu_stat/Documentation/cgroups/cpu.txt
+                if Dir.exist?("/sys/fs/cgroup/cpu,cpuacct") && (cfs_quota_us = File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i) > 0
+                  (cfs_quota_us / File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us").to_f).ceil
                 else
                   cores = {} # unique physical ID / core ID combinations
                   phy   = 0
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -36,17 +36,21 @@ module Concurrent
               when /darwin\d\d/
                 IO.popen("/usr/sbin/sysctl -n hw.physicalcpu", &:read).to_i
               when /linux/
-                cores = {} # unique physical ID / core ID combinations
-                phy   = 0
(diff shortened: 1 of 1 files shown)


lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -36,17 +36,21 @@ def compute_physical_processor_count
...
-                cores = {} # unique physical ID / core ID combinations
-                phy   = 0
-                IO.read("/proc/cpuinfo").scan(/^physical id.*|^core id.*/) do |ln|
-                  if ln.start_with?("physical")
-                    phy = ln[/\d+/]
-                  elsif ln.start_with?("core")
-                    cid        = phy + ":" + ln[/\d+/]
-                    cores[cid] = true if not cores[cid]
+                if Dir.exist?("/sys/fs/cgroup/cpu,cpuacct") && (cfs_quota_us = IO.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i) > 0
+                  (cfs_quota_us / IO.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us").to_i.to_f).ceil
```
eregon, line 40: Is there some documentation about these cgroups files?
It'd be nice to add a link as a code comment here.

eregon, line 40: .to_i.to_f can probably be just .to_f

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -36,17 +36,21 @@ def compute_physical_processor_count
...
               when /linux/
-                cores = {} # unique physical ID / core ID combinations
-                phy   = 0
-                IO.read("/proc/cpuinfo").scan(/^physical id.*|^core id.*/) do |ln|
-                  if ln.start_with?("physical")
-                    phy = ln[/\d+/]
-                  elsif ln.start_with?("core")
-                    cid        = phy + ":" + ln[/\d+/]
-                    cores[cid] = true if not cores[cid]
+                if Dir.exist?("/sys/fs/cgroup/cpu,cpuacct") && (cfs_quota_us = IO.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i) > 0
```
eregon, line 39: Could you use File.read instead of IO.read? It seems clearer (same on the next line).

eregon (commented):
This looks good.
Did you test it works as expected?

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -36,17 +36,22 @@ def compute_physical_processor_count
...
-                cores = {} # unique physical ID / core ID combinations
-                phy   = 0
-                IO.read("/proc/cpuinfo").scan(/^physical id.*|^core id.*/) do |ln|
-                  if ln.start_with?("physical")
-                    phy = ln[/\d+/]
-                  elsif ln.start_with?("core")
-                    cid        = phy + ":" + ln[/\d+/]
-                    cores[cid] = true if not cores[cid]
+                # https://kernel.googlesource.com/pub/scm/linux/kernel/git/glommer/memcg/+/cpu_stat/Documentation/cgroups/cpu.txt
+                if Dir.exist?("/sys/fs/cgroup/cpu,cpuacct") && (cfs_quota_us = File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i) > 0
```
byroot, line 40: IMO we shouldn't change the behavior of `physical_processor_count`. We should instead have a `cpu_quota` method that returns a float, and then eventually a `usable_processor_count` which returns `cpu_quota * processor_count`.

This way it's up to the caller to decide if they want to `floor/round/ceil`

byroot: Also I think what you implemented here is cgroups V1? Not sure how much it's used these days, in cgroups V2 the info is in `/sys/fs/cgroup/cpu.max`.

eregon:
Thank you for the PR, let's continue this on https://github.com/ruby-concurrency/concurrent-ruby/pull/1038

PR #1038, opened 2024-01-29
Add Concurrent.available_processor_count that is cgroups aware

casperisfine:
Closes: https://github.com/ruby-concurrency/concurrent-ruby/pull/1035

A running gag since the introduction of containerization is software that starts one process per logical or physical core while running inside a container with a restricted CPU quota and totally blowing up memory usage in containerized environments.

The proper question to ask is how many CPU cores are usable, not how many the machine has. To do that we have to read the cgroup info from `/sys`. There is two way of doing it depending on the version of cgroups used.

Co-Authored-By: @usiegl00

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -153,7 +153,7 @@ module Concurrent
   # For performance reasons the calculated value will be memoized on the first
   # call.
   #
-  # @return [nil, Float] number of available processors seen by the OS or Java runtime
+  # @return [nil, Float] number of available processors
   def self.available_processor_count
     processor_counter.available_processor_count
   end


lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +143,16 @@ def self.processor_count
...
+  # In most cases it is the same than `processor_count`, however if the
+  # process is inside a cgroup with a dedicated CPU quota (typically Docker),
+  # the quota will be returned instead.
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [Float] number of usable processors seen by the OS or Java runtime
+  def self.usable_processor_count
+    processor_counter.usable_processor_count
+  end
```
eregon, line 157: Could you expose `cpu_quota` on `Concurrent` too and document it?

casperisfine: Sure, no problem.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -21,6 +22,23 @@ def physical_processor_count
         @physical_processor_count.value
       end
 
+      def usable_processor_count
```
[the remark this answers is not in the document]
casperisfine: I have no strong opinion on the name.

> And that maybe shouldn't count hyperthreads (because they very rarely give parallelism), i.e. be based on physical_processor_count? I'm unsure, WDYT?

Unfortunately I'm afraid this ship has sailed, when working on the cloud, it's all logical cores (vCPU), and unless I'm mistaken, the cgroup cpu_quota is in logical cores too, so I think it would be incorrect to only use physical ones.

casperisfine:
Arf:
```
     NoMethodError:
       undefined method `match?' for "linux":String
```

I'll fix 2.3 compat.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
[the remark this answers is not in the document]
eregon: > Unfortunately I'm afraid this ship has sailed, when working on the cloud, it's all logical cores (vCPU)

Argh, indeed that seems to be the case.
At least if we expose `cpu_quota` people can use it with either `physical_processor_count` or `processor_count` depending if they want true parallelism count or basically that x 2 if hyperthreaded.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +143,20 @@ def self.processor_count
...
+  # Number of processors usable for process scheduling.
+  # Returns `nil` if there is no quota, or a `Float` if the
+  # process is inside a cgroup with a dedicated CPU quota (typically Docker).
+  #
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [nil, Float] number of usable processors seen by the OS or Java runtime
+  def self.available_processor_count
+    processor_counter.available_processor_count
```
eregon, line 156: It feels weird to return a Float for a method named `*_count`.
Maybe we should round down instead?
And if people want a different rounding they can use `cpu_quota` directly

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -21,6 +22,23 @@ def physical_processor_count
         @physical_processor_count.value
       end
 
+      def available_processor_count
+        cpu_count = processor_count.to_f
+        quota = cpu_quota
```
eregon, line 27: Shouldn't we multiply those? (like mentioned in https://github.com/ruby-concurrency/concurrent-ruby/pull/1035#discussion_r1469717571)

byroot: Nah, that was a mistake on my part. The cgroup quota is expressed in how many microsecond of CPU time you can use over a given period (also in micro-second).

So if you use docker with something like `--cpus=2.5`, the `cpu.max` will be something like `250000000 10000000`, so when we divide them we get our `2.5` back.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
byroot: I'd rather not round because it's too arbitrary. For very small local service it's not uncommon to give a very small fraction of CPU, e.g `0.5`.

So IMO it's up to the caller to decide whether they want to use a "partial" core or not. e.g. if the quota is `2.5`, should you start 2 or 3 processes to maximize utilization? Well it depends, how CPU intensive are these processes? Hence I prefer not to truncate the data.

> And if people want a different rounding they can use cpu_quota directly

That's another option, but in the context of very portable code, `cpu_quota` is likely to return `nil`, so you need to fallback on the other method. This one was really meant as some short of handy shortcut. 

But up to you.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -21,6 +22,23 @@ def physical_processor_count
...
+
+        return cpu_count if quota.nil?
+
+        if quota > cpu_count
```
eregon, line 31: Can quota be higher than processor_count?
Because if not it could be just `cpu_quota || processor_count.to_f` which is quite a bit more readable.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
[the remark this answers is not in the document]
eregon: > At least if we expose `cpu_quota` people can use it with either `physical_processor_count` or `processor_count` depending if they want true parallelism count or basically that x 2 if hyperthreaded.

Unfortunately that won't be easy because the quota is a number of CPUs and not a ratio, so one needs to know if the quota is in logical or physical cores.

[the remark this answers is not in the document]
eregon: I played a bit with it and `podman run --cpus=2 --rm -it fedora:latest` with 2 `bash -c 'while true; do true; done'  &` definitely uses 2 real cores. But I guess in general one never knows and it would in practice depend what else is running on the host.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
casperisfine: Yes it can:

```
$ podman run -it ruby:3.3 bash
root@c7a0b7656940:/# ruby -r etc -e 'p Etc.nprocessors'
4
-- exit
$ podman run --cpus 42 -it ruby:3.3 bash
root@221a93a8b707:/# cat /sys/fs/cgroup/cpu.max
4200000 100000
```

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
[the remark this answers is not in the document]
casperisfine: https://docs.sylabs.io/guides/latest/user-guide/cgroups.html#cpu-limits seem to indicate that `cpu.max` is indeed expressed in physical cores. I'm starting to have a bit of a headache here...

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
eregon: Thanks, could you add a comment about it in the code?
(and gah this CPU quota feature seems pretty hacky, doesn't integrate into `sysconf(_SC_NPROCESSORS_ONLN)`, no proper limits, etc.)

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
eregon: OK, I think I agree with the current PR & semantics then.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +143,20 @@ def self.processor_count
...
   end
+
+  # Number of processors usable for process scheduling.
+  # Returns `nil` if there is no quota, or a `Float` if the
+  # process is inside a cgroup with a dedicated CPU quota (typically Docker).
+  #
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [nil, Float] number of usable processors seen by the OS or Java runtime
```
eregon, line 156:
```suggestion
  # @return [nil, Float] number of available processors seen by the OS or Java runtime
```

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +143,20 @@ def self.processor_count
   def self.physical_processor_count
     processor_counter.physical_processor_count
   end
+
+  # Number of processors usable for process scheduling.
```
eregon, line 149:
```suggestion
  # Number of processors available for process scheduling.
```

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +145,20 @@ def self.processor_count
...
+  # Returns `nil` if there is no quota, or a `Float` if the
```
eregon, line 150:
```suggestion
  # Returns `nil` if there is no #cpu_quota, or a `Float` if the
```
Ideally it'd link to the other method, not sure the correct way to do that with YARD.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +145,20 @@ def self.processor_count
...
+  #
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [nil, Float] number of usable processors seen by the OS or Java runtime
+  def self.available_processor_count
+    processor_counter.available_processor_count
+  end
+
+  def self.cpu_quota
```
eregon, line 161: Could you document this?

eregon (commented):
Looks good, just need a bit more documentation care.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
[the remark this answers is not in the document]
eregon: > https://docs.sylabs.io/guides/latest/user-guide/cgroups.html#cpu-limits seem to indicate that `cpu.max` is indeed expressed in physical cores. I'm starting to have a bit of a headache here...

I'm not sure, it's really not clear :sweat: 

I guess what docker/podman --cpus sets is the cgroup `--cpus` given it can be a fraction? But maybe there is x2 in between?

The linked docs says:
> --cpus sets the number of CPUs, or fractional CPUs, that the container can use. The minimum is 0.01 or one hundredth of a physical CPU.
> The maximum is the number of CPU cores on your system.

So if by CPU cores they mean real core, ignoring hyper threads then indeed it's physical cores.

But, I just tried with `--cpus=8` and `--cpus=16` on my 8 core + 8 hyper threads AMD machine, and with `--cpus=8` it uses approximately 50% of all CPU (looking at htop and `Load average` which was ~8), while with `--cpus=16` it uses everything (`Load average` which was ~16).
So I think it is logic processor count in the end for the Podman (and I guess Docker too) --cpus argument.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
eregon: AH also this doesn't return `nil` so probably worth a while pass over the docs of this method.
It should also mention it uses `processor_count` aka logical processor count.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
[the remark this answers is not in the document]
eregon: So I think the code is fine as-is and we should use logical, and document it.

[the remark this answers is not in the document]
casperisfine: Oh, thanks for testing. I guess none of that documentation can be trusted -_-.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -60,6 +80,24 @@ def compute_physical_processor_count
       rescue
         return 1
       end
+
+      def compute_cpu_quota
+        if RbConfig::CONFIG["target_os"].match(/linux/i)
```
eregon, line 85:
```suggestion
        if RbConfig::CONFIG["target_os"].include?("linux")
```

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +145,31 @@ def self.processor_count
...
+  #
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [nil, Float] number of available processors seen by the OS or Java runtime
+  def self.available_processor_count
+    processor_counter.available_processor_count
+  end
+
+  # The maximun number of processors cores available for process scheduling.
```
eregon, line 161:
```suggestion
  # The maximum number of processors cores available for process scheduling.
```

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +145,31 @@ def self.processor_count
...
+  # Returns `nil` if there is no enforced limit, or a `Float` if the
+  # process is inside a cgroup with a dedicated CPU quota (typically Docker).
+  #
+  # Note that nothing prevent to set a CPU quota higher than the actual number of
+  # cores on the system.
+  #
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [nil, Float] Maximum number of available processors seen by the OS or Java runtime
```
eregon, line 171:
```suggestion
  # @return [nil, Float] Maximum number of available processors as set by a cgroup CPU quota, or nil if none set
```

spec/concurrent/utility/processor_count_spec.rb
```diff
@@ -17,4 +17,79 @@ module Concurrent
...
+      expect(counter.cpu_quota).to be == 1.5
+    end
+
+  end
+
+  RSpec.describe '#available_processor_count' do
+
+    it 'returns #processor_count if #cpu_quota is nil' do
+      expect(Concurrent::processor_counter).to receive(:cpu_quota).and_return(nil)
+      available_processor_count = Concurrent::available_processor_count
```
eregon, line 75:
```suggestion
      available_processor_count = Concurrent.available_processor_count
```

spec/concurrent/utility/processor_count_spec.rb
```diff
@@ -17,4 +17,79 @@ module Concurrent
...
+      expect(available_processor_count).to be == Concurrent::processor_count
+      expect(available_processor_count).to be_a Float
+    end
+
+    it 'returns #processor_count if #cpu_quota is higher' do
+      expect(Concurrent::processor_counter).to receive(:cpu_quota).and_return(Concurrent::processor_count.to_f * 2)
+      available_processor_count = Concurrent::available_processor_count
```
eregon, line 82:
```suggestion
      available_processor_count = Concurrent.available_processor_count
```

eregon (approved):
Thank you!

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -107,4 +145,31 @@ def self.processor_count
...
   end
+
+  # Number of processors cores available for process scheduling.
+  # Returns `nil` if there is no #cpu_quota, or a `Float` if the
+  # process is inside a cgroup with a dedicated CPU quota (typically Docker).
+  #
+  # For performance reasons the calculated value will be memoized on the first
+  # call.
+  #
+  # @return [nil, Float] number of available processors seen by the OS or Java runtime
```
eregon, line 156:
```suggestion
  # @return [nil, Float] number of available processors
```

casperisfine:
Thanks!

nertzy:
Thanks for this! The PR's title has a different method name than the one that landed. I added PR #1049 to fix the CHANGELOG.

edmorley:
Hi! The PR title says the new API was named `Concurrent.usable_processor_count`, however, it looks like it landed as `Concurrent.available_processor_count` instead?

Would someone mind updating the PR title to match?

PR #1042, opened 2024-03-01
Align Java Executor Service behavior for `shuttingdown?`, `shutdown?`

### LINKED ISSUE #1041: ThreadPoolExecutor#shutdown? inconsistency in JRuby and C Ruby
In the scenario in which a ThreadPoolExecutor has received `#shutdown` and there are active threads/tasks, there is an inconsistency between what JRuby and C Ruby return for `#shutdown?`:

- JRuby will return `#shutdown? => true` while there are still active threads/tasks, at the same time as `#shuttingdown? => true`. [Java's executor service](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ExecutorService.html) narrowly defines `#isShutdown` as "no new tasks will be accepted" which is distinct from `#isTerminated` which is "all tasks have completed following shut down".
- The Ruby implementation does more of what my expectations are: `#shutdown? => false` while there are still active threads/tasks, at the same time as `#shuttingdown? => true`.

It's fairly easy to work around this inconsistency using `executor.shutdown? && !executor.shuttingdown?` to achieve the same result in both rubies for determining when there are no more threads/tasks on a shutdown executor. It would be nice if the behavior was aligned.

bensheldon:
Fixes #1041 

- `shuttingdown?` now uses `isShutdown && !isTerminated`
- `shutdown?` now accurately represents when all tasks have terminated
- ~~`kill` now waits for all tasks to terminate (it waits indefinitely; I'm not sure if it's possible for that to hang on Java)~~ Extracted `#kill` change to #1044

### DIFF
--- a/lib/concurrent-ruby/concurrent/executor/java_executor_service.rb
+++ b/lib/concurrent-ruby/concurrent/executor/java_executor_service.rb
@@ -57,15 +57,11 @@ if Concurrent.on_jruby?
       end
 
       def ns_shuttingdown?
-        if @executor.respond_to? :isTerminating
-          @executor.isTerminating
-        else
-          false
-        end
+        @executor.isShutdown && !@executor.isTerminated
       end
 
       def ns_shutdown?
-        @executor.isShutdown || @executor.isTerminated
+        @executor.isTerminated
       end
 
       class Job
--- a/spec/concurrent/executor/executor_service_shared.rb
+++ b/spec/concurrent/executor/executor_service_shared.rb
@@ -3,7 +3,7 @@ require 'concurrent/atomic/atomic_boolean'
 require 'concurrent/atomic/atomic_fixnum'
 require 'timeout'
 
-RSpec.shared_examples :executor_service do
+RSpec.shared_examples :executor_service do |immediate_type: false|
 
   after(:each) do
     subject.shutdown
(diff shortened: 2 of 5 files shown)


lib/concurrent-ruby/concurrent/executor/java_executor_service.rb
```diff
@@ -46,6 +46,7 @@ def shutdown
       def kill
         synchronize do
           @executor.shutdownNow
+          ok = @executor.awaitTermination(60, java.util.concurrent.TimeUnit::SECONDS) until ok
```
eregon, line 49: Could you reuse `wait_for_termination` here?
Also I think this line should be moved the `synchronize`, it's typically problematic to do any blocking call in `synchronize`.

bensheldon:
1. Something in here is causing a deadlock on JRuby in the TimerSet tests (I'm guessing it's the wait_for_termination)
2. We need a maximum Actions workflow timeout of not 6 hours.

eregon:
Could you make a PR to add timeouts of 10 minutes (which looks enough from looking at previous runs) for all CI workflows?

eregon:
Can you reproduce the issue locally with TimerSet on JRuby?
I guess it might be related to that `kill` change.
So it might be worth to extract that change to a separate PR.
In general I wonder if it's really worth the trouble & complexity of having different implementations for JRuby when there is already a pure-Ruby implementation of some concurrent-ruby class.

bensheldon:
@eregon Thank you for the help! 🙇🏻 

- Extracted `kill` to #1042 
- Added GitHub actions timeout in #1043 

> In general I wonder if it's really worth the trouble & complexity of having different implementations for JRuby when there is already a pure-Ruby implementation of some concurrent-ruby class.

I feel mixed. A lot of Concurrent Ruby is inspired by Java objects, so it seems like it would be _better_ (simpler? more performant? less potential for bikeshedding or 2nd-systeming them?) to lightly wrap them rather than use a Ruby implementation. I have similar vague feelings about Native C Extensions too; which to chit-chat also makes me wonder if YJIT would make that less worthwhile.

eregon (approved):

eregon:
> I have similar vague feelings about Native C Extensions too; which to chit-chat also makes me wonder if YJIT would make that less worthwhile.

The concurrent-ruby C ext is [only for atomic variables](https://github.com/ruby-concurrency/concurrent-ruby/blob/8b9b0da4a37585ce5eb71516aca55e93bde39115/ext/concurrent-ruby-ext/rb_concurrent.c).
Those are likely quite a bit faster than [using a Mutex](https://github.com/ruby-concurrency/concurrent-ruby/blob/8b9b0da4a37585ce5eb71516aca55e93bde39115/lib/concurrent-ruby/concurrent/atomic/mutex_atomic_boolean.rb#L8).

PR #1051, opened 2024-06-06
Remove dependency on `win32ole`

### LINKED ISSUE #1048: Ruby 3.5 to drop `win32ole`
Ruby is continuing its path of requiring more gems to be specified. `win32ole` is next on the list (among others) and will start warning in Ruby 3.4: https://github.com/ruby/ruby/commit/f365bef0c78085b7db9b6736ae59c1657baa79a3

Usage is here, to get the number of cores on windows: https://github.com/ruby-concurrency/concurrent-ruby/blob/b5e090b9da95c63cd006721ae6a7e594ec78512b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb#L71-L75

Something like the following may be a suitable replacement:
```rb
IO.popen("wmic cpu get NumberOfCores", &:read).scan(/\d+/).map(&:to_i).reduce(:+)
```

Output of that command:
`NumberOfCores  \n\n4              \n\n\n\n`

I have also openend the same report on `parallel` at https://github.com/grosser/parallel/issues/345 if you are interested. Both implementations for this are basically the same.

Earlopain:
This will become bundled in Ruby 3.5

Unfortunately there is no portable way of checking for this. The wmic command is deprecated, though I don't observe this myself on W11 (yet?)

Closes #1048

The powershell variant will work in at least releases from 2012 (Windows 8):
https://learn.microsoft.com/en-us/previous-versions/powershell/module/cimcmdlets/get-ciminstance?view=powershell-3.0
https://learn.microsoft.com/en-us/previous-versions/powershell/module/microsoft.powershell.utility/select-object?view=powershell-3.0
https://learn.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.4

There are no docs from earlier so I can't say about versions from before that. Maybe supporting 12 year old versions is good enough and `wmic` doesn't even need to be used? Let me know.

This is overall very defensive in what could happen because I'm not primarily developing on windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,20 +6,24 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}-latest
     timeout-minutes: 10
 
     strategy:
       fail-fast: false
       matrix:
         ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, jruby, truffleruby]
+        os: [ubuntu]
+        include:
+          - ruby: 3.3
+            os: windows
 
     env:
       JAVA_OPTS: '-Xmx1024m'
       RUBYOPT: '-w'
       JRUBY_OPTS: '--dev'
 
-    name: "Tests: Ruby ${{ matrix.ruby }}"
+    name: "Tests: Ruby ${{ matrix.ruby }} - ${{ matrix.os }}"
     steps:
     - name: Clone Repo
       uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


eregon (approved):
This looks great, thank you!

eregon:
@MSP-Greg I wonder if you might know what's the best way to get the number of (physical) cores on a Windows machine from Ruby?

MSP-Greg:
@eregon

The PowerShell command `Get-CimInstance -ClassName Win32_Processor | Select-Object -Property NumberOfCores` works.

On my system, it shows 10.  `Etc.nprocessors` shows 20, both on Windows and WSL2/Ubuntu.  I have a single CPU.

Do you mean via c code?

eregon:
I meant via the command-line.
OK, thanks for confirming it works.

eregon:
I made a release with this: https://github.com/ruby-concurrency/concurrent-ruby/releases/tag/v1.3.2

PR #1053, opened 2024-06-08
Improve speed for windows `Get-CimInstance`

Earlopain:
This is a small followup for #1051.

In terms of the previous code, the powershell invocation currently does `select * from Win32_Processor` which is slower. I missed this in the other PR, this improves the speed when detection runs on windows.

The pipe is still needed since that command returns all properties anyways, they are just empty/not computed.

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -69,7 +69,7 @@ module Concurrent
                 cores.count
               when /mswin|mingw/
                 # Get-CimInstance introduced in PowerShell 3 or earlier: https://learn.microsoft.com/en-us/previous-versions/powershell/module/cimcmdlets/get-ciminstance?view=powershell-3.0
-                result = run('powershell -command "Get-CimInstance -ClassName Win32_Processor  | Select-Object -Property NumberOfCores"')
+                result = run('powershell -command "Get-CimInstance -ClassName Win32_Processor -Property NumberOfCores | Select-Object -Property NumberOfCores"')
                 if !result || $?.exitstatus != 0
                   # fallback to deprecated wmic for older systems
                   result = run("wmic cpu get NumberOfCores")


eregon (approved):
Thanks.
Do you have any measurements before & after?

Earlopain:
Yeah, sure. My bad, I should have lead with that:

On my machine the old command takes about 1s, this new one only about 0.2s.

I've measures this inside powershell by wrapping the expressions inside `Measure-Command { ... }` which I guess is its equivalent to `time`.

eregon:
Could you also measure `wmic cpu get NumberOfCores`? I wonder how fast that is.

Yeah 1s seems a lot, I'll try to do a release with this improvement soon.

Earlopain:
`wmic` completes in about 0.1s. Some overhead comes from starting up powershell, that alone takes about 50ms.`win32ole` is a c extension and (probably) doesn't go through the trouble of starting a process in the first place: it completes in ~60ms

eregon:
I wonder, is powershell available on all Windows machines?
Otherwise we could have an issue if/when `wmic` is removed, if it's possible to also not have `powershell` on these Windows machines.

eregon:
FWIW I left a comment on https://bugs.ruby-lang.org/issues/20309#note-22, making win32ole a bundled gem seems to cause more harm than help (in performance & portability).

Earlopain:
Well, the speed regression is my fault, I just wasn't careful enough. As for if powershell will be available, I would say you can rely on this as much as `cmd.exe`. Sometime in Windows 10/11 they actually made it the default shell.

The features available of this gem go far beyond just querying wmi, like interacting with excel/word, or other modern software like internet explorer. Basically a full-on automation interface. If you look at the gem itself, the C code to pipe that all through to ruby is quite extensive.

Earlopain:
In that sense it was more like using a canon to kill a fly. It is sad that `etc` doesn't have a equivalent for this since  `nprocessors` exists

PR #1058, opened 2024-07-30
Add `Concurrent.cpu_requests` that is cgroups aware.

heka1024:
In K8s environments, cpu requests are often more useful than limits, but concurrent-ruby currently lacks a related method, so I've added it. Below are links to articles about the usefulness of cpu requests.

- https://medium.com/directeam/kubernetes-resources-under-the-hood-part-3-6ee7d6015965
- https://x.com/thockin/status/1134193838841401345

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -12,6 +12,7 @@ module Concurrent
         @processor_count          = Delay.new { compute_processor_count }
         @physical_processor_count = Delay.new { compute_physical_processor_count }
         @cpu_quota                = Delay.new { compute_cpu_quota }
+        @cpu_shares               = Delay.new { compute_cpu_shares }
       end
 
       def processor_count
@@ -39,6 +40,10 @@ module Concurrent
 
       def cpu_quota
         @cpu_quota.value
+      end
+
+      def cpu_shares
+        @cpu_shares.value
       end
 
       private
@@ -110,6 +115,20 @@ module Concurrent
             return nil if max == 0
             period = File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us").to_f
             max / period
+          end
+        end
+      end
+
+      def compute_cpu_shares
+        if RbConfig::CONFIG["target_os"].include?("linux")
+          if File.exist?("/sys/fs/cgroup/cpu.weight")
+            # cgroups v2: https://docs.kernel.org/admin-guide/cgroup-v2.html#cpu-interface-files
+            # Ref: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2254-cgroup-v2#phase-1-convert-from-cgroups-v1-settings-to-v2
+            weight = File.read("/sys/fs/cgroup/cpu.weight").to_f
+            ((((weight - 1) * 262142) / 9999) + 2) / 1024
+          elsif File.exist?("/sys/fs/cgroup/cpu/cpu.shares")
+            # cgroups v1: https://kernel.googlesource.com/pub/scm/linux/kernel/git/glommer/memcg/+/cpu_stat/Documentation/cgroups/cpu.txt
+            File.read("/sys/fs/cgroup/cpu/cpu.shares").to_f / 1024
           end
         end
       end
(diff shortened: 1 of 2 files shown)


eregon:
Could you summarize in one/a few sentences why one should use cpu requests instead of cpu limits?

heka1024:
@eregon CPU limits can lead to throttling and reduced performance under high load. So, as [Tim Hockin](https://x.com/thockin) suggested, I've set only requests in my application at company.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -187,4 +212,12 @@ def self.available_processor_count
   def self.cpu_quota
     processor_counter.cpu_quota
   end
+
+  # The CPU shares requested by the process. For performance reasons the calculated
+  # value will be memoized on the first call.
+  #
+  # @return [Float, nil] CPU shares requested by the process, or nil if not set
+  def self.cpu_requests
```
eregon, line 220: I wonder what's the best name here.
`cpu_shares` might be better since it reads from `/sys/fs/cgroup/cpu.weight` or `/sys/fs/cgroup/cpu/cpu.shares`.
And we already have `cpu_quota` which reads from `cpu.cfs_quota_us`/`cpu.max`.

`requests` sounds like a Kubernetes-specific term, notably Docker calls it shares in https://docs.docker.com/config/containers/resource_constraints/#cpu

So could you rename to `cpu_shares`?

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -41,6 +42,10 @@ def cpu_quota
         @cpu_quota.value
       end
 
+      def cpu_requests
+        @cpu_requests.value / 1024
```
eregon, line 46: This will NoMethodError if there is neither cgroup 1 or 2.
I think it makes sense to move the `/ 1024` in `compute_cpu_requests`

eregon:
@byroot @casperisfine Could you review this too?

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
heka1024: @eregon Thank you for suggestion. I've made changes based on comment.

eregon (approved):

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -113,6 +118,26 @@ def compute_cpu_quota
           end
         end
       end
+
+      def compute_cpu_shares
+        if RbConfig::CONFIG["target_os"].include?("linux")
+          if cgroups_v2?
+            # Ref: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2254-cgroup-v2#phase-1-convert-from-cgroups-v1-settings-to-v2
+            weight = File.read('/sys/fs/cgroup/cpu.weight').to_f
```
byroot, line 126: It's probably fine, but in `compute_cpu_quota` we only ever read files we tested for existence prior.

Here we check `/sys/fs/cgroup/cpu.max` and assume it means `/sys/fs/cgroup/cpu.weight` exists? Maybe that's a correct assertion, but I personally don't know if it's true.

I think ultimately it would be simpler to not extract `cgroups_v1? / cgroups_v2?` and check if `/sys/fs/cgroup/cpu.weight` exists here.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -113,6 +118,26 @@ def compute_cpu_quota
...
+            ((((weight - 1) * 262142) / 9999) + 2) / 1024
+          elsif cgroups_v1?
+            File.read('/sys/fs/cgroup/cpu/cpu.shares').to_f / 1024
```
byroot, line 129: Same issue here.

byroot (approved):
I'd undo the "extract method" for extra safety, but other than that LGTM.

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
eregon: Agreed, this seems safer

heka1024:
@byroot @heka1024 I'd undo the "extract method" and replace it with file existence check. Could you review this once again?

byroot (approved):

eregon (approved):

PR #1059, opened 2024-08-07
Fix the doc of `Concurrent.available_processor_count`

y-yagi:
Currently, the doc says that this methods returns `nil` if there is no cpu_quota, but actually not.
https://github.com/ruby-concurrency/concurrent-ruby/blob/6f7c91ab5a4e99d850f69c958572a536426277e8/lib/concurrent-ruby/concurrent/utility/processor_counter.rb#L30

This PR fixes the doc to match the behavior.

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -83,7 +83,7 @@ module Concurrent
                   # Bail out if both commands returned something unexpected
                   processor_count
                 else
-                  # powershell: "\nNumberOfCores\n-------------\n            4\n\n\n" 
+                  # powershell: "\nNumberOfCores\n-------------\n            4\n\n\n"
                   # wmic:       "NumberOfCores  \n\n4              \n\n\n\n"
                   result.scan(/\d+/).map(&:to_i).reduce(:+)
                 end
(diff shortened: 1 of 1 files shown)


lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -181,13 +181,14 @@ def self.physical_processor_count
   end
 
   # Number of processors cores available for process scheduling.
-  # Returns `nil` if there is no #cpu_quota, or a `Float` if the
-  # process is inside a cgroup with a dedicated CPU quota (typically Docker).
+  # Returns number of processors seen by the OS if there is no #cpu_quota, or
+  # the value considering CPU quota if the process is inside a cgroup with a
+  # dedicated CPU quota (typically Docker).
```
eregon, line 186: It seems hard to read given we describe the condition reversed, like if we explain the `else` part before the `then` part.
How about?
```suggestion 
  # This method takes in account the CPU quota if the process is inside a cgroup with a
  # dedicated CPU quota (typically Docker).
  # Otherwise it returns the same value as #processor_count but as a Float.
```

eregon:
cc @byroot

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
y-yagi: Thanks for your review. 
I applied your suggestion.

PR #1060, opened 2024-08-09
Fix the return value of `Concurrent.available_processor_count` when `cpu.cfs_quota_us` is -1

y-yagi:
I tried to use `Concurrent.available_processor_count` in `parallel` gem, but we got some feedback `Concurrent.available_processor_count` returned a negative value.
https://github.com/grosser/parallel/pull/348#issuecomment-2275859126 
https://github.com/grosser/parallel/issues/349#issuecomment-2275953547

According to the https://docs.kernel.org/scheduler/sched-bwc.html#management, The default value of `cpu.cfs_quota_us` is -1. In that case, cgroup does not adhere to any CPU time restrictions.

This PR adds the case of `cpu.cfs_quota_us` is -1 to `#cpu_quota` to return processor count from `Concurrent.available_processor_count` in that case.

### DIFF
--- a/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
+++ b/lib/concurrent-ruby/concurrent/utility/processor_counter.rb
@@ -112,7 +112,9 @@ module Concurrent
           elsif File.exist?("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us")
             # cgroups v1: https://kernel.googlesource.com/pub/scm/linux/kernel/git/glommer/memcg/+/cpu_stat/Documentation/cgroups/cpu.txt
             max = File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i
-            return nil if max == 0
+            # If the cpu.cfs_quota_us is -1, cgroup does not adhere to any CPU time restrictions
+            # https://docs.kernel.org/scheduler/sched-bwc.html#management
+            return nil if max <= 0
             period = File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us").to_f
             max / period
           end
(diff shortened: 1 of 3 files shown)


eregon:
cc @byroot @casperisfine

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
```diff
@@ -112,7 +112,9 @@ def compute_cpu_quota
           elsif File.exist?("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us")
             # cgroups v1: https://kernel.googlesource.com/pub/scm/linux/kernel/git/glommer/memcg/+/cpu_stat/Documentation/cgroups/cpu.txt
             max = File.read("/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us").to_i
-            return nil if max == 0
+            # If the cpu.cfs_quota_us is -1, cgroup does not adhere to any CPU time restrictions
+            # https://docs.kernel.org/scheduler/sched-bwc.html#management
+            return nil if max == 0 || max == -1
```
eregon, line 117: Maybe we should check `if max <= 0`?
Seems a tiny bit safer and more efficient as we can't interpret a value <= 0 safely.
Could you change to that?

eregon:
Thanks and sorry for the bug.
Could you add a CHANGELOG entry?

lib/concurrent-ruby/concurrent/utility/processor_counter.rb
y-yagi: Thanks for your review. I updated the condition.

y-yagi:
I've added a CHANGELOG entry.

eregon (approved):

eregon:
I released https://github.com/ruby-concurrency/concurrent-ruby/releases/tag/v1.3.4 with this fix

Issue #1061, opened 2024-09-04
"Logger" warning with Ruby 3.3.5


voxik:
Running railties test suite with Ruby 3.3.5, I observer following failures due to warning which appears to come from concurrent-ruby:

~~~
* Test file: test/application/bin_setup_test.rb
/usr/share/gems/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/deprecation.rb:1: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
/usr/share/gems/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/deprecation.rb:1: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
Run options: --seed 53132

# Running:

F

Failure:
ApplicationTests::BinSetupTest#test_bin_setup_output [test/application/bin_setup_test.rb:51]:
--- expected
+++ actual
@@ -2,10 +2,16 @@
 The Gemfile's dependencies are satisfied
 
 == Preparing database ==
+/usr/share/gems/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/deprecation.rb:1: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
+You can add logger to your Gemfile or gemspec to silence this warning.
 Created database 'app_development'
 Created database 'app_test'
 
 == Removing old logs and tempfiles ==
+/usr/share/gems/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/deprecation.rb:1: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
+You can add logger to your Gemfile or gemspec to silence this warning.
 
 == Restarting application server ==
+/usr/share/gems/gems/concurrent-ruby-1.1.9/lib/concurrent-ruby/concurrent/concern/deprecation.rb:1: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
+You can add logger to your Gemfile or gemspec to silence this warning.
 "


rails test test/application/bin_setup_test.rb:30

.

[...]

PR #1062, opened 2024-09-04
Remove dependency on logger

eregon:
* To fix the Ruby 3.3.5 warnings: https://github.com/ruby-concurrency/concurrent-ruby/issues/1061
* concurrent-ruby only uses 7 constants from Logger, so just copy those over.

This should be fully compatible and address the warnings.
Only if one uses the long-deprecated `Concurrent.create_stdlib_logger` they might need to add `logger` to their Gemfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ To use the tools in the Edge gem it must be required separately:
 require 'concurrent-edge'
 ```
 
-If the library does not behave as expected, `Concurrent.use_stdlib_logger(Logger::DEBUG)` could 
+If the library does not behave as expected, `Concurrent.use_simple_logger(:DEBUG)` could
 help to reveal the problem.
 
 ## Installation
--- a/docs-source/actor/celluloid_benchmark.rb
+++ b/docs-source/actor/celluloid_benchmark.rb
@@ -7,11 +7,7 @@ require 'celluloid/autostart'
 # require 'stackprof'
 # require 'profiler'
 
-logger                          = Logger.new($stderr)
-logger.level                    = Logger::INFO
-Concurrent.configuration.logger = lambda do |level, progname, message = nil, &block|
-  logger.add level, message, progname, &block
-end
+Concurrent.use_simple_logger(:INFO)
 
 scale       = 1
 ADD_TO      = (100 * scale).to_i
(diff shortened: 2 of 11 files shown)


ioquatix (approved):
It looks okay to me, but I'd personally try to remove logging as a concern of this library.

ioquatix:
Probably (1) and (3) - basically remove it entirely.

Fryguy:
The release of this seems to have broken Rails 7.0 codebases with 

```
rake aborted!
NameError: uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger (NameError)

    Logger::Severity.constants.each do |severity|
    ^^^^^^
/Users/jfrey/.gem/ruby/3.3.6/gems/activesupport-7.0.8.7/lib/active_support/logger_thread_safe_level.rb:12:in `<module:LoggerThreadSafeLevel>'
/Users/jfrey/.gem/ruby/3.3.6/gems/activesupport-7.0.8.7/lib/active_support/logger_thread_safe_level.rb:9:in `<module:ActiveSupport>'
/Users/jfrey/.gem/ruby/3.3.6/gems/activesupport-7.0.8.7/lib/active_support/logger_thread_safe_level.rb:8:in `<top (required)>'
/Users/jfrey/.gem/ruby/3.3.6/gems/activesupport-7.0.8.7/lib/active_support/logger_silence.rb:5:in `<top (required)>'
/Users/jfrey/.gem/ruby/3.3.6/gems/activesupport-7.0.8.7/lib/active_support/logger.rb:3:in `<top (required)>'
/Users/jfrey/.gem/ruby/3.3.6/gems/activesupport-7.0.8.7/lib/active_support.rb:29:in `<top (required)>'
/Users/jfrey/.gem/ruby/3.3.6/gems/railties-7.0.8.7/lib/rails.rb:7:in `<top (required)>'
/Users/jfrey/.gem/ruby/3.3.6/gems/sprockets-rails-3.5.2/lib/sprockets/railtie.rb:1:in `<top (required)>'
/Users/jfrey/dev/manageiq-ui-classic/lib/manageiq/ui/classic/engine.rb:2:in `<top (required)>'
```

I think ActiveSupport::LoggerThreadSafeLevel happened to be depending on Logger implicitly, so the fix is likely to just update Rails to require 'logger' there.  I'll send a PR over there for the fix.  Workaround locally is to add the require early.

Fryguy:
See https://github.com/rails/rails/pull/54264

PR #1071, opened 2024-11-25
Set rake-compiler source and target to Java 8

headius:
This patch forces rake-compiler to emit bytecode for Java 8 level of JVM. rake-compiler currently defaults to Java 1.7 source and target, but that version is no longer supported by recent JDKs such as Java 21 (see https://github.com/rake-compiler/rake-compiler/pull/242 for a patch to update rake-compiler).

This patch sets our minimum Java level to 8, bypassing the default in rake-compiler and allowing builds on Java 21+.


headius:
This can be merged and released at any time. Without this, it is not possible to build the concurrent-ruby extension for JRuby on Java 21 or higher.

headius:
Small correction: this does not require immediate release. Merging to the repository fixes issues in JRuby 10's CI which runs a minimum of Java 21.

Once released, all extensions will be Java 8 minimum, but I believe the current 1.7 extensions can still load on Java 21 (it just can't output 1.7 bytecode).

Issue #1066, opened 2024-10-10
Unexpected pruning behaviour with consecutive task batches


joshuay03:
```
* Operating system:                macOS Sequoia 15.0.1
* Ruby implementation:             ruby 3.4.0preview1
* `concurrent-ruby` version:       1.3.3
* `concurrent-ruby-ext` installed: no
* `concurrent-ruby-edge` used:     no
```

Cross-post of https://github.com/rails/rails/pull/53211.

Specifically for the part in the included script where a second 'batch' of work is assigned to a `Concurrent::ThreadPoolExecutor` pool and it is incorrectly pruned before the work is processed:

```ruby
begin
  require "concurrent-ruby"
  
  pool = Concurrent::ThreadPoolExecutor.new(
    min_threads: 1,
    max_threads: 4,
    max_queue: 0,
    idletime: 3
  )
  # First thread is lazily spawned.
  pp pool.length #=> 0
  
  work = -> { sleep 2 }
  
  # Batch (gap-less individual units) of work.
  10.times { pool << work }
  # Wait for state updates.
  sleep 0.25
  # Expected scale up.
  pp pool.length #=> 4
  # Wait for all work to be processed.
  # This is sufficient cause work is I/O bound and parallel.
  sleep 10
  pp pool.length #=> 4
  # Wait until idle time of all threads has elapsed.
  # This is sufficient; only needs to be greater than the idle time of the last busy thread.
  sleep 5
  # Not scaled down.
  # Prune will only take place when next unit of work is received, despite idle time elapse.
  pp pool.length #=> 4

  # Wait for a while to show no change.
  sleep 20
  pp pool.length #=> 4
  
  # Another batch of work.
  10.times { pool << work }
  # Wait for state updates.
  sleep 0.25
  # This case is the most interesting, and might need to be addressed in concurrent-ruby.
  # If bulk work comes in, since prune is called right after assignment / queuing, there's a race
  # condition between when the ready workers size is checked for prune, and the threads start
  # processing the work, which is when the ready size is updated. As a result, we end up with a
  # single thread handling all the work i.e. the pool is prematurely scaled down, and stays that way
  # since all units of work have been assigned / queued.
  pp pool.length #=> 1
  # Wait for all work to be processed.
  # Work is now sequential.
  sleep 25
  pp pool.length #=> 1
  # Wait until idle time of all threads has elapsed.
  sleep 5
  pp pool.length #=> 1

  # Wait for a while to show no change.
[...]

Issue #1075, opened 2024-12-19
CachedThreadPool does not spin down idle threads


SamSaffron:
In the Ruby implementation: 

```
require 'concurrent'

puts "Thread Count: #{Thread.list.count}"

pool = Concurrent::CachedThreadPool.new( min_threads: 0, max_threads: 5, idletime: 1 )

puts "Thread Count: #{Thread.list.count}"
# we have 1 thread for now

5.times do
  pool.post do
    puts "Hello from thread #{Thread.current.object_id}"
  end
end
# we now have 5

sleep 2
puts "Thread Count: #{Thread.list.count}"
# we still have 5

pool.post do
  puts "Hello from thread #{Thread.current.object_id}"
end

sleep 1

puts "Thread Count: #{Thread.list.count}"
# we just issued a prune as a side effect so we have 2

pool.prune_pool
sleep 2
# we need to sleep cause prune_pool is async

puts "Thread Count: #{Thread.list.count}"
# we now have 1 thread
```

CachedThreadPool only spins down threads when you queue work, so if you are done queuing work you need to remember to wait for it to be done and then issue a #prune_pool call 

This is technically very tricky cause you need to carry an extra thread for that. The Ruby design should change so threads just wait idle_time for new work and spin down if no new work arrives.

bensheldon:
What's the cost/downside you're seeing with a sleeping thread on Queue#pop?

To prune precisely I think we have a couple of options:

- another thread to sleep and prune. Downside as you note is another thread. 
- replace the Queue object with a queue-with-wait implementation (I don't think Concurrent Ruby has one, though TimerSet is close it) so the thread could prune itself.

SamSaffron:
@bensheldon mainly complexity, it is complexity we do not want to carry. 

I wrote a queue with a conditional var here: 

https://github.com/discourse/discourse/pull/30364

I am more than happy for this to be lifted into concurrent ruby, it has a few tiny discoursisms (multisite support and logging patterns) but is safe.

For us this is perfect, I have not benched it against the current implementation but for our usage we always tend to make DB calls so we are not optimising for microsecond execution times shortest would be usually 10s of milliseconds.

eregon:
> I wrote a queue with a conditional var here:

FWIW I also recall writing one in https://github.com/ruby/timeout/blob/607d8c6fbe4d86db1cf22846e89198b47cec7161/lib/timeout.rb#L105-L109
[...]

PR #1079, opened 2025-01-17
Less aggressive async pruning for RubyThreadPoolExecutor

joshuay03:
Fixes https://github.com/ruby-concurrency/concurrent-ruby/issues/1066

<details>
<summary>Issue script output with this patch</summary>

```ruby
0
4
4
4
1
pool should ideally be scaled up here
4
1
1
1
4
4
4
1
1
1
1
```
</details>

Adds a pruner thread to `RubyThreadPoolExecutor` with the intention of making pruning a lot less aggressive which should reduce the likelihood of prematurely scaling down the size of the pool right after being assigned a large workload.

### DIFF
--- a/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
+++ b/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
@@ -20,8 +20,11 @@ module Concurrent
     # @!macro thread_pool_executor_constant_default_max_queue_size
     DEFAULT_MAX_QUEUE_SIZE     = 0
 
-    # @!macro thread_pool_executor_constant_default_thread_timeout
+    # @!macro thread_pool_executor_constant_default_thread_idle_timeout
     DEFAULT_THREAD_IDLETIMEOUT = 60
+
+    # @!macro thread_pool_executor_constant_default_pool_prune_timeout
+    DEFAULT_POOL_PRUNETIMEOUT = 30
 
     # @!macro thread_pool_executor_constant_default_synchronous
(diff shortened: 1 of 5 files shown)


lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -20,9 +20,12 @@ class RubyThreadPoolExecutor < RubyExecutorService
     # @!macro thread_pool_executor_constant_default_max_queue_size
     DEFAULT_MAX_QUEUE_SIZE     = 0
 
-    # @!macro thread_pool_executor_constant_default_thread_timeout
+    # @!macro thread_pool_executor_constant_default_thread_idle_timeout
     DEFAULT_THREAD_IDLETIMEOUT = 60
 
+    # @!macro thread_pool_executor_constant_default_pool_prune_timeout
+    DEFAULT_POOL_PRUNETIMEOUT = 30
```
joshuay03, line 27: A somewhat arbitrary number, but this is also what puma uses.

eregon:
Have you seen https://github.com/ruby-concurrency/concurrent-ruby/issues/1075? It seems quite related.
Would this PR solve that issue too?

joshuay03:
> Have you seen #1075? It seems quite related. Would this PR solve that issue too?

Oh I haven't, last I checked was when I opened my issue and there wasn't anything similar. Yes this should solve that too, I'll leave a comment there.

joshuay03:
Closing in favour of https://github.com/ruby-concurrency/concurrent-ruby/pull/1082

PR #1082, opened 2025-02-08
Better pruning for RubyThreadPoolExecutor

joshuay03:
Closes https://github.com/ruby-concurrency/concurrent-ruby/issues/1066
Closes https://github.com/ruby-concurrency/concurrent-ruby/issues/1075
Alternative to https://github.com/ruby-concurrency/concurrent-ruby/pull/1079


lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -114,9 +120,9 @@ def worker_task_completed
       synchronize { @completed_task_count += 1 }
     end
 
-    # @!macro thread_pool_executor_method_prune_pool
-    def prune_pool
```
joshuay03, line 118: This public API no longer makes sense now that the pruning is automatic. It seems it was only [added](https://github.com/ruby-concurrency/concurrent-ruby/pull/938) because of the Ruby thread pool's synchronous pruning, which the Java version didn't suffer from, hence why it's a no-op there.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
```diff
@@ -0,0 +1,58 @@
...
+      alias_method :<<, :push
+
+      def pop(non_block = false, timeout: nil)
+        if non_block && timeout
+          raise ArgumentError, "can't set a timeout if non_block is enabled"
+        end
+
+        if non_block
+          super(true)
+        elsif @mutex.synchronize { empty? && timed_out?(timeout) { @cond_var.wait(@mutex, timeout) } }
```
bensheldon, line 33: Just flagging that there could be a spurious wake up signal on the Condition Variable that prematurely invokes this condition. If that happened, I think you'd accidentally end up in a blocking pop. 

I think the recommended design is to set an additional variable flag in addition to the signal to disambiguate between the meaningful signal and the spurious ones. The primitive for this is a Concurrent::Event.

joshuay03: Ah good catch, I'll address, thanks!

joshuay03: > I think the recommended design is to set an additional variable flag in addition to the signal (in push) to disambiguate between the meaningful signal and the spurious ones. The primitive for this is a Concurrent::Event.

I think we can rely on `#empty?` instead? - it can only be `true` if there was a spurious wakeup.

joshuay03: I've addressed, let me know what you think.

bensheldon: I imagine that's sufficient 👍🏻 Thank you!

I think the only thing is there could be a race condition on `empty?` between a non-timeout `pop` and a timeout `pop` on the same Queue instance, because the non-timeout call to `super` is not synchronized with the same mutex. ...but we're building this for a very specific use case where that wouldn't happen.

joshuay03: Good point! Although I think it’s best to ensure the implementation is robust in case it gets relied on more heavily in future work.

I've pushed a slightly different approach where we rely on the non-blocking super call instead.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
```diff
@@ -0,0 +1,79 @@
...
+        end
+      end
+      alias_method :deq, :pop
+      alias_method :shift, :pop
+
+      private
+
+      def ns_timed_out?(timeout, &cond_var_wait)
+        # https://github.com/ruby/ruby/pull/4256
+        timed_out = if RUBY_VERSION >= '3.1'
```
eregon, line 59: I'd rather simplify to the other path, I don't want to rely on wait returning `nil`, it makes things more complicated and might not be implemented in 3.1 on non-CRuby.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
```diff
@@ -0,0 +1,79 @@
...
+        elsif timeout
+          @mutex.synchronize do
+            while timeout.positive? # handle spurious wakeups
+              begin
+                return super(true)
+              rescue ThreadError
+                # empty, wait until woken up
+              end
+
+              timed_out, timeout = ns_timed_out?(timeout) { @cond_var.wait(@mutex, timeout) }
```
eregon, line 42: Could you simplify to something like https://github.com/ruby/timeout/blob/607d8c6fbe4d86db1cf22846e89198b47cec7161/lib/timeout.rb#L105-L109 (which I wrote) and avoid the extra method?
That seems a lot simpler and I believe it's fully correct.

eregon (commented):
Finally taking a look at this, sorry for the delay

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
```diff
@@ -0,0 +1,79 @@
+module Concurrent
+  module Collection
+    # @!visibility private
+    # @!macro ruby_timeout_queue
+    class RubyTimeoutQueue < ::Queue
```
eregon, line 5: Let's use `private_constant` on this to make it very clear

lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
eregon: Removing it seems too much of a breaking change though, how about keeping it as noop, maybe with a deprecation warning?

joshuay03: Put it back and add a deprecation warning for this and the Java implementation.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
joshuay03: Done.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
joshuay03: Done.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
joshuay03: Done.

lib/concurrent-ruby/concurrent/collection/ruby_timeout_queue.rb
```diff
@@ -0,0 +1,57 @@
...
+              begin
+                return super(true)
+              rescue ThreadError
+                # empty, wait until woken up
+              end
+
+              @cond_var.wait(@mutex, deadline - now)
+            end
+
+            nil
```
eregon, line 46:
```suggestion
            while (now = Concurrent.monotonic_time) < deadline and empty?
              @cond_var.wait(@mutex, deadline - now)
            end
            begin
              return super(true)
            rescue ThreadError
              # still empty
              nil
            end
```
This way we avoid many `ThreadError` exceptions, and we're under the mutex so there is no race possible between the `empty?` and the `pop`/super.

Also importantly even if the timeout is very small (or even 0) it still tries to pop once.

joshuay03: [Addressed](https://github.com/ruby-concurrency/concurrent-ruby/compare/9739e9ad72a9e9f7789d911d50dbe25ede887d29..6aa4ee5bea7bec58d3186c404533c7b86e729894).

lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -274,23 +279,19 @@ def ns_remove_busy_worker(worker)
...
-        end
+    def ns_remove_ready_worker(worker)
+      if index = @ready.index { |rw, _| rw == worker }
+        @ready.delete_at(index)
       end
+      true
+    end
 
-      @next_gc_time = Concurrent.monotonic_time + @gc_interval
+    def ns_prunable_capacity
```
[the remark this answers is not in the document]
joshuay03: https://github.com/ruby-concurrency/concurrent-ruby/compare/6aa4ee5bea7bec58d3186c404533c7b86e729894..101b66eec768283b5d69e7fd17d5f6446e6f90f3

lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -95,8 +97,16 @@ def remaining_capacity
     end
 
     # @!visibility private
-    def remove_busy_worker(worker)
-      synchronize { ns_remove_busy_worker worker }
+    def prunable_capacity
+      synchronize { ns_prunable_capacity }
```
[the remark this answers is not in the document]
joshuay03: Oh good point, we should synchronize the entire pruning process else a worker that just became idle could be assigned work between when it thinks it should prune itself and when it actually does...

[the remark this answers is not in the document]
joshuay03: [Addressed](https://github.com/ruby-concurrency/concurrent-ruby/compare/101b66eec768283b5d69e7fd17d5f6446e6f90f3..4d088a031c9c33989191248b425ca701ba1b37bc).

bensheldon (approved):
I went through this and I think it looks good. I'm very excited about this new behavior 🙇🏻 

- The Ruby TimeoutQueue implementation looks solid and resilient to idle wakeups and matches the interface of the native Ruby Queue-with-timeout.
- The pruning behavior in RubyThreadPoolExecutor looks correct. I spent a lot of time trying to understand the `prunable` behavior, but I think I get it now; I wasn't able to think up a scenario where it wouldn't result in unintended behavior.
- Test cleanup looks defensible 👍🏻

lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb
```diff
@@ -338,17 +339,26 @@ def kill
...
+            loop do
+              timeout = prunable && my_pool.running? ? my_idletime : nil
+              case message = my_queue.pop(timeout: timeout)
+              when nil
+                if my_pool.prunable_capacity.positive?
+                  my_pool.remove_worker(self)
+                  throw :stop
+                end
+
+                prunable = false
```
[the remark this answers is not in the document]
bensheldon: I spent a lot of time staring at `prunable` and `ns_prunable_capacity`. If I understand it correctly, the intent is to avoid unnecessarily adding a timeout/wakeup to Threads that are part of the executor's `min_length`. 

I have no idea how to make the logic any clearer, but just noting (for myself more than anyone) that I reasoned through it.

joshuay03:
Thank you both for the reviews!

PR #1105, opened 2026-06-14
Modernize the concurrent-ruby-ext C extension

joshuay03:
Replace the per-platform CAS and memory-barrier code in `CAtomicReference`, `CAtomicBoolean`, and `CAtomicFixnum` with the abstractions in `ruby/atomic.h`, and switch the three classes from the old `DATA_PTR` / `RUBY_NEVER_FREE` pattern to `TypedData_Make_Struct` with `dmark`, `dfree`, `dsize`, and (on Ruby >= 2.7) `dcompact`. Stores go through `RB_OBJ_WRITE` and successful CAS calls fire `RB_OBJ_WRITTEN`, so write barriers are honored and GC compaction is supported.

Two version-gated fallbacks keep us building back to Ruby 2.3:

- `ruby/atomic.h` only became a public extension header in Ruby 3.0, so `extconf.rb` probes for it via `have_header`. When it isn't available the C code falls back to GCC/Clang `__atomic_*` builtins.
- GC compaction APIs were added in Ruby 2.7. A `CR_GC_COMPACTION` macro gated on `RUBY_API_VERSION_CODE` selects between `rb_gc_mark_movable` / `rb_gc_location` plus the `dcompact` slot and a plain `rb_gc_mark` on older Rubies.

`extconf.rb` also drops the `libkern/OSAtomic.h` probe and the `-march` flags, which no longer matter once we're on `ruby/atomic.h` or the compiler atomic builtins.

No behavioral change to any public API; the spec suite passes unchanged, including the atomic specs under `GC.stress`.

### DIFF
--- a/ext/concurrent-ruby-ext/atomic_boolean.c
+++ b/ext/concurrent-ruby-ext/atomic_boolean.c
@@ -1,54 +1,67 @@
-#include <ruby.h>
-
 #include "atomic_boolean.h"
 #include "atomic_reference.h"
 
-void atomic_boolean_mark(void *value) {
-  rb_gc_mark_maybe((VALUE) value);
-}
-
-const rb_data_type_t atomic_boolean_type = {
-  "Concurrent::CAtomicBoolean",
-  {
-    atomic_boolean_mark,
-    RUBY_NEVER_FREE,
-  },
-};
+CR_DEFINE_ATOMIC_DATA_TYPE(atomic_boolean_type, "Concurrent::CAtomicBoolean");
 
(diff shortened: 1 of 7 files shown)


eregon:
Sounds great, is it ready for review?

joshuay03:
> Sounds great, is it ready for review?

Not yet, I'll ping you when it is. This is hopefully the first PR of a few to modernize the extension and lay the groundwork for opt-in Ractor support across the atomic primitives.

joshuay03:
@eregon Ready for review 👍🏽

PR #1106, opened 2026-06-16
Add Ruby 4.0 in CI

eregon:
It was already tested in `Test isolated` mode, but good to test in "together" mode too

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, 3.4, jruby, truffleruby]
+        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, 3.4, '4.0', jruby, truffleruby]
         os: [ubuntu]
         include:
           - ruby: ruby


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, 3.4, jruby, truffleruby]
+        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, 3.4, '4.0', jruby, truffleruby]
```
joshuay03, line 16: Optional follow-up nit for consistency:

```suggestion
        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, 3.4, 4.0, jruby, truffleruby]
```

eregon: The problem is that would then pick 4.1 when 4.1 is released, because GitHub's YAML parser parses `4.0` as `4` :/
See https://github.com/ruby/setup-ruby#matrix-of-ruby-versions
Makes me want to do https://github.com/ruby/setup-ruby/issues/447 so at least we have the most likely intended behavior.

joshuay03: Sigh, all good in that case 👍🏽

PR #1107, opened 2026-06-18
Programatically set the `backtrace_location` on exception

Edouard-chin:
Hello 👋 , thank you for maintaining this library ! I'd like to introduce a change to allow for exception to be consistent when either calling `backtrace` or `backtrace_locations` on a rejected Future.

### Problem

When a Future is rejected, we set the backtrace with an array of string locations. This makes for some inconsistency because `exception.backtrace_locations` return nil but `exception.backtrace` return the backtrace.

I'd like to be able to programmatically set the backtrace_location as well. This would allow some libraries that depend on the backtrace_location to correcly detect the backtrace of an error.

### Solution

Since Ruby 3.4, it's now possible to set the backtrace_locations programatically. The way do do it is to call `set_backtrace` but with an array of `Thread::Backtrace::Location` objects.

### DIFF
--- a/lib/concurrent-ruby/concurrent/promises.rb
+++ b/lib/concurrent-ruby/concurrent/promises.rb
@@ -915,6 +915,7 @@ module Concurrent
     # Represents a value which will become available in future. May reject with a reason instead,
     # e.g. when the tasks raises an exception.
     class Future < AbstractEventFuture
+      SET_BACKTRACE_LOCATIONS_SUPPORTED = RUBY_VERSION >= '3.4'
 
       # Is it in fulfilled state?
       # @return [Boolean]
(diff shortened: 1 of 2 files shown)


lib/concurrent-ruby/concurrent/promises.rb
```diff
@@ -1014,17 +1014,23 @@ def exception(*args)
         raise Concurrent::Error, 'it is not rejected' unless rejected?
         raise ArgumentError unless args.size <= 1
         reason = Array(internal_state.reason).flatten.compact
+        locations_supported = RUBY_VERSION >= '3.4'
```
eregon: Let's make it a constant like:
```suggestion
        SET_BACKTRACE_LOCATIONS_SUPPORTED = RUBY_VERSION >= '3.4'
```
I would prefer to feature-detect this but it's probably a bit messy (would cause an exception on older Ruby, not nice with `-d`) so this is fine enough.

eregon (approved):
Looks good, just needs one change

Edouard-chin:
Thanks for the quick review !

PR #1109, opened 2026-07-08
Allow to use Concurrent::Map#compute_if_absent in a Ractor

Edouard-chin:
Allow to use Concurrent::Map#compute_if_absent in a Ractor

### Problem

It is currently not possible to use the `compute_if_absent` method in a Ractor because there is a NULL object used solely as a sentinel that isn't shareable.

```ruby
Ractor.new do
  map = Concurrent::Map.new
  map.compute_if_absent(3) { }
end

# Ractor::IsolationError:
#   can not access non-shareable objects in constant Concurrent::NULL by non-main ractor.
```

### Solution

I'd like to simply freeze the object being hold in the `NULL` constant.


bensheldon (approved):
Looks correct to me 👍

Would a release sooner rather than later be helpful for Ractor work?

Edouard-chin:
Thanks for the offer ! This would be super helpful to get a release sooner than later. I'm not seeing any other Ractor fixes for concurrent ruby (we have a backlog of ractor safety issues in a few gems but this was the only in this gem for our current project scope).

eregon:
I can do the release, I'd say let's wait a week or so just in case something else comes up.

eregon:
https://github.com/ruby-concurrency/concurrent-ruby/releases/tag/v1.3.8

Edouard-chin:
Thank you so much !!

PR #1112, opened 2026-08-29
Preserve reentrant lock state after invalid release

OskarEichler:
Checks read/write ownership before decrementing the thread-local hold count. Invalid releases still raise, but no longer wrap the counter and poison later valid lock operations.

Verified by focused reentrant lock specs and models; the review composite passes 2,796 examples with no failures.


bensheldon (commented):
Can this be tested?

OskarEichler:
Yes. I added regression coverage for both read and write releases. Each test first performs an invalid release, verifies IllegalOperationError, and then acquires and releases the same lock successfully; before this fix, the invalid release mutated the per-thread held count and the later acquire failed. The focused lock spec passes: 41 examples, 0 failures.

PR #1113, opened 2026-08-29
Reject serialized execution without a task

OskarEichler:
Raises the documented ArgumentError immediately when SerializedExecution post is called without a block, rather than queuing a nil task that fails later.

Verified by focused executor specs and models; the review composite passes 2,796 examples with no failures.


bensheldon (changes requested):
Please add a test for this

OskarEichler:
Added coverage for calling SerializedExecution#post without a block. The focused executor spec passes: 31 examples, 0 failures (2 existing pending examples).

bensheldon (approved):

PR #1114, opened 2026-08-29
Validate AtomicFixnum update results before mutation

OskarEichler:
Validates the block result through the same integer assignment path before changing state. Invalid update results now raise without corrupting the mutex-backed or native atomic value.

Verified by focused native/mutex specs and models; the review composite passes 2,796 examples with no failures.

### DIFF
--- a/spec/concurrent/atomic/atomic_fixnum_spec.rb
+++ b/spec/concurrent/atomic/atomic_fixnum_spec.rb
@@ -152,6 +152,17 @@ RSpec.shared_examples :atomic_fixnum do
       atomic = described_class.new(1000)
       expect(atomic.update { |v| v + 1 }).to eq 1001
     end
+
+    it 'rejects a non-integer result without changing the value' do
+      atomic = described_class.new(1000)
+
+      expect {
+        atomic.update { 'not an integer' }
+      }.to(raise_error { |error|
+        expect(error.class).to be(ArgumentError).or(be(TypeError))
+      })
+      expect(atomic.value).to eq 1000
+    end
   end
 end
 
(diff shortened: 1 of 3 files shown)


bensheldon (commented):
How would it become corrupted? Can it be tested?

OskarEichler:
The invalid block result replaces the stored integer in the mutex implementation; the C implementation likewise passed the arbitrary Ruby VALUE through compare-and-set without enforcing the numeric invariant. Later numeric operations then operate on a non-integer value. I added a shared regression example that returns a String, verifies the update raises, and verifies the previous integer remains intact. It runs against both MutexAtomicFixnum and CAtomicFixnum; after compiling the extension, the focused suite passes with 59 examples and 0 failures.

spec/concurrent/atomic/atomic_fixnum_spec.rb
```diff
@@ -152,6 +152,17 @@
...
       expect(atomic.update { |v| v + 1 }).to eq 1001
     end
+
+    it 'rejects a non-integer result without changing the value' do
+      atomic = described_class.new(1000)
+
+      expect {
+        atomic.update { 'not an integer' }
+      }.to(raise_error { |error|
+        expect(error.class).to be(ArgumentError).or(be(TypeError))
```
bensheldon, line 162: Can you be specific here? Why would it be 2 different kinds of error?

OskarEichler: The two implementations already expose different validation errors: the C extension uses Ruby’s `Check_Type(...)`, which raises `TypeError`, while `MutexAtomicFixnum#ns_set` delegates to `NativeInteger.ensure_integer`, which raises `ArgumentError`. This shared spec runs against both implementations, so it accepts the existing implementation-specific class while asserting the important invariant: the failed update does not mutate the value.

bensheldon:
Thank you!

PR #1115, opened 2026-08-29
Return the Promises waiter count for inspection

OskarEichler:
Returns the AtomicFixnum value used by the lock-free event waiter registry. The current inspection path calls each on AtomicFixnum and raises instead of reporting the count.

Verified by focused Promises specs and models; the review composite passes 2,796 examples with no failures.

### DIFF
--- a/lib/concurrent-ruby/concurrent/promises.rb
+++ b/lib/concurrent-ruby/concurrent/promises.rb
@@ -726,7 +726,7 @@ module Concurrent
       # For inspection.
       # @!visibility private
       def waiting_threads
-        @Waiters.each.to_a
+        @Waiters.value
       end
 
       # @!visibility private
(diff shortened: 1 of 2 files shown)


bensheldon (changes requested):
Please add a test for the behavior as it sounds like this has existed because it's untested.

OskarEichler:
Added a regression spec that checks the waiter count before, during, and after a blocked wait. The focused Promises spec passes: 61 examples, 0 failures.

OskarEichler:
Follow-up: the first regression spec used an `ensure` form that Ruby 2.3 and 2.4 could not parse, which caused the three legacy-Ruby CI failures. I rewrote the cleanup using an explicit `begin`/`ensure`, removed the unnecessary safe-navigation calls, and added the missing assertion that the waiter count returns to zero after resolution.

The focused Promises spec now passes with 61 examples and 0 failures on Ruby 4.0.6, and the updated spec is syntactically compatible with the legacy parser form exercised by CI.

